PR #1, opened 2019-03-28
Add warning when using HTTP

SISheogorath:
We want to spread some awareness about plaintext HTTP connections. So we
show a big warning for plain HTTP connections and ask users to ask their
admin to setup HTTPS.

We may want to extent this to use link to a guide on how to do this.

Previously: https://github.com/hackmdio/codimd/pull/937


davidmehren:
We could still land this in 1.6.1, what do you think @SISheogorath?

SISheogorath:
I think it's dead in the water. It was an idea at some point, but the implementation was never really sophisticated and it's probably not solving any issue. So let's kill it off :D

PR #2, opened 2019-03-28
WIP: Add ratelimiting to all requests

SISheogorath:
This patch adds rate-limiting to make it harder for brute-force and DOS
attacks or similar to success.

There is some evaluation needed on what are really useful values for the
rate limiting especially when proxies and/or docker is involved.

Previously: https://github.com/hackmdio/codimd/pull/970


davidmehren:
Any hard feelings on pushing this to 2.0?

davidmehren:
Moved to 2.0 with #442

PR #3, opened 2019-03-28
Feature/asciinema

SISheogorath:
This PR is a working implementation for an asciinema integration. But due to security concerns it should be rewritten to not dynamically embed external JS from https://asciinema.org.

Instead we should try to rewrite this to use asciinema-player instead which is also available on CDNJS.

Feel free to pick this up and figure it out during Hacktober :)

Previously: https://github.com/hackmdio/codimd/pull/978

### DIFF
--- a/lib/csp.js
+++ b/lib/csp.js
@@ -5,7 +5,7 @@ var CspStrategy = {}
 
 var defaultDirectives = {
   defaultSrc: ['\'self\''],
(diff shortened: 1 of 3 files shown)


davidmehren:
I've created a corresponding [issue](https://github.com/codimd/react-client/issues/322) for the new 2.0 frontend, as this most likely won't make it into any earlier release.

PR #5, opened 2019-03-28
Add some version checks to setup script

SISheogorath:
There are some distros out there, shipping quite outdated packages with
them, which results in weird issues. Let's check that we have a more or
less up to date version of node and yarn installed.

This patch adds those checks and this way should provide some helpful
error messages to people who try to install CodiMD.

Previously: https://github.com/hackmdio/codimd/pull/1075

### DIFF
--- a/bin/setup
+++ b/bin/setup
@@ -2,6 +2,8 @@
 
 set -e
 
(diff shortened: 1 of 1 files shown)


davidmehren:
I think this is still a helpful feature. If you update the versions we could get this into 1.6.1. :smiley:

DerMolly (approved):

InnayTool (approved):

mrdrogdrog (approved):

PR #6, opened 2019-03-28
Add config for toobusy middleware

SISheogorath:
With very low CPU frequency or bad IO situation, as well as not-loaded
JS CodiMD happens to present unneeded "I'm busy"-messages to users.

This patch allows to configure the lag. The default is taken from the
library but set in our own default configs.

Previously: https://github.com/hackmdio/codimd/pull/1078

### DIFF
--- a/docs/configuration-env-vars.md
+++ b/docs/configuration-env-vars.md
@@ -35,6 +35,7 @@ defaultNotePath can't be set from env-vars
 | `CMD_FORBIDDEN_NOTE_IDS` | `'robots.txt'` | disallow creation of notes, even if `CMD_ALLOW_FREEURL` is `true` |
 | `CMD_IMAGE_UPLOAD_TYPE` | `imgur`, `s3`, `minio`, `lutim` or `filesystem` | Where to upload images. For S3, see our Image Upload Guides for [S3](guides/s3-image-upload.md) or [Minio](guides/minio-image-upload.md), also there's a whole section on their respective env vars below. |
 | `CMD_SOURCE_URL` | `https://github.com/codimd/server/tree/<current commit>` | Provides the link to the source code of CodiMD on the entry page (Please, make sure you change this when you run a modified version) |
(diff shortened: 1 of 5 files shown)


docs/configuration-env-vars.md
```diff
@@ -35,6 +35,7 @@ defaultNotePath can't be set from env-vars
 | `CMD_FORBIDDEN_NOTE_IDS` | `'robots.txt'` | disallow creation of notes, even if `CMD_ALLOW_FREEURL` is `true` |
 | `CMD_IMAGE_UPLOAD_TYPE` | `imgur`, `s3`, `minio`, `lutim` or `filesystem` | Where to upload images. For S3, see our Image Upload Guides for [S3](guides/s3-image-upload.md) or [Minio](guides/minio-image-upload.md), also there's a whole section on their respective env vars below. |
 | `CMD_SOURCE_URL` | `https://github.com/codimd/server/tree/<current commit>` | Provides the link to the source code of CodiMD on the entry page (Please, make sure you change this when you run a modified version) |
+| `CMD_TOOBUSY_LAG` | `70` | CPU time for one eventloop tick until node throttles connections. (milliseconds) |
```
ccoenen: This whould either be `CMD_TOOBUSYLAG` or `CMD_TOO_BUSY_LAG` - because of capitalisation of the variable, probably the latter. I know this has already been merged, but we should fix this before the 1.4 release.

SISheogorath: Somehow I would prefer `CMD_TOOBUSYLAG`, agreed?

SISheogorath: Òn the other hand there might be something like `CMD_TOOBUSY_ENABLED` in future

ccoenen: I think if we're going with `tooBudyLag` as a variable name, it should be `CMD_TOO_BUSY_LAG`. I'm hoping that we'll refactor all of the config stuff eventually, and the more it follows a pattern (in this case: each word camel case / each word `_` underscored) that's quite easy.

PR #21, opened 2019-04-01
Refactoring highlight code

SISheogorath:
Was working on this a while ago, but never open a PR. Time to change :3 This moves most of the highlight code into the rendering process of markdown-it and this way hopefully helps to make future extensions easier.

Not sure how much of a difference it makes from a performance perspective, but hopefully simplifies the control-flow for highlighting so it becomes easier to understand.

### DIFF
--- a/public/js/lib/highlight/externals.js
+++ b/public/js/lib/highlight/externals.js
@@ -1,0 +1,30 @@
+'use strict';
+
+import {escapeHtml}  from './utils'
+
+let externals = {}
+
+externals.checkLang = function (lang) {
+  if (typeof lang !== 'string') {
+    return false
+  }
+  const languages = ['sequence', 'flow', 'graphviz', 'mermaid', 'abc']
+  return languages.includes(lang.toLowerCase())
+}
+
+externals.highlight = function (code, lang) {
+  const escapedCode = escapeHtml(code)
+  if (lang === 'sequence') {
+    return `<div class="sequence-diagram raw">${escapedCode}</div>`
+  } else if (lang === 'flow') {
+    return `<div class="flow-chart raw">${escapedCode}</div>`
+  } else if (lang === 'graphviz') {
+    return `<div class="graphviz raw">${escapedCode}</div>`
+  } else if (lang === 'mermaid') {
+    return `<div class="mermaid raw">${escapedCode}</div>`
+  } else if (lang === 'abc') {
+    return `<div class="abc raw">${escapedCode}</div>`
+  }
+}
+
+module.exports = externals
--- a/public/js/lib/highlight/index.js
+++ b/public/js/lib/highlight/index.js
@@ -1,26 +1,43 @@
 'use strict';
 
-import S     from 'string'
 import hljs  from './hljs'
 import Prism from './prism'
+import externals from './externals'
+import {escapeHtml, addLineNumbers} from './utils'
 
 exports.listLanguages = function () {
   return [].concat(hljs.listLanguages()).concat(Prism.listLanguages())
 }
 
 exports.render = function (code, lang) {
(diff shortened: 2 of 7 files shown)


public/js/lib/highlight/externals.js
```diff
@@ -0,0 +1,30 @@
+'use strict';
+
+import {escapeHtml}  from './utils'
+
+let externals = {}
+
+externals.checkLang = function (lang) {
```
ccoenen: checkLang i.e. `lang` is not a good name for this, it implies that everything in there must be a language, which is a stretch for things lime mermaid. And might be completely wrong for plugins to come.

In the proposal in #20, I suggested "keyword" or "magic word" for that reason, but it might also be something completely different like "activator" (verbalised) or "type" (which is a bit too generic for my taste).

Also, the functionality is basically the same for all blocks that exist as of now. So it would be more fitting to make this an array (and have the `array#includes(keyword)` as part of the surrounding code, not for every individual highlighter.

public/js/lib/highlight/externals.js
```diff
@@ -0,0 +1,30 @@
...
+  if (typeof lang !== 'string') {
+    return false
+  }
+  const languages = ['sequence', 'flow', 'graphviz', 'mermaid', 'abc']
```
ccoenen: (basically we could just return this and have the surrounding code figure it out)

public/js/lib/highlight/prism.js
```diff
@@ -0,0 +1,53 @@
...
+require('prismjs/components/prism-go')
+require('prismjs/components/prism-rust')
+require('prismjs/components/prism-typescript')
+require('prismjs/components/prism-jsx')
+require('prismjs/components/prism-makefile')
+require('prismjs/components/prism-gherkin')
+
+import Prism from 'prismjs'
+
+const languages = ['haskell', 'go', 'rust', 'typescript', 'typescript', 'js', 'jsx', 'gherkin', 'tiddlywiki', 'mediawiki', 'wiki', 'cmake']
```
ccoenen: can't we get this list from prism somehow?

public/js/lib/highlight/prism.js
```diff
@@ -0,0 +1,52 @@
...
+
+Prism.listLanguages = function () {
```
ccoenen: We extend a foreign library? This seems counter-intuitive? And also I don't think it's needed?

public/js/lib/highlight/index.js
```diff
@@ -0,0 +1,43 @@
...
+  if (startnumber !== -1) {
+    return `<pre class="part"><code>${addLineNumbers(escapedCode, lang, startnumber)}</code></pre>`
+  }
+  return `<pre class="part"><code>${escapedCode}</code></pre>`
+}
+
+
+function highlight (code, lang) {
+  if (Prism.checkLang(lang)) {
+    return Prism.highlight(code, Prism.selectLang(lang))
```
ccoenen: I would prefer if this `highlight()` worked the same way all the others. Do. There's another comment below on monkey patching Prism itself, so the two go togeteher in a way.

SISheogorath:
I agree, we should maybe figure out some interface specification before implementing it :see_no_evil:

I think it started out as getting rid of the string library back then and then grew into this :sweat_smile:

I'll write some basic spec and provide it :) Should also make it into the dev docs :+1:

ccoenen:
Also, #20 may be a starting point for this. It is not the same part of the codebase or the same target, but the two go together, in my opionion.

A document extension as described there would consist of both rendering (view view) and highlight (editor view) and perhaps even auto-complete-hints.

davidmehren:
Closing this, as we will rewrite the highlight code anyway in the [new frontend](https://github.com/codimd/react-client).

PR #28, opened 2019-04-04
Anonymous views

SISheogorath:
Continuing @hakoai's work from https://github.com/hackmdio/codimd/pull/1152.

(We will close this PR in favor of one by the original author)


clvs7-gh:
Hello.

Do you have any update on this?
hackmdio's one seems already merged. I think it is safe to merge.

SISheogorath:
To be honest I'm not sure if we really want to do this. I see the reason why this might be useful, but I'm not really sure about the name of this option. I would prefer to put some time and effort into doing this right instead of merging this half-way through PR .-.

davidmehren:
If the concept still needs discussion, we should probably move this to an issue? This seems related to [the other](https://github.com/codimd/server/issues?q=is%3Aissue+is%3Aopen+sort%3Aupdated-desc+label%3Afeature%3Apermissions) permission issues.

davidmehren:
Closing, as this got pretty stale and we are developing a new permissions system anyway with 2.0.

PR #29, opened 2019-04-04
WIP: Add plantuml support

SISheogorath:
Continue work from @chulkilee from https://github.com/hackmdio/codimd/pull/1096

(We'll close this PR in favor of a PR by the author)


SISheogorath:
Docs are missing and yarn.lock shouldn't be part of the PR

ccoenen:
(for documentation)

In our matrix chat, we had a discussion about the rendering on a third party server. We find this to be unexpected from a user's point of view and would not want to have this be the default.

At least, the default to plantuml should _not_ be configured by default, so that an admin has to do this themselves. If an external service does the rendering, it should also be treated like other embeds (i.e. behind an opt-in banner) or be cached locally by an image-proxy.

davidmehren:
Moved to https://github.com/codimd/react-client/issues/324

PR #31, opened 2019-04-04
Hide port minio

SISheogorath:
Continuing work by @Thor77 from https://github.com/hackmdio/codimd/pull/1095

---

With the current state this behaviour is enabled by default and can be disabled by setting
```json
{
  "minio": {
    "hideDefaultPorts": false
  }
}
```

or `CMD_MINIO_HIDE_DEFAULT_PORTS=false`.


SISheogorath:
Docs are still missing and should be added before merging

PR #32, opened 2019-04-04
make aws s3 endpoint configurable

SISheogorath:
(Continuing work by @madddi from https://github.com/hackmdio/codimd/pull/1142)

### DIFF
--- a/lib/config/environment.js
+++ b/lib/config/environment.js
@@ -40,7 +40,8 @@ module.exports = {
   s3: {
     accessKeyId: process.env.CMD_S3_ACCESS_KEY_ID,
     secretAccessKey: process.env.CMD_S3_SECRET_ACCESS_KEY,
-    region: process.env.CMD_S3_REGION
+    region: process.env.CMD_S3_REGION,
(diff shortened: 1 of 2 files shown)


lhw:
This would make the minio config redundant I think if the endpoint would be allowed to supply a port and protocol. Would remove a good few lines from the config.

soulchild:
I don't know snyk, but it looks like this is only blocked because snyk didn't report back. What else is keeping this from getting released? We're using a CodiMD fork for months now, because somehow this IMHO uncritical pull request won't get merged. Is there anything I can help with to get this out of the door?

ccoenen:
I think mostly it's not merged because other things came up and the last release needed to be put out quickly [because of a security issue with pdf export](https://github.com/codimd/server/blob/master/public/docs/release-notes.md#-150--2019-08-15-0000).

The Snyk check has been added very recently - more recently than this pull request.

oupala:
The newly added variable is not documented yet: `CMD_S3_ENDPOINT`.

SuperSandro2000:
I created #349 to not forget this as I am not sure about the function of the env.

PR #44, opened 2019-04-12
Fix toolbar day mode and upload button

### LINKED ISSUE #23: Editor: Toolbar is dark in day mode
dark toolbar

The toolbar stays dark if the editor is changed to day mode.
I would expect it to turn light as well.

pferreir:
Fixes #23

I'm taking care of this one since @SISheogorath said it would be a nice issue to work on.

You'll notice I've enabled support for SASS/SCSS and moved the toolbar code to an SCSS file. I saw that  the loader was already present in the webpack config, so I guessed it would be OK to use it. I think it makes our life much easier when it comes to theming, but let me know if you'd like me to fall back to standard CSS.
Still, I believe it would be worth start splitting the CSS into smaller chunks.

### DIFF
--- a/public/js/lib/editor/index.js
+++ b/public/js/lib/editor/index.js
@@ -219,8 +219,8 @@ export default class Editor {
     makeComment.click(() => {
       utils.insertText(this.editor, '> []')
     })
+
     uploadImage.bind('change', function (e) {
-      console.log("tiggered")
       var files = e.target.files || e.dataTransfer.files
       e.dataTransfer = {}
       e.dataTransfer.files = files
--- a/public/js/lib/editor/index.js
+++ b/public/js/lib/editor/index.js
@@ -3,6 +3,8 @@ import config from './config'
 import statusBarTemplate from './statusbar.html'
 import toolBarTemplate from './toolbar.html'
 
+import '../../../css/ui/toolbar.css'
+
 /* config section */
 const isMac = CodeMirror.keyMap.default === CodeMirror.keyMap.macDefault
 const defaultEditorMode = 'gfm'
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -10,6 +10,7 @@ module.exports = {
     // wrong.
     "import/first": ["warn"],
     "indent": ["warn"],
+    "no-console": ["warn"],
     "no-multiple-empty-lines": ["warn"],
     "no-multi-spaces": ["warn"],
     "object-curly-spacing": ["warn"],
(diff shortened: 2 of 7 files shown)


public/js/lib/editor/index.js
```diff
@@ -219,8 +221,8 @@ export default class Editor {
     makeComment.click(() => {
       utils.insertText(this.editor, '> []')
     })
+
     uploadImage.bind('change', function (e) {
-      console.log("tiggered")
```
SISheogorath: Good catch! Looks like someone left some surgical tools…

pferreir: Is it OK if I add `no-console` to `.eslintrc`?
Actually, now that I look at `.eslintrc`, it does look a little bit minimal. Any plans to change that?

pferreir:
Anyway, commit removed!

public/js/lib/editor/index.js
SISheogorath: Yes, I think we should improve this. First step should be to get rid of the warnings (which will touch all files). If you want to go for it, feel free to go. Otherwise, I guess I'll do this before a release.

ccoenen:
I'm not a fan of sass. It is a _huge_ dependency, especially since it is a compiled one. I would rather not have introduced it over a change as small as this.

[node-sass dependency tree](http://npm.anvaka.com/#/view/2d/node-sass)

pferreir:
No problem, I can switch it back to CSS. But then we should probably remove
it from the webpack config.

SISheogorath:
@pferreir Agreed, feel free to throw it out there :) Looking forward to your changes!

pferreir:
OK, I'll have a look later.

pferreir:
Removed SASS and corresponding webpack config.

ccoenen:
This looks quite good. I think all problems raised in #23 are fixed. There's three organisational things left to do:

- yarn.lock should not be changed within the merge request (it causes merge conflicts frequently because the file is generated and has lots of changes all the time)
- all the commits will have to be signed-off. Most of them are fine, but the commit 93a9bdd is missing it.


light mode screenshot

dark mode screen shot

pferreir:
> all the commits will have to be signed-off. Most of them are fine, but the commit 93a9bdd is missing it.

That commit is @SISheogorath's and it's as it is in master. I can rebase to the latest master, though.

ccoenen:
Thanks for taking care of this :-)

PR #61, opened 2019-05-06
Add name of directory to clone into

archemiro:
The docker setup instructions, on the `git clone` step, are missing the name of the directory to clone into that is used in the next step: `cd codimd-container`.


SISheogorath:
Good catch! Thanks a lot!

PR #65, opened 2019-05-12
WIP drop node 6 support

ccoenen:
We will no longer test on node6 and instead focus on 8+. This won't break node6 immediately, but we will no longer go out of our way supporting a version that does not receive security updates.

Has been discussed and started in https://github.com/hackmdio/codimd/pull/1120

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,7 +2,7 @@
   "presets": [
     ["env", {
       "targets": {
-        "node": "6",
+        "node": "8",
         "uglify": true
       }
     }]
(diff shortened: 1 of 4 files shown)


ccoenen:
scrypt fails for node12 apparently.

    ../src/node-boilerplate/inc/scrypt_common.h:39:63: error: no matching function for call to ‘v8::Value::Uint32Value()’

seems to be the problem while compiling scrypt.

ccoenen:
closing in favor of #66

PR #66, opened 2019-05-13
Making Password hashing async / dropping node 6 support

ccoenen:
Basically I use this PR to get Travis to run this in various node versions.


ccoenen:
you're correct! I just tried on a fresh VM and the master branch fails at scrypt (intentionally omitting libssl-dev), and this updated branch runs `yarn` without a hitch. Works fine, too.

I changed the manual setup docs accordingly.

PR #68, opened 2019-05-14
add note to omit trailing slash on gitlab baseurl

mcnesium:


ccoenen:
I would like to request adding this note to the [config **file** document](https://github.com/codimd/server/blob/master/docs/configuration-config-file.md#gitlab-login) as well. You don't need to provide the whole table, but mentioning that the trailing slash should not be included for `gitlab.baseURL` would be nice!

davidmehren:
Hi @mcnesium!
It's been a long time since activity here, therefore I'm closing this.
If you still have interest in contributing, feel free to reopen.

PR #81, opened 2019-05-25
Fix CodeMirror toolbar hiding content

### LINKED ISSUE #80: Portion of the last text line hidden by the bottom editor bar
The bottom half of the last line of text is hidden by the bottom bar when using the `Both` view and the `Edit` view

image

SISheogorath:
As it may happens that the codemirror content flows underneath the
status bar, this patch should help to avoid it. It adds the size of the
status bar as margin-bottom so the codemirror window itself is forced
above the statusbar.

Fixes #80

### DIFF
--- a/public/css/index.css
+++ b/public/css/index.css
@@ -106,7 +106,10 @@ body.night{
     color: #78B2F2 !important;
 }
 .CodeMirror-sizer {
-    margin-bottom: 0px !important;
+    /* Make sure CodeMirror doesn't hide text under the status bar
+     * 26px is the height of the status bar.
+     */
+    margin-bottom: 26px !important;
 }
 .CodeMirror-insert-match {
   background: lawngreen;


public/css/index.css
```diff
@@ -106,7 +106,7 @@ body.night{
     color: #78B2F2 !important;
 }
 .CodeMirror-sizer {
-    margin-bottom: 0px !important;
+    margin-bottom: 26px !important;
```
ccoenen: magic number! This needs either a comment what the 26px refer to.

SISheogorath: Git blame would tell, but let me add a comment here :)

ccoenen: In that direction, yes. But I'm thinking the other way round: If someone were to modify the status bar, they would never find `.CodeMirror-sizer` or "all things 26px". Coming from that other angle, git blame won't help.

PR #94, opened 2019-05-30
Fix hidden MathJax output

### LINKED ISSUE #91: Equations are not rendered correctly
I found that codimd cannot render complex latex equations, for example, the following latex 
```
$$
\alpha
$$
```
can be rendered correctly, while
```
$$
\alpha\beta
$$
```
cannot be rendered, (rendered equation disappear immediately, then invisible forever)

SISheogorath:
In order to have a better experience when linking to headlines based on
their ID, a patch[1] introduced a new CSS contruct to add some space in
front of HTML tags with an id field. Therefore they would no longer be
hidden by a visible navbar.

This cause a regression bug by moving the rendered mathjax out of its
visible area. This patch fixes the problem by restricting the previous
change to headlines only.

[1]: commit 388f034750a1a3b11aa49b88f3589486e9b5de76

Fixes #91


ccoenen:
I think you referenced the wrong commit, but other than that it looks good to me.

The commit in question might have been in this pull request instead: https://github.com/codimd/server/pull/77/files

SISheogorath:
Huh! You are right!

PR #97, opened 2019-05-30
Fix eslint warnings

SISheogorath:
Since we are about to release it's time to finally fix our linting. This
patch basically runs eslint --fix and does some further manual fixes.
Also it sets up eslint to fail on every warning on order to make
warnings visable in the CI process.

There should no functional change be introduced.


ccoenen:
I'm not comfortable sneaking this in a few hours before a release. This touches a ton of files, even if 95% of it is just white space. I did review it, and it "looks" ok, but I have no way of knowing if I didn't miss something.

ccoenen:
The diff without white spaces actually is not _that_ long. "it's probably fine". Still, not a big fan of last-minute large patch.

SISheogorath:
After shower I thought the same. Let's make it one of the first patches for 1.4.1 :)

PR #99, opened 2019-06-02
Update app.js - HSTS Syntax Update

ndom91:
Updated spelling of `includeSubdomain` to `includeSubDomain`.

This was creating unnecessary error messages on launch as `hsts` had deprecated the old spelling.

image

### DIFF
--- a/app.js
+++ b/app.js
@@ -84,7 +84,7 @@ app.use(compression())
 if (config.hsts.enable) {
   app.use(helmet.hsts({
     maxAge: config.hsts.maxAgeSeconds,
-    includeSubdomains: config.hsts.includeSubdomains,
+    includeSubDomains: config.hsts.includeSubDomains,
     preload: config.hsts.preload
   }))
 } else if (config.useSSL) {


ccoenen:
Thanks for contributing this. There is two ways this configuration can be set: config file or env-vars, so these would also have to be corrected. Ideally, we should make sure people update their configurations in case they are using this setting (we've been discussing if this is used by anyone, so far very few people seem to be changing this setting).

Here's some places that also would need to be updated:

https://github.com/codimd/server/blob/678d1396916523effd393791238879c0cac15e05/lib/config/default.js#L17

The env vars in the next ones would also have to be renamed from 
`...INCLUDE_SUBDOMAINS` to
`...INCLUDE_SUB_DOMAINS` for consistency:

https://github.com/codimd/server/blob/678d1396916523effd393791238879c0cac15e05/lib/config/environment.js#L18

https://github.com/codimd/server/blob/36c083277e6c74bb6eac495971522e6932cfc207/lib/config/hackmdEnvironment.js#L14


The documentation needs to change in these places:

https://github.com/codimd/server/blob/678d1396916523effd393791238879c0cac15e05/docs/configuration-env-vars.md#csp-and-hsts

https://github.com/codimd/server/blob/678d1396916523effd393791238879c0cac15e05/docs/configuration-config-file.md#csp-and-hsts

PR #108, opened 2019-06-10
WIP: Switch to upstream CodeMirror

SISheogorath:
Since some changes by HackMD broke future deployments of CodiMD by
changing package names, this change starts a transition from moving
CodiMD from the HackMD CodeMirror fork, back to the upstream version of
CodeMirror in order to simplify maintenance and follow the
upstream-first philosophy.

This patch moves various changes from the HackMD CodeMirror version to
`public/vendor/codemirror-hackmd` and integrate these changes with the
upstream CodeMirror version. Further we introduce an addons.js file
which will take care of including all addons to load them with webpack.


davidmehren:
Closing this, as it seems to be too big of a change for 1.6.1 and no longer relevant for 2.0 as the [new frontend](https://github.com/codimd/react-client) already uses upstream codemirror.

PR #109, opened 2019-06-11
Allow arbitary file upload

zawlin:
* This is only tested with `filesystem` image upload type.

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -181,6 +181,7 @@ switch (config.imageUploadType) {
     break
   default:
     config.allowedUploadMimeTypes = [
+      '*',
       'image/jpeg',
       'image/png',
       'image/jpg',
--- a/public/vendor/inlineAttachment/inline-attachment.js
+++ b/public/vendor/inlineAttachment/inline-attachment.js
@@ -304,6 +304,8 @@
           var replacements = [];
           var string = this.settings.progressText.replace(this.filenameTag, id);
           var lines = this.editor.getValue().split('\n');
+          this.settings.urlText = (this.origFile.type.includes("image")?"!":"")+"["+this.origFile.name+ "]({filename})";
+
         var newValue = this.settings.urlText.replace(this.filenameTag, filename);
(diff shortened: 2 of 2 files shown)


lib/config/index.js
```diff
@@ -181,6 +181,7 @@ switch (config.imageUploadType) {
     break
   default:
     config.allowedUploadMimeTypes = [
+      '*',
```
SISheogorath: We should be careful about this. I would still recommend to limit the uploads a bit to "passive" documents. Just putting a wildcard here, appears a bit too broad to me.

The alternative would be to blacklist mimetypes, but that makes things even more difficult.

I also think we need to add a case for lutim in here as I don't expect lutim to support generic files.

public/vendor/inlineAttachment/inline-attachment.js
```diff
@@ -345,6 +347,7 @@
    * @return {Void}
    */
   inlineAttachment.prototype.onFileInserted = function(file, id) {
+    this.origFile =file;
```
SISheogorath: There is a space missing in front of "file".

public/vendor/inlineAttachment/inline-attachment.js
```diff
@@ -304,6 +304,8 @@
           var replacements = [];
           var string = this.settings.progressText.replace(this.filenameTag, id);
           var lines = this.editor.getValue().split('\n');
+          this.settings.urlText = (this.origFile.type.includes("image")?"!":"")+"["+this.origFile.name+ "]({filename})";
```
SISheogorath: Let's add some spaces in here in order to improve readablity

zawlin:
Ok I will spend some time this weekend. I will test with other storage providers and see how it can work out. It also should be configurable I guess from the main config file.

davidmehren:
Hi @zawlin!
Are you still interested in working on this?

InnayTool:
I think we can close this PR, it seems to be abandoned.

davidmehren:
I agree. If anybody is interested in working on this, please let us know!

PR #114, opened 2019-06-13
Fix some minor quirks in the LinuxServer.io docs

SISheogorath:
The current documents might end up confusing people and are not
completely accessible. This minor fixes should clear up the situation
and add alt texts to all badges, explain the links at the end of the
docs, and list LinuxServer.io in the supported provider section of the
README.

Some reasoning on the change in the listing:
Since we maintain an own container image which is for sure kept updated
on release, this is our first listing, as well as general solutions that
are build on that image, like the K8s integration.

The next listings are integrated provides which allow self-hosting, like
Cloudron and I also consider LinuxServer.io as this kind of providers.
Which try to enable people to run CodiMD on their own hardware or rented
servers in a very easy way, but by using their own images.

As third category I would look at hosted offers, like Heroku, which are
not completely SaaS but far enough away from the self-hostability that
I consider them as an own category. PaaS-based solutions are not as
FOSS-style as we want our setups to be, but of course still supported.

Finally the manual setup. We keep it down here, because we support it,
but don't recommend it in general. It's hard to upgrade and can cause
problems when dependencies are not correctly updated or people don't run
the db migrations.

/cc @CHBMB

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,9 +37,9 @@ You can run CodiMD in a number of ways, and we created setup instructions for
 all of these:
 
 * [Docker](docs/setup/docker.md)
-* [Docker (LinuxServer.io)](docs/setup/docker-linuxserver.md)
 * [Kubernetes](docs/setup/kubernetes.md)
 * [Cloudron](docs/setup/cloudron.md)
+* [LinuxServer.io (multi-arch docker)](docs/setup/docker-linuxserver.md)
 * [Heroku](docs/setup/heroku.md)
 * [Manual setup](docs/setup/manual-setup.md)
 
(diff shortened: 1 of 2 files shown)


CHBMB:
Looks good to me, this project is your baby, whatever is best, I'm good with. :+1:

docs/setup/docker-linuxserver.md
```diff
@@ -1,14 +1,14 @@
...
 
-[LinuxServer.io](https://linuxserver.io) have created an Ubuntu based multiarch container for x86-64, arm64 and armhf which supports PDF export from all architectures using PhatomJS. 
+[LinuxServer.io](https://linuxserver.io) have created an Ubuntu-based multiarch container image for x86-64, arm64 and armhf which supports PDF export from all architectures using PhatomJS.
```
CHBMB: > have created an Ubuntu-based multiarch container image

Should probably be 

> have created an Ubuntu-based multiarch image

docs/setup/docker-linuxserver.md
```diff
@@ -1,14 +1,14 @@
...
 
 It gets rebuilt on new releases from CodiMD and also weekly if necessary to update any other package changes in the underlying container, making it easy to keep your CodiMD instance up to date.
 
 It also details how to easily [utilise Docker networking to reverse proxy](https://github.com/linuxserver/docker-codimd/#application-setup) CodiMD using their [LetsEncrypt docker image](https://github.com/linuxserver/docker-letsencrypt)
 
-[Github Repository](https://github.com/linuxserver/docker-codimd/)
-[Docker Hub Image](https://hub.docker.com/r/linuxserver/codimd)
\ No newline at end of file
+In order to contribute check the LinuxServer.io [GitHub repository](https://github.com/linuxserver/docker-codimd/) for CodiMD.
+And to find all tags and versions of the container image, check the [Docker Hub repository](https://hub.docker.com/r/linuxserver/codimd).
```
CHBMB: > And to find all tags and versions of the container image
Should probably read
> And to find all tags and versions of the image

README.md
```diff
@@ -37,9 +37,9 @@ You can run CodiMD in a number of ways, and we created setup instructions for
 all of these:
 
 * [Docker](docs/setup/docker.md)
-* [Docker (LinuxServer.io)](docs/setup/docker-linuxserver.md)
 * [Kubernetes](docs/setup/kubernetes.md)
 * [Cloudron](docs/setup/cloudron.md)
+* [LinuxServer.io (container-based)](docs/setup/docker-linuxserver.md)
```
CHBMB: > [LinuxServer.io (container-based)](docs/setup/docker-linuxserver.md)

should probably read

> [LinuxServer.io (Docker Image)](docs/setup/docker-linuxserver.md)

CHBMB:
Docker purists get upset when you confuse the term image and container, I'm terrible at this and my original documentation was flawed.  Sorry about that.  But if you want to be technically correct I've made some suggestions.

README.md
SISheogorath: There is mainly one reason why I try to avoid the word "Docker" in here: https://www.youtube.com/watch?v=PQaz2PDW4MM&t=1m29s (also the reason why the repository is called container, not docker)

PR #128, opened 2019-07-01
DB URL: Secret File Support

dargmuesli:
Adds a possibility to read the db URL from a file, that could e.g. be provided by a [Docker secret].(https://docs.docker.com/engine/swarm/secrets/).
See https://github.com/codimd/container/pull/36 too.

### DIFF
--- a/lib/config/dockerSecret.js
+++ b/lib/config/dockerSecret.js
@@ -13,6 +13,7 @@ function getSecret (secret) {
 
 if (fs.existsSync(basePath)) {
   module.exports = {
+    dbURL: getSecret('dbURL'),
     sessionsecret: getSecret('sessionsecret'),
     sslkeypath: getSecret('sslkeypath'),
     sslcertpath: getSecret('sslcertpath'),


SISheogorath:
We already have support for docker secrets, you might want to extend it to cover the database URL as well :)

https://github.com/codimd/server/blob/master/lib/config/dockerSecret.js

dargmuesli:
Well, that would be the easy way to do it 🤯

dargmuesli:
It's off-topic, but would it be ok for me to add a prefix (`codimd_`) to the secret names so that one can differentiate between multiple `sessionsecret`s in a stack with multiple services for example, keeping the old naming too?

SISheogorath:
Mhm, for that Docker has an own syntax, which is the reason why we didn't go for prefixes: https://docs.docker.com/compose/compose-file/#long-syntax-2

dargmuesli:
Today seems to be the day on which I nudge every edge of my Docker knowledge.
image
Thanks for telling me!

dargmuesli:
But wait. Do I understand it correctly that CodiMD assumes a e.g. `/run/secrets/sessionsecret` to exist? What does the possibility in Docker change about the issue that another service might assume such a secret for itself too?

SISheogorath:
Yes, CodiMD checks if `/run/secrets/sessionsecret` exists and loads it. But it shouldn't conflict within docker there should be only one process/service i.e. codimd per container. Which also means when another service uses the same name, you just don't assign this secret to codimd (you shouldn't anyway).

dargmuesli:
Oh well, yep you're perfectly right. Sorry, maybe two hours of sleep weren't really enough today. Don't know why I'm thinking so overcomplicated 🥴

SISheogorath:
No problem :) It's good to see this feature used, I proposed it pretty much at the very beginning of my time contributing to this project but had no swarm around to actively use it. :+1: Go go go! :rocket:

SISheogorath:
thanks a lot!

PR #168, opened 2019-09-02
Config: Return String Instead Of Buffer For Docker Secrets

dargmuesli:
Prevents "TypeError: Cannot freeze array buffer views with elements".


SISheogorath:
We might need another function call here: https://nodejs.org/api/all.html#fs_fs_readfilesync_path_options

dargmuesli:
Ah yes, sure. I never really worked with node before, just for fixes like this :)

SISheogorath:
For that you do a great job 👍🏼

dargmuesli:
You do a great job too with giving a direction to work to with links to the documentation and not just giving the solution! That's how one learns.

PR #170, opened 2019-09-04
Added endpoint for note-creation with given alias

ErikMichelson:
Added a new endpoint `/new/:noteId` that allows creation of a note by HTTP-POST with the specified alias as noteId.

### DIFF
--- a/lib/response.js
+++ b/lib/response.js
@@ -41,12 +41,12 @@ var response = {
   errorServiceUnavailable: function (res) {
     res.status(503).send("I'm busy right now, try again later.")
   },
-  newNote: newNote,
   showNote: showNote,
   showPublishNote: showPublishNote,
   showPublishSlide: showPublishSlide,
   showIndex: showIndex,
   noteActions: noteActions,
+  postNote: postNote,
   publishNoteActions: publishNoteActions,
   publishSlideActions: publishSlideActions,
   githubActions: githubActions,
@@ -107,8 +107,7 @@ function responseCodiMD (res, note) {
   })
 }
 
-function newNote (req, res, next) {
-  var owner = null
+function postNote (req, res, next) {
   var body = ''
   if (req.body && req.body.length > config.documentMaxLength) {
     return response.errorTooLong(res)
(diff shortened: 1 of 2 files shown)


lib/response.js
```diff
@@ -123,10 +123,10 @@ function newNote (req, res, next) {
   }
   models.Note.create({
     ownerId: owner,
-    alias: req.alias ? req.alias : null,
+    alias: req.alias ? req.alias : (config.allowFreeURL ? (req.params.alias ? req.params.alias : null) : null),
```
SISheogorath: This line is rather complex to read and also doesn't comply with the logic we use in another place:

https://github.com/codimd/server/blob/529075fd67ad0d41ba0b25a7de69817dd4cd76fb/lib/response.js#L161-L164

We should might do a proper abstraction of the code above and use it for both cases :)

lib/response.js
```diff
@@ -123,10 +123,10 @@ function newNote (req, res, next) {
...
     content: body
   }).then(function (note) {
-    return res.redirect(config.serverURL + '/' + models.Note.encodeNoteId(note.id))
+    return res.redirect(config.serverURL + '/' + (note.alias ? note.alias : models.Note.encodeNoteId(note.id)))
```
SISheogorath: I think we can still redirect to the noteId even when the note alias exists. This otherwise just makes everything more complicated.

lib/response.js
ErikMichelson: Sounds good, I'm working on it in the next few days.

lib/response.js
ErikMichelson: In which matter does it make everything more complicated? I would see the redirect as a form of confirmation that the import succeeded. When the user is redirected to the noteId instead of the alias, he might assume that the alias was not properly set.

SISheogorath: Mhm, good point, @ccoenen what's your input on this?

ccoenen: The alias'd version takes precedence in other places, so we should do so here as well.

(the following lines are just a few examples I found quickly)

https://github.com/codimd/server/blob/529075fd67ad0d41ba0b25a7de69817dd4cd76fb/lib/response.js#L185

https://github.com/codimd/server/blob/529075fd67ad0d41ba0b25a7de69817dd4cd76fb/lib/response.js#L202

https://github.com/codimd/server/blob/529075fd67ad0d41ba0b25a7de69817dd4cd76fb/lib/response.js#L249

https://github.com/codimd/server/blob/529075fd67ad0d41ba0b25a7de69817dd4cd76fb/lib/response.js#L253

ErikMichelson:
I refactored the code for creating the note. Regarding documentation, I'm going to write an OpenAPIv3 conform documentation of the CodiMD endpoints (as mentioned in #172) and would include this new endpoint in it.

SISheogorath:
I can't give enough thumbs up to this statement! :+1: :+1: :+1: 

I really enjoy your work!

SISheogorath:
@ErikMichelson do you want to document this as a separate PR?

ErikMichelson:
Yes, I'm going to open a separate PR for documentation later. I hadn't the time to work on this yet, but in the next weeks I'll do so.

PR #171, opened 2019-09-05
Move sequelize-cli from devDependencies to dependencies, because it is needed to run migrations at run-time

### LINKED ISSUE #161: sequelize-cli is a devDependency, when it should be a dependency
The [docker-entrypoint.sh](https://github.com/codimd/container/blob/master/resources/docker-entrypoint.sh#L27) script from the [codimd/container](https://github.com/codimd/container) repository runs `sequelize-cli` during startup to run pending migrations. At that point in time, the `sequelize` script provided by `sequelize-cli` is already gone, because it is listed as a devDependency and gets removed by the [subsequent call](https://github.com/codimd/container/blob/master/debian/Dockerfile#L66) to `yarn install` in the `Dockerfile`, which prunes all non-devDependencies (because `NODE_ENV=production`).

I propose moving `sequelize-cli` to the dependencies, because it is used at runtime.

soulchild:
Fixes #161


soulchild:
Well, that didn't quite work. 😞

Can you manually retrigger the reviewing process? I got rid of my last commit and pushed a new version with just the `package.json` updated.

SISheogorath:
You can do so by doing `git stash && git commit --amend --no-edit && git push -f && git stash pop`

This will:

1. Get your changes out of the way
2. Redo your last commit
3. Force push it here
4. restore your changes

But since you use the master branch you might need to check your git repository to allow force pushing to the master branch.

In future I would recommend to use a separate branch (created by `git checkout -b myNewBranchName`) since this makes it easier to develop for you :)

SISheogorath:
Thanks a lot for you work!

PR #176, opened 2019-09-10
Add security note to repository

SISheogorath:
In order to simplify the communication with security researcher and
allow reporting of issues, this document should provide a rough idea
about:

1. What versions are supported
2. Who to contact
3. How to send findings properly secured
4. What to expect from an approved security issue
5. What if it's not considered a security issue


SECURITY.md
```diff
@@ -0,0 +1,32 @@
+# Security Policy
+
+## Supported Versions
+
+Only the latest release of CodiMD is supported. We don't have the
```
ccoenen: I would even argue that even _with_ enough resources there would be little to no benefit in supporting legacy versions.

PR #180, opened 2019-09-18
Fixed #179 (redirect loop with a trailing slash)

### LINKED ISSUE #179: Trying to open /build causes a redirect loop
When someone tries to open /build as URL path, this causes a redirect loop.

ErikMichelson:
Requests for `/build`, `/uploads` and so on will not result in a redirection loop anymore but respond with an HTTP 404.

Maybe we should add specific routes for these directories that return an HTTP 403?

### DIFF
--- a/app.js
+++ b/app.js
@@ -126,9 +126,9 @@ app.use(i18n.init)
 
 // routes without sessions
 // static files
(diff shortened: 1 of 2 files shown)


SISheogorath:
I think 404 is fine. We should might add them to forbidden note names to prevent them from overwriting those paths in free url mode

PR #184, opened 2019-09-25
Fix undefined function

SuperSandro2000:


SISheogorath:
Good catch! But to be honest, we should get rid of this part of the code entirely. It's code used for backwards compatibility of the frontend from around a year ago(?)

This should no longer run anywhere and therefore this code section shouldn't be needed at all :)

Feel free to update the PR to remove it.

SuperSandro2000:
> This should no longer run anywhere and therefore this code section shouldn't be needed at all :)

I am afraid that it does. It crashed a server I maintain multiple times and I have no clue why.

SISheogorath:
Fixed by #187

Issue #58, opened 2019-04-20
Remove legacy code path for noteId in socket io connections


ccoenen:
socket-io used to use the referer field to determine the note-id. This was changed over a year ago towards the query part of the url. A backwards compatibility was left in place. I think it's time to remove that.

https://github.com/codimd/server/blob/321114db120c55885bee4dcbb66134f305733be6/lib/realtime.js#L280-L289

This was originally discussed in https://github.com/hackmdio/codimd/issues/623 and changed in https://github.com/codimd/server/commit/c4f8fb78eeef1c9536caae4c8cced4f0f50f1932

Amolith:
I can take a look at this one!

PR #187, opened 2019-10-03
remove legacy code to solve #58

Amolith:
Did about half an hour of testing before class; from what I can tell, everything still works fine!

Signed-off-by: Amolith <amolith@nixnet.xyz>


SISheogorath:
Nice! I guess this will also solve #184

Amolith:
> I guess this will also solve #184

I didn't even notice that one lol. I just saw that the build was failing and looked at the output from Travis CI and removed the offending line

SISheogorath:
Thanks a lot!

PR #188, opened 2019-10-03
Remove edit URL from read-only note

ritesh-pandey:
Remove edit URL from read-only note

Signed-off-by: Ritesh Pandey <ritpan2003@gmail.com>

### DIFF
--- a/public/views/pretty.ejs
+++ b/public/views/pretty.ejs
@@ -42,7 +42,7 @@
                 &nbsp;<span class="text-uppercase ui-status-lastchange"></span>
                 <span class="ui-lastchange text-uppercase" data-createtime="<%- createtime %>" data-updatetime="<%- updatetime %>"></span>
             </span>
-            <span class="pull-right"><%- viewcount %> views <a href="#" class="ui-edit" title="Edit this note"><i class="fa fa-fw fa-pencil"></i></a></span>
+            <span class="pull-right"><%- viewcount %> views </span>
             <br>
             <% if(ownerprofile && owner !== lastchangeuser) { %>
             <span class="ui-owner">


ErikMichelson:
If a note is published, it's url will be the short-id of the note. But if FreeURL mode is enabled and the note has an alias set, the published url equals to the alias. In this case someone already has the link to the edit page, so the removal of the edit link is not enough. Therefore we should maybe change the publish function so that it always redirects to the short-id of the note.

ErikMichelson:
One additional thing: The route `/s/example-note/edit` should also be removed otherwise this whole change would be useless.

see [here](https://github.com/codimd/server/blob/master/lib/response.js#L454-L456) and [here](https://github.com/codimd/server/blob/master/lib/response.js#L440-L442)

SISheogorath:
I actually considered this more of a UX and cosmetic than a functional change.

This link shouldn't be remove entirely. But when the user has no edit access to the note.

Removing the `/s/<shortid>/edit` endpoint doesn't make sense, since you can also access the note using /<shortid>`.

As written in the corresponding issue, please make it a conditional, not just a removal of code.

ritesh-pandey:
I am new to this project. Hence I am unable to understand the scope of this feature. This is requiring multiple commits. I should have used a draft PR here. But the feature to convert a PR to draft PR is not available yet. Please suggest how to take this ahead.

ErikMichelson:
Maybe we should overthink the whole permission system before finalizing this PR/issue. There are great ideas of new concepts like groups or extensive tagging. These changes need to result in a change of our current permission system. Therefore it would be double work if we implement the conditions here now, or?

davidmehren:
As we are planning to overthink the permissions system in the 2.0 release, I'm closing this for now.

PR #189, opened 2019-10-03
Added Czech translation

PetrTodorov:
Added Czech Translation.


SISheogorath:
Cool! We should definitely add this. Feel free to join the translation group at https://translate.codimd.org :)

I'll import the translations from here as soon as it's merged!

Issue #40, opened 2019-04-08
Add opengraph metadata


SISheogorath:
Collaborative markdown note taking is what we do. This definitely includes sharing links to other people. Be it on social media, in chats, email and more. A lot of those media implement a link preview, that provides a short abstract, a title and maybe a picture that appears on the linked page. The data used for those previews is usually obtained using the [Open Graph protocol](http://ogp.me/).

CodiMD doesn't provide any of those metadata for now, which causes some link preview implementations to simply take the title attribute, the first text that appears and the first picture in the HTML to create this preview.

![example of a preview in Riot](https://user-images.githubusercontent.com/8719867/55759880-d5adbb80-5a5a-11e9-9e87-663029f09204.png)

In order to make the information shown in the link preview more valuable, it would be great to implement the Open Graph protocol for CodiMD.

Points to start with would be:

https://github.com/codimd/server/blob/c6384567b8cf51edac6b3056b8c5b4139b86bd20/public/views/index/head.ejs#L7-L9

and:

https://github.com/codimd/server/blob/c6384567b8cf51edac6b3056b8c5b4139b86bd20/public/views/codimd/head.ejs#L6-L7

If you want to build it nice and fancy, you may also want to create an abstract version of it in the [shared views directory](https://github.com/codimd/server/tree/master/public/views/shared).

Finally, if you want to rock it completely, feel free to add some [yaml metadata](https://demo.codimd.org/yaml-metadata#) to modify it, but that's optional :)

https://github.com/codimd/server/blob/ee725dc58c89f1d9662f14a791a34cc93e2409ff/lib/response.js#L97-L99

If you have any questions, feel free to reach out here or in [our community channel](https://riot.im/app/#/room/#codimd:matrix.org). Looking forward to your contributions! :rocket:

NataliaSuarez:
Hi! Can I work in this issue?

ccoenen:
That would be great! Feel free to get started, and in case you have questions, you can post them here or on the commuinity channel :-)

ccoenen:
@NataliaSuarez did you get a chance to try this one? If you have questions (or if you dropped working on it) just let us know :-)

PR #191, opened 2019-10-04
Add customizable opengraph metadata for notes (see #40)

ErikMichelson:
Adds fixed opengraph metadata to the CodiMD index page and modifiable opengraph metadata to the notes. By default, the title and description of the note will be used, but with the YAML frontmatter this can be overridden.
See feature request #40

### DIFF
--- a/lib/models/note.js
+++ b/lib/models/note.js
@@ -409,8 +409,17 @@ module.exports = function (sequelize, DataTypes) {
       if (meta.GA && (typeof meta.GA === 'string' || typeof meta.GA === 'number')) { _meta.GA = meta.GA }
       if (meta.disqus && (typeof meta.disqus === 'string' || typeof meta.disqus === 'number')) { _meta.disqus = meta.disqus }
       if (meta.slideOptions && (typeof meta.slideOptions === 'object')) { _meta.slideOptions = meta.slideOptions }
+      if (meta.opengraph && (typeof meta.opengraph === 'object')) { _meta.opengraph = meta.opengraph }
     }
     return _meta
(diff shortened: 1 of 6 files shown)


SISheogorath:
Great work! Could we get a default picture for notes pointing to a useful image like this one?

https://github.com/codimd/server/blob/master/public/codimd-icon-1024.png

Original reason to fix this, was that by default most og:image using websites, simply defaulted to the first picture shown anywhere in the page and that often ended up as being a smiley.

ErikMichelson:
Well, as Open Graph protocol does not support relative urls, we should think about which image url we should use by default. There are three options: 
1. We use the codimd-icon-1024.png from the demo instance. The problem is that this requires the demo-instance to stay online under the same url. Furthermore there will be increased load on the server by all open graph requests hitting it.
2. We use the [icon in this repository](https://raw.githubusercontent.com/codimd/server/master/public/codimd-icon-1024.png). For this we have to assume that the file will keep staying in master branch of this repository. GitHub may not change the path of the raw files too.
3. We use the icon file on the note's home-server. Therefore we would need to make the ejs template aware of its absolute url (data from the http request) and add the icon filename, because `CMD_DOMAIN` is not always set.

SISheogorath:
I would say we take solution for:

We use `<%- serverURL %>` as prefix. We do this everywhere else anyway. And people who don't set `CMD_DOMAIN` properly :woman_shrugging: not our problem.

There is a warning on startup when this variable is not set properly it can break things. When people decided to ignore this warning, we can't help them.

PR #193, opened 2019-10-06
Fixed some typos in polish translation

mhajder:


SISheogorath:
Can you submit them through our translation platform, please? https://translate.codimd.org

While it sounds like a hassle in first place, it helps us to keep everything more maintainable, when we keep some basic flows, thanks :)

Issue #172, opened 2019-09-08
Add OpenAPIv3 documentation


ErikMichelson:
It would be great to have an [OpenAPIv3](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md) conform documentation of the already existing REST endpoints like `/new`, `/{note}/info`, `/status` and so on.

ccoenen:
I believe @SISheogorath has started working on this already?

SISheogorath:
I started something, but never really finished the work on it. Would be nice when people would work on that. The original idea was to create a proper (versioned) REST API under the `/api/`, but maybe just documenting existing endpoints could already help.

SISheogorath:
Thanks for your work @ErikMichelson!

PR #200, opened 2019-10-12
Add API documentation (#172)

ErikMichelson:
Adding documentation for the available REST endpoints in markdown and OpenAPIv3 format.
Some endpoints are not included. These are mainly login (and callbacks)/logout and quite internal things.
See #172

### DIFF
--- a/docs/dev/openapi.yml
+++ b/docs/dev/openapi.yml
@@ -1,0 +1,194 @@
+openapi: 3.0.1
+
+info:
+  title: CodiMD
+  description: CodiMD is an open source collaborative note editor. Several tasks of CodiMD can be automated through this API.
+  version: 1.6.0
+  contact:
+    name: CodiMD on GitHub
+    url: https://github.com/codimd/server
+  license:
+    name: AGPLv3
+    url: https://github.com/codimd/server/blob/master/LICENSE
+
+externalDocs:
+  url: https://github.com/codimd/server/blob/master/docs/dev/api.md
+
+
+paths:
+
+  /new:
+    get:
+      tags:
+        - note
+      summary: Creates a new note.
+      description: A random id will be assigned and the content will equal to the template (blank by default). After note creation a redirect is issued to the created note.
+      responses:
+        default:
+          description: Redirect to the new note
+    post:
+      tags:
+        - note
+      summary: Imports some markdown data into a new note.
+      description: A random id will be assigned and the content will equal to the body of the received HTTP-request.
+      requestBody:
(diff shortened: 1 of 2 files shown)


docs/dev/openapi.yml
```diff
@@ -188,7 +188,294 @@ paths:
...
+    post:
+      tags:
+        - user
+      summary: Replace user's history with a new one.
+      description: The body must be form-encoded and contain a field `history` with a JSON-encoded array like its returned from the server when exporting the history.
+      requestBody:
+        required: true
+        content:
+          'application/x-www-form-urlencoded':
+            example: 'history=[{"id":"example","text":"Untitled","time":1556275442010,"tags":[],"pinned":false]'
```
SISheogorath: The JSON here is invalid

docs/dev/openapi.yml
```diff
@@ -0,0 +1,481 @@
...
+          in: path
+          required: true
+          description: The note for which the pinned state should be toggled
+          content:
+            'text/plain':
+              example: my-note
+    delete:
+      tags:
+        - user
+      summary: Deletes a note from the user's history.
```
SISheogorath: Either this or the rest of this section is wrong, since this talks about note deletion, the rest about pinned notes.

ErikMichelson: I just tested again the requests made by the frontend. A POST request to the endpoint needs the pinned=true/false body, a DELETE request removes the note from the history. Omitting the pinned=true/false body while doing a POST request to the endpoint does not create the history entry but returns a HTTP 500. I'll investigate the code behind this.

ErikMichelson: The methods being called by the history-router are `historyDelete` and `historyPost` in the lib/history.js.

The [`historyPost`](https://github.com/codimd/server/blob/ce99a79352ea145cfbfc579592b96e53ba843b66/lib/history.js#L154-L170) method differentiates between the endpoints with note id and without. In the part for the endpoint with the note id only the toggle for the pinned state will be executed.

The [`historyDelete`](https://github.com/codimd/server/blob/master/lib/history.js#L184-L194) method differentiates similar and in the relevant part for this endpoint it just deletes the note from the user's history.

SISheogorath: We should really work on a proper `/api/` endpoint collection :see_no_evil: 

Thanks a lot for your investigative work here!

ErikMichelson: Yes we should. But documenting the current status is at least something.

Issue #175, opened 2019-09-10
Download HTML file links to Cloudflare and Google-font


popindavibe:
Hello,

When I download a CodiMD note as an HTML page, it uses Cloudflare CDN and import Google-font from the generated html file.

I was (wrongly?) expecting the html file to keep the CSS properties of the page itself (which does not source anything from outside). 
Testing by replacing Cloudflare link ref with my domain `https://<domain>/build/...` seems to display fine. Same goes for Google font import, replaced by `https://<domain>/build/font-pack.css`.

SISheogorath:
Good point, let's work on solving it :)

If you want to start, here you go:

https://github.com/codimd/server/blob/master/public/views/html.hbs

This is the template that is used by the client side function to render the page that is downloaded.

https://github.com/codimd/server/blob/c6e4f3757e16e39f8f1f75c898eb20eb2eb9db4a/public/js/extra.js#L650-L681

jadjay:
Same on **pretty** and **slide**

SISheogorath:
Pretty and Slide respect the `useCDN` setting:

https://github.com/codimd/server/blob/c6e4f3757e16e39f8f1f75c898eb20eb2eb9db4a/public/views/pretty.ejs#L20-L30

https://github.com/codimd/server/blob/c6e4f3757e16e39f8f1f75c898eb20eb2eb9db4a/public/views/slide.ejs#L18-L29

popindavibe:
Thanks @SISheogorath, I'll see what I can do, looks easy enough.

popindavibe:
ok I take that back, it's out of my league :-/
But I agree to get rid of the CDN altogether \o/

Along the same line of idea, the Googleapi import of SourceSansPro font seems redundant, as it is already part of the repo. I'm talking about `public/css/google-font.css`, which ends up in the `{{{css}}}` of the html template (first call in `public/build/html.min.css`). Removing it seems harmless.

ccoenen:
Thank you for trying, and more importantly: thank you for reporting back! This is often forgotten and then issues are left hanging for weeks before this is discovered - so: you did everything right!

popindavibe:
Thanks @ccoenen 

On my last tests, I realised that the fonts are failing to load because of missing CORS. I know codimd handles CSP (works great) but I don't know if it can do CORS as well.

[...]

PR #202, opened 2019-10-18
Dropping CDN support

ErikMichelson:
Removing the CDN support as stated in #175.

Help wanted: There's only one reference to the CDNs yet that resides in the `html.hbs` for the HTML export. I don't know which local files I should add there as replacement.


SISheogorath:
I think we are missing a webpack output for the html.hbs. In general we can probably have a look at the webpack.js in order to reduce the complexity of it. But I think the webpack part is out of scope of this issue.

It would be great when @davidmehren could have a look at the webpack part :)

ErikMichelson:
Even if the webpack output for html.hbs is still using some CDN sources, could this be merged prior to 1.6-release as a first step?

SISheogorath:
Personally I would like to avoid this being done half-way through. I'll may find some time to fix this completely during this week.

davidmehren:
Closing, we are redoing a CDN-less frontend anyway :smiley:

PR #205, opened 2019-10-22
Allow to generate lower case header references through the config

hoijui:
Allow to generate lower case header references through the config

This makes the references consistent/compatible with GitHub,
GitLab, Pandoc and many other tools.

This behavior can be enabled in _config.json_ with:

```
"linkifyHeaderStyle": "gfm"
```

See [the related issue on HackMd](https://github.com/hackmdio/codimd/issues/1305).

### DIFF
--- a/config.json.example
+++ b/config.json.example
@@ -123,6 +123,7 @@
         {
           "connectionString": "change this",
           "container": "change this"
-        }
+        },
+        "linkifyHeaderStyle": "gfm"
     }
 }
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -157,5 +157,6 @@ module.exports = {
   allowEmailRegister: true,
   allowGravatar: true,
   allowPDFExport: true,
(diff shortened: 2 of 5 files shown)


SISheogorath:
On one hand I like the idea, at the same time, I see tons of links breaking when we enable this by default.

I guess we should think about storing this kind of configs along with the documents.

hoijui:
though the same. :/
That s why it is off by default, and needs to be explicitly enabled as described, through the config.

SISheogorath:
Oh seems like I mixed up the defaults and the docs setting when I skimmed over the changes. Nice!

hoijui:
...ouh I just saw that I have `gfm` in _config.json.example_, meaning for new installs, it would be the default. Is that ok?

SISheogorath:
Yes, I'm don't think it's a problem for new setups :) Just existing once shouldn't break.

PR #208, opened 2019-10-25
Replace `npm` with `yarn` commands in docs.

jfowl:


SuperSandro2000:
This is technically right but I think in 2019 you should probably use npm instead of yarn.

jfowl:
> This is technically right but I think in 2019 you should probably use npm instead of yarn.

This change is focussed on bringen the documentation on par with the setup script.
Could you open the discussion of "npm vs yarn" in a new issue?

SuperSandro2000:
> > This is technically right but I think in 2019 you should probably use npm instead of yarn.
> 
> This change is focussed on bringen the documentation on par with the setup script.
> Could you open the discussion of "npm vs yarn" in a new issue?

Na, doesn't really matter that much.

PR #213, opened 2019-10-27
First steps in refactoring the backend code

davidmehren:
This is the first of (hopefully) a few more PRs to refactor the backend code.

I am trying to work toward the following structure:
- Routers only call methods on controllers
- Controllers extract the required data from the HTTP request and pass them on. No `req` or `res` objects exist "beyond" the controllers.
- The models only contain "simple" functionality, like "create", "find in database" etc. More complex functions should reside in separate helpers.

This isn't done yet (`res` and `req` objects are used pretty much everywhere), I plan to continue the work in the next weeks.

This PR contains these major changes:
- The errors were extracted into a separate file.
- All routes referenced from `noteRouter` were moved from `response.js` into files under `lib/web/note`
- `controller.js` and `slide.js` are intended as controllers in the way outlined above. The naming isn't perfect yet and I will probably move slide-related stuff to it's own folder in the future.


lib/web/note/util.js
```diff
@@ -0,0 +1,109 @@
...
+    opengraph: ogdata
+  }
+  callback(data)
+}
+
+function isRevealTheme (theme) {
+  if (fs.existsSync(path.join(__dirname, '..', 'public', 'build', 'reveal.js', 'css', 'theme', theme + '.css'))) {
+    return theme
+  }
+  return undefined
```
SISheogorath: Just a thought, but writing undefined as theme, might be a bad idea? Does it fallback to the default then?

Even when this is most likely out of scope of this PR.

PR #214, opened 2019-10-27
Rework config management

### LINKED ISSUE #155: Improve config validation / feedback
We see a lot of issues caused by wrong configuration. This is an indicator that our docs are not sufficient and even more that our software takes garbage to produce garbage.

We should improve our configs in the perspective that we get proper validation for all options. Most common problems are caused by the `domain` option. This along with `https` prefixing and port numbers (where we already have some feedback), should prevent half of all problems.

We can also work on linking error messages to docs, to improve the ability of figuring out problems.

Feel free to comment where you think work is needed the most.

SISheogorath:
At the moment, CodiMD has an self-written config management library
which works more or less great. It leads to confusion about config
values, as well as providing semi-great feedback about the values.

The new library called node-convict is written and maintained by Mozilla
and provides schema-based validation for configuration. allows easier
extension of the config mangement to formats like yaml and commandline
arguments. And can be used to improve the configs themselves over time.

This patch is a first step towards massively improving the config
managment. Next steps should be:

- Let components like the github-login-integration provide an own schema
that is used for validation.
- Add more config file formats
- Improve formats for existing config variables
- Further reduce the handling of own config settings.
- Split actual configs from internal state knowledge (like
`isGitHubEnabled`)
- Writing docs for the new config management

Fixes #155


app.js
```diff
@@ -29,20 +29,20 @@ var csp = require('./lib/csp')
 // server setup
 var app = express()
 var server = null
-if (config.useSSL) {
+if (config.get('useSSL')) {
```
ccoenen: I'll not comment this on each item, but I think we should also "namespace" this a little more. And perhaps while we're at it we could use the more correct "useTLS" because luckily nobody is using SSL anymore.

app.js
```diff
@@ -171,40 +171,40 @@ app.use(require('./lib/web/middleware/codiMDVersion'))
...
+  openID: config.get('isOpenIDEnable'),
+  email: config.get('isEmailEnable'),
+  allowEmailRegister: config.get('allowEmailRegister')
 }
 
-// Export/Import menu items
-app.locals.enableDropBoxSave = config.isDropboxEnable
-app.locals.enableGitHubGist = config.isGitHubEnable
-app.locals.enableGitlabSnippets = config.isGitlabSnippetsEnable
+//') Export/Import menu items
```
ccoenen: this `//')` is probably a copy&paste / search&replace remnant.

lib/config/convict.js
```diff
@@ -0,0 +1,835 @@
...
+        throw new Error('Loglevel has to be set to `debug`, `verbose`, `info`, `warn`, or `error`')
+      }
+    },
+    env: 'CMD_LOGLEVEL',
+    default: 'info'
+  },
+  urlAddPort: {
+    doc: 'set to add port on callback URL (ports `80` or `443` won\'t be applied) (only applied when domain is set)',
+    format: 'Boolean',
+    env: 'CMD_URL_ADDPORT',
```
ccoenen: we could fix this inconsistency: if it is (camel case) `urlAddPort` it should be `CMD_ADD_PORT` (note the last underscore)

lib/config/convict.js
```diff
@@ -0,0 +1,835 @@
...
+  },
+  useSSL: {
+    doc: 'Set to use SSL server (if `true`, will auto turn on `protocolUseSSL`',
+    format: 'Boolean',
+    env: 'CMD_USESSL',
+    default: false
+  },
+  hsts: {
+    enable: {
+      doc: 'Set to enable [HSTS](https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security) if HTTPS is also enabled',
```
ccoenen: does markdown make sense here? at a quick glance, I didn't find anything on markdown in the node-convict documentation.

lib/config/convict.js
```diff
@@ -0,0 +1,835 @@
...
+      default: true
+    },
+    upgradeInsecureRequests: {
+      default: 'auto'
+    },
+    reportURI: {
+      doc: 'Allows to add a URL for CSP reports in case of violations.',
+      format: '*',
+      default: undefined
+    }
```
ccoenen: is it correct that all of these do not have ENV equivalents?

lib/config/convict.js
```diff
@@ -0,0 +1,835 @@
...
+  },
+  forbiddenNoteIDs: {
+    doc: 'List of note IDs that can\'t be used. Main purpose is to prevent bugs due usage of paths in free-url-mode.',
+    env: 'CMD_FORBIDDEN_NOTE_IDS',
+    default: ['robots.txt', 'favicon.ico', 'api', 'build', 'css', 'docs', 'fonts', 'js', 'uploads', 'vendor', 'views']
+  },
+  defaultPermission: {
+    doc: 'Set notes default permission (only applied on signed-in users).',
+    format: function (val) {
+      if (!/(freely|editable|limited|locked|protected|private)/.test(val)) {
```
ccoenen: would probably be easier to read with `['freely', ... 'private'].includes(val)` but would be a little longer. Even more readable might be `Object.keys(Permission).includes(val)` with `Permission` from here: https://github.com/codimd/server/blob/50c80c99a4246c60026517b8632a1bbb05e1fe5c/lib/config/enum.js#L9-L16

lib/config/convict.js
```diff
@@ -0,0 +1,835 @@
...
+        throw new Error('defaultPermission has to be set to `freely`, `editable`, `limited`, `locked`, `protected`, or `private`')
```
ccoenen: has to be set to `Object.keys(Permission).join(', ')` if the note above is implemented.

lib/config/convict.js
```diff
@@ -0,0 +1,835 @@
...
+  dbURL: {
+    doc: 'Database URL',
+    format: '*',
+    default: undefined,
+    sensitive: true
+  },
+  db: {
+    host: {
+      doc: 'Database host name/IP',
+      format: '*',
```
ccoenen: wouldn't using the domain regex make some sende? (this would at least exclude white spaces). same goes for `dbUrl` a few lines above.

lib/config/convict.js
```diff
@@ -0,0 +1,835 @@
...
+  sslCAPath: {
+    doc: 'SSL CA certificate path absolute or relative to CodiMD\'s base director (only need when you set `useSSL`)',
+    format: 'optional-path',
+    default: ''
+  },
+  dhParamPath: {
+    doc: 'SSL DH parameter absolute or relative to CodiMD\'s base director (only need when you set `useSSL`)',
+    format: 'optional-path',
+    default: ''
+  },
```
ccoenen: might be a good time to dump these?

lib/config/convict.js
```diff
@@ -0,0 +1,835 @@
...
+  // socket.io
+  heartbeatInterval: {
+    doc: 'Time interval in milliseconds used for heartbeats that are send out by the client to check that the websocket connection is still conencted.',
+    format: 'nat',
+    default: 5000
+  },
+  heartbeatTimeout: {
+    doc: 'Time in milliseconds that is allowed to pass without heartbeat before the websocket connection is considered disconnected.',
+    format: 'nat',
+    default: 10000
```
ccoenen: we could verify that it's longer than `heartbeatInterval`, just in case. Not that anyone I know ever modified any of the two.

lib/config/convict.js
```diff
@@ -0,0 +1,835 @@
...
+  },
+  // too busy timeout
+  tooBusyLag: {
+    doc: 'Maximum lag allowed in the NodeJS event loop, before the CodiMD tells new connections, that the server is too busy at the moment, and handles existing requests first.',
+    format: 'nat',
+    env: 'CMD_TOOBUSY_LAG',
```
ccoenen: camel case convention would mean spelling it as `CMD_TOO_BUSY_LAG`

lib/config/convict.js
```diff
@@ -0,0 +1,835 @@
...
+      }
+    },
+    default: 'keep-case'
+  },
+  version: {
+    doc: 'Version number of CodiMD',
+    format: 'semver',
+    default: null
+  },
+  minimumCompatibleVersion: {
```
ccoenen: this file is getting exceedingly long. Maybe we should break it into a few sections? I tried something similar, when I rewrote the two readme files, look at the h2 headings for what I mean by beaking it up: https://github.com/codimd/server/blob/master/docs/configuration-env-vars.md

lib/config/convict.js
```diff
@@ -0,0 +1,835 @@
...
+  isGoogleEnable: {
+    default: false
+  },
+  isDropboxEnable: {
+    default: false
+  },
+  isTwitterEnable: {
+    default: false
+  },
+  isEmailEnable: {
```
ccoenen: could we auto-enable these (if they are not already auto-enabled) just by having any form of configuration on any of its values?

lib/config/index.js
```diff
@@ -34,25 +34,20 @@ const configFilePath = path.resolve(appRootPath, process.env.CMD_CONFIG_FILE ||
...
-merge(config, require('./hackmdEnvironment'))
-merge(config, require('./environment'))
-merge(config, require('./dockerSecret'))
-
-if (['debug', 'verbose', 'info', 'warn', 'error'].includes(config.loglevel)) {
+let config = require('./convict')
+config.load(debugConfig)
+config.load(packageConfig)
+config.load(fileConfig)
+config.load(require('./dockerSecret'))
```
ccoenen: If we break down the convict configuration, I think this would be the place to put it together. (not married to that, though if anything else is as readable).

lib/config/index.js
```diff
@@ -34,25 +34,20 @@ const configFilePath = path.resolve(appRootPath, process.env.CMD_CONFIG_FILE ||
...
+
+if (['debug', 'verbose', 'info', 'warn', 'error'].includes(config.get('loglevel'))) {
   logger.level = config.loglevel
 } else {
   logger.error('Selected loglevel %s doesn\'t exist, using default level \'debug\'. Available options: debug, verbose, info, warn, error', config.loglevel)
 }
```
ccoenen: this sanity check is already in convict config, including the error message. It should be correct at this point.

lib/config/index.js
```diff
@@ -34,25 +34,20 @@ const configFilePath = path.resolve(appRootPath, process.env.CMD_CONFIG_FILE ||
...
 
 // load LDAP CA
-if (config.ldap.tlsca) {
+if (config.get('ldap').tlsca) {
```
ccoenen: would this be `'ldap.tlsca'` ?

lib/config/index.js
```diff
@@ -63,75 +58,75 @@ if (config.ldap.tlsca) {
...
 }
 
-// cache result, cannot change config in runtime!!!
-config.isStandardHTTPsPort = (function isStandardHTTPsPort () {
-  return config.useSSL && config.port === 443
-})()
-config.isStandardHTTPPort = (function isStandardHTTPPort () {
-  return !config.useSSL && config.port === 80
-})()
+config.load({permission: permissions})
```
ccoenen: Depending on how this is used, I'd even go for `availablePermissions` or something as descriptive. `permission` (singular) feels somewhat wrong.

lib/config/index.js
```diff
@@ -63,75 +58,75 @@ if (config.ldap.tlsca) {
...
-config.serverURL = (function getserverurl () {
-  var url = ''
-  if (config.domain) {
-    var protocol = config.protocolUseSSL ? 'https://' : 'http://'
-    url = protocol + config.domain
-    if (config.urlAddPort) {
-      if (!config.isStandardHTTPPort || !config.isStandardHTTPsPort) {
-        url += ':' + config.port
+config.load({
+  serverURL: (function getserverurl () {
```
ccoenen: why is this a named function (`getserverurl`)?

lib/config/index.js
```diff
@@ -63,75 +58,75 @@ if (config.ldap.tlsca) {
...
+    const isStandardHTTPsPort = config.get('useSSL') && config.get('port') === 443
+    const isStandardHTTPPort = config.get('useSSL') && config.get('port') === 80
```
ccoenen: the second one should probably be `!config.get('useSSL')` (note the "not" in front)

lib/config/index.js
```diff
@@ -63,75 +58,75 @@ if (config.ldap.tlsca) {
...
+    let url = ''
+    if (config.get('domain')) {
+      const protocol = config.get('protocolUseSSL') ? 'https://' : 'http://'
+      url = protocol + config.get('domain')
```
ccoenen: we could also use a url builder here instead of string concatenation. (extends to the next few lines as well)

lib/config/index.js
```diff
@@ -63,75 +58,75 @@ if (config.ldap.tlsca) {
...
+    return url
+  })()
+})
 
 if (config.serverURL === '') {
   logger.warn('Neither \'domain\' nor \'CMD_DOMAIN\' is configured. This can cause issues with various components.\nHint: Make sure \'protocolUseSSL\' and \'urlAddPort\' or \'CMD_PROTOCOL_USESSL\' and \'CMD_URL_ADDPORT\' are configured properly.')
 }
 
-config.Environment = Environment
+logger.debug(typeof config.get('facebook.clientID') && config.get('facebook.clientSecret'))
```
ccoenen: leftover from debugging?

lib/config/index.js
```diff
@@ -63,75 +58,75 @@ if (config.ldap.tlsca) {
...
+  isTwitterEnable: toBooleanConfig(config.get('twitter.consumerKey') && config.get('twitter.consumerSecret')),
+  isEmailEnable: toBooleanConfig(config.get('email')),
+  isOpenIDEnable: toBooleanConfig(config.get('openID')),
+  isGitHubEnable: toBooleanConfig(config.get('github.clientID') && config.get('github.clientSecret')),
+  isGitLabEnable: toBooleanConfig(config.get('gitlab.clientID') && config.get('gitlab.clientSecret')),
+  isMattermostEnable: toBooleanConfig(config.get('mattermost.clientID') && config.get('mattermost.clientSecret')),
+  isLDAPEnable: toBooleanConfig(config.get('ldap.url')),
+  isSAMLEnable: toBooleanConfig(config.get('saml.idpSsoUrl')),
+  isOAuth2Enable: toBooleanConfig(config.get('oauth2.clientID') && config.get('oauth2.clientSecret')),
+  isPDFExportEnable: toBooleanConfig(config.get('allowPDFExport'))
```
ccoenen: ah here we go, it _is_ auto-set, just not in the place I expected.

lib/config/index.js
```diff
@@ -63,75 +58,75 @@ if (config.ldap.tlsca) {
...
+})
+
+
 // If gitlab scope is api, enable snippets Export/import
-config.isGitlabSnippetsEnable = (!config.gitlab.scope || config.gitlab.scope === 'api') && config.isGitLabEnable
+config.load({isGitlabSnippetsEnable: toBooleanConfig(!config.get('gitlab').scope || config.get('gitlab').scope === 'api') && config.get('isGitLabEnable')})
```
ccoenen: `isGitlabSnippetsEnable` should be `isGitlabSnippetsEnable_d_` for more correct english.

lib/config/index.js
```diff
@@ -150,66 +145,56 @@ for (let i = keys.length; i--;) {
   }
 }
```
ccoenen: we could start getting rid of this as well. Or maybe convict has a facility of dealing with this? would it throw an error if something unexpected was configured? if not, we should look into if it _could_ throw errors or at least warnings.

lib/config/index.js
```diff
@@ -150,66 +145,56 @@ for (let i = keys.length; i--;) {
...
 
-// Notify users about the prefix change and inform them they use legacy prefix for environment variables
-if (Object.keys(process.env).toString().indexOf('HMD_') !== -1) {
-  logger.warn('Using legacy HMD prefix for environment variables. Please change your variables in future. For details see: https://github.com/codimd/server#environment-variables-will-overwrite-other-server-configs')
-}
-
```
ccoenen: yeah \o/

lib/config/index.js
```diff
@@ -150,66 +145,56 @@ for (let i = keys.length; i--;) {
...
+// })
+//
+// config.sslCertPath = path.resolve(appRootPath, config.sslCertPath)
+// config.sslKeyPath = path.resolve(appRootPath, config.sslKeyPath)
+// config.dhParamPath = path.resolve(appRootPath, config.dhParamPath)
+// config.viewPath = path.resolve(appRootPath, config.viewPath)
+// config.tmpPath = path.resolve(appRootPath, config.tmpPath)
+// config.defaultNotePath = path.resolve(appRootPath, config.defaultNotePath)
+// config.docsPath = path.resolve(appRootPath, config.docsPath)
+// config.uploadsPath = path.resolve(appRootPath, config.uploadsPath)
```
ccoenen: in a final release they would just be thrown out, I presume.

lib/csp.js
```diff
@@ -79,16 +79,16 @@ function getCspNonce (req, res) {
 }
 
 function addUpgradeUnsafeRequestsOptionTo (directives) {
-  if (config.csp.upgradeInsecureRequests === 'auto' && config.useSSL) {
+  if (config.get('csp').upgradeInsecureRequests === 'auto' && config.get('useSSL')) {
```
ccoenen: can this be `.get('csp.upgradeInsecureRequests')`?

lib/models/note.js
```diff
@@ -94,24 +94,26 @@ module.exports = function (sequelize, DataTypes) {
             var body = null
             let filePath = null
             if (!note.alias) {
-              filePath = config.defaultNotePath
+              logger.info(path.join( config.get('defaultNotePath')))
```
ccoenen: leftover from debugging?

lib/models/note.js
```diff
@@ -94,24 +94,26 @@ module.exports = function (sequelize, DataTypes) {
...
+              filePath = config.get('defaultNotePath')
             } else {
-              filePath = path.join(config.docsPath, note.alias + '.md')
+              logger.info(path.join(config.get('docsPath'), note.alias + '.md'))
```
ccoenen: left over from debugging?

lib/models/note.js
```diff
@@ -195,7 +197,8 @@ module.exports = function (sequelize, DataTypes) {
           }
         }).then(function (note) {
           if (note) {
-            let filePath = path.join(config.docsPath, noteId + '.md')
+            logger.info(path.join(config.get('docsPath'), noteId + '.md'))
```
ccoenen: left over from debugging?

ccoenen:
Wow that's a LOT of work that went into it. I'm looking forward to seeing this in action. My review was purely reading it in the browser, I haven't given it a test-drive yet.

ccoenen:
also: I mixed various levels of tediousness into this review. Feel free to just skip any comment if it isn't the right level for this pull request :-)

lib/config/convict.js
SISheogorath: Yeah, it's rather questionable. It was also mainly copy-paste from the docs.

lib/config/convict.js
SISheogorath: For half of these, it's correct. But ReportURI has one.

app.js
SISheogorath: This would break everyone's setup that uses this variable. I mean, we can do that' but I wanted to mention it before we do that.

lib/csp.js
SISheogorath: Both works. the regex I used to transform those, was focused on the first option.

https://github.com/mozilla/node-convict/#configgetname

lib/config/convict.js
SISheogorath: This actually should become a custom format in future. The problem is that it could be undefined and the domain name or similar would simply break that.

I guess this will be implemented with an own format properly towards the end of the week.

lib/config/convict.js
SISheogorath: Sounds like an idea. We removed similar validation code before.

lib/config/convict.js
SISheogorath: Entirely? This would eliminate the need for useTLS completely. But it would also prevent secure backend setups, where the reverse-proxy is not running on the same host as the CodiMD instance.

lib/config/convict.js
SISheogorath: You mean, that's it's not shorter than heartbeatInterval? Yes and now. I don't think we can already use defaults we just configured. And even when we can, my head has a hard time to wrap around that. We might want to do this validation out of band?

lib/config/convict.js
SISheogorath: Yes, we should do this. For now, this was the first working version I wanted to provide in order to make it possible to use those.

In the next iteration, I would like to try to separate config schemas for modules, from the core ones. This should already shorten the file a lot.

lib/config/convict.js
SISheogorath: We have the actual config variables for that above. `isEmailEnable` is actually just a internal state that is not actually configured by admins. That's done through `email: true`, which is by default enabled.

Those variables basically exists for now as part of the convict config to keep the current application logic as it is and not having the need to rewrite half of that with a separate state library.

lib/config/index.js
SISheogorath: Convict doesn't work that way. You need bring the schema together before submitting it into convict. I would prefer to have once big schema for the top-level and then other schemas as part of the types for settings of components like CSP, LDAP, …

This should simplify config mangement over time and is listed as a next step above.

lib/config/index.js
SISheogorath: Because.

It was named like that before I took care of it, therefore I didn't change it. We could do this now, but it's also not really a problem. I actually consider to make serverURL an public option that can be used in stead of the `domain`, `protocolUseSSL` and `protocolAddPort` construct which is unneeded complicated.

lib/config/index.js
SISheogorath: The new URL API is not that great for that actually. We might want to let users set it an just verify it as part of the schema.

lib/config/convict.js
ccoenen: do we / can we generate markdown files from the configuration? if so, i'd leave it in. If not, i'd go for 
```suggestion
      doc: 'Set to enable HTTP Strict Transport Security if HTTPS is also enabled. For more info visit https://en.wikipedia.org/wiki/HTTP_Strict_Transport_Security',
```

lib/config/convict.js
ccoenen: we can't be the only ones who have interdependent configuration items? anyway it's probably not really important.

lib/config/convict.js
ccoenen: OK, got it :-)
Then I'd request adding a line about that as a comment before all of the `is....Enable` statements.

app.js
```diff
@@ -81,13 +81,13 @@ var sessionStore = new SequelizeStore({
...
 
 // use hsts to tell https users stick to this
-if (config.hsts.enable) {
+if (config.get('hsts').enable) {
   app.use(helmet.hsts({
-    maxAge: config.hsts.maxAgeSeconds,
-    includeSubdomains: config.hsts.includeSubdomains,
-    preload: config.hsts.preload
+    maxAge: config.get('hsts').maxAgeSeconds,
+    includeSubdomains: config.get('hsts').includeSubdomains,
```
ErikMichelson: See #99 - we should use `includeSubDomains` here.

lib/config/convict.js
```diff
@@ -0,0 +1,836 @@
...
+    },
+    default: 'keep-case'
+  },
+  version: {
+    doc: 'Version number of CodiMD',
+    format: 'semver',
+    default: null
+  },
+  minimumCompatibleVersion: {
+    doc: 'Versio number of compatible CodiMD frontends',
```
ErikMichelson: Just a little typo: '_Version number_' instead of '_Versio number_'

lib/web/imageRouter/s3.js
```diff
@@ -43,12 +43,12 @@ exports.uploadImage = function (imagePath, callback) {
...
-        s3Endpoint = config.s3.endpoint
-      } else if (config.s3.region && config.s3.region !== 'us-east-1') {
-        s3Endpoint = `s3-${config.s3.region}.amazonaws.com`
+      if (config.get('s3').endpoint) {
+        s3Endpoint = config.get('s3').endpoint
+      } else if (config.get('s3').region && config.get('s3').region !== 'us-east-1') {
+        s3Endpoint = `s3-${config.get('s3').region}.amazonaws.com`
       }
-      callback(null, `https://${s3Endpoint}/${config.s3bucket}/${params.Key}`)
+      callback(null, `https://${s3Endpoint}/${config.get('s3bucket}/${params').Key}`)
```
ErikMichelson: This url seems to be broken.

SISheogorath:
Just updated the PR to the most recent changes on master. Preparing for Release 1.6.0.

SuperSandro2000:
Does this break current configs?

SISheogorath:
It'll definitely break previously deprecated configs, yes. This will be part of the release notes and I guess most people will notice due to the big error message that will show up on startup, before exiting in case of broken configs.

SuperSandro2000:
Can you jump to 1.6.0 to indicate that? Thanks in advance!

SISheogorath:
I didn't plan anything else. By "preparing for Release 1.6.0" I meant that this should be merged and the final release work for 1.6.0 should start, so yes :)

lib/config/convict.js
```diff
@@ -0,0 +1,850 @@
...
+      doc: 'Database host name/IP',
+      format: '*',
+      default: undefined
+    },
+    name: {
+      doc: 'Database name',
+      format: '*',
+      default: undefined
+    }
+  },
```
ErikMichelson: Are the database config options limited to `name` and `host` by this?
AFAIK this object is given directly into sequelize's constructor which also may take parameters like `retry`, etc.

InnayTool:
Is this PR relevant for release 2.0? @SISheogorath do you plan to implement a new config management for 2.0 or do you have some advice how we should do it?

SISheogorath:
Sorry, I'm currently involved in too many other projects to redo this one. Kind of hurts because I really wanted to do that, but also don't want to block you :)

So here is the idea:

with node-convict one can define proper data models for config elements. So the idea is to create a semantic configuration file, and config definition which provides error messages as well as ensures that no invalid config is provided. As you can already see in some parts of this PR.

Something that is definitely still to do: Separate application knowledge/conclusions i.e. settings that result out of a combination of configuration options (e.g. a github login being available due to app-id and app-secret being set) from actual user-configurable settings.

Important things to implement:
- Replacing the `domain`, `protocolUseSSL` and `urlAddPort` and replace it with a single `url` config element
- rework the example config into a YAML format that has proper sections. ([example can be found in the forum](https://community.codimd.org/t/the-config-files-in-2-0/85))
- automatically detect that files exist during config validation as well as all kinds of string formats (https://github.com/codimd/server/pull/214/files#diff-b3419dc34b439f4644507498ff30f620R9-R58)

This will make our all lives a lot easier and prevent the most common configuration mistakes. Some parts are already done and might need minor changes, the major work is to replace the config object with `config.get` statements or similar. And maybe you'll need some work to get convict going with TypeScript but it also shouldn't be a major blocker.

I hope that gives you the info you were looking for :slightly_smiling_face:

InnayTool:
Due to the last minor release of version 1, I closed this PR.

PR #215, opened 2019-10-28
Upgrade all ORM/database related packages

SISheogorath:
This patch provides some major upgrades to all database backend library.
It also fixes an issues that appears since the change from sequelize v3
to v5 where mariadb was originally handled by mysql2 and is now handled
by an own mariadb library.

Signed-off-by: Sheogorath <sheogorath@shivering-isles.com>


ccoenen:
is the mysql2 package still needed with the mariadb package in place?

SISheogorath:
Yes and no. The as stated in the commit message, sequelize 5 utilizes either library depending on the used backend.

See: https://sequelize.org/v5/manual/dialects.html

PR #225, opened 2019-11-23
[WIP] TypeScript demo

davidmehren:
While working on more refactoring in the backend I thought "When I'm already touching most code, why not add some types?" and started playing around with adding TypeScript to the project. I mostly followed [the official guide](https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html). In the progress I moved the `app.js` file into `lib`, so the TypeScript compiler could recognize it.

This PR mostly serves as a proof of concept that we could gradually add types to this project and therefore make future refactoring easier. Of course, just running `node app.js` is no longer possible, `tsc` has to run first, that takes about 8 seconds on my notebook. I don't think that's a big problem, we already run the db migrations in our `yarn start` script.

What do you think, should we pursue the TypeScript route?


package.json
```diff
@@ -2,18 +2,18 @@
...
-    "eslint": "node_modules/.bin/eslint --max-warnings 0 lib public test app.js",
+    "eslint": "node_modules/.bin/eslint --max-warnings 0 lib public test",
     "jsonlint": "find . -not -path './node_modules/*' -type f -name '*.json' -o -type f -name '*.json.example' | while read json; do echo $json ; jq . $json; done",
     "mocha-suite": "NODE_ENV=test CMD_DB_URL=\"sqlite::memory:\" mocha --exit",
     "standard": "echo 'standard is no longer being used, use `npm run eslint` instead!' && exit 1",
     "dev": "webpack --config webpack.dev.js --progress --colors --watch",
     "heroku-prebuild": "bin/heroku",
     "build": "webpack --config webpack.prod.js --progress --colors --bail",
-    "start": "sequelize db:migrate && node app.js"
+    "start": "tsc && sequelize db:migrate && node built/app.js"
```
ErikMichelson: Sure about `built` as directory? I would rather suggest using `dist`.

lib/models/author.ts
```diff
@@ -1,11 +1,11 @@
-'use strict'
 // external modules
-var Sequelize = require('sequelize')
+import {DataTypes} from 'sequelize';
 
-module.exports = function (sequelize, DataTypes) {
-  var Author = sequelize.define('Author', {
+
+function createAutorModel(sequelize) {
```
ErikMichelson: Probably misspelled: `Author` instead of `Autor`

package.json
davidmehren: That is what the [handbook](https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html) used, but the naming is debatable, yeah ;)

SISheogorath:
When I recall correct @ccoenen is a big fan of typescript and I don't mind to learn something new here. I definitely welcome the stronger typing since it'll help to prevent various issues that appear with not validated input.

I have no strong opinion on either way and therefore would leave it to other community members to express their opinion on it, before making a final decision :)

ccoenen:
I would totally love to learn TypeScript, as such I'm not (yet?) a fan ;-) I have very little knowledge about TypeScript other than reading opinions by random strangers on the internet.

SISheogorath:
I decided that at least from me, in 2.0 it would be a good point to get TypeScript started :)

SISheogorath:
This was integrated as fundament for 2.0

PR #226, opened 2019-11-27
Fixing redirection after SAML login

foobarable:
Saving referer into session in SAML auth so passport can redirect correctly after SAML login.
Before this, after the SAML login the user was always redirected to config.serverURL + '/'


foobarable:
Messed up git repo while signing, creating new MR.

PR #227, opened 2019-11-27
Fixing redirection after SAML login

foobarable:
Saving referer into session in SAML auth so passport can redirect correctly after SAML login.
Before this, after the SAML login the user was always redirected to config.serverURL + '/'

Signed-off-by: Ralph Krimmel <rkrimme1@gwdg.de>


SISheogorath:
thanks a lot!

PR #239, opened 2019-12-15
[DOC] Use `npm start`, not `yarn start` to start

bluehood:
`yarn start --production` ignores the `--production` flag,
`npm start --production` does not.

### DIFF
--- a/docs/setup/manual-setup.md
+++ b/docs/setup/manual-setup.md
@@ -26,7 +26,7 @@ Manual Installation
    - `postgres://username:password@localhost:5432/codimd`
    - `mysql://username:password@localhost:3306/codimd`
    - `sqlite://:memory:`
-8. It is recommended to start your server manually once: `yarn start --production`, this way it's easier to see warnings or errors that might occur (leave out `--production` for development).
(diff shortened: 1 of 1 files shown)


bluehood:
Resolved conflicts, ready to merge

ccoenen:
I know this was already merged, but shouldn't we also update

https://github.com/codimd/server/blob/313eb74ed697c5260df9293b1aefe3695cf65a5a/bin/setup

at the same time? If we stop using yarn (which I'm perfectly fine with!) we should also remove it from the setup script.

bluehood:
I can't comment on dropping yarn altogether. There are still calls to yarn in the manual setup instructions. The PR only changed `yarn start` to `npm start` because otherwise the `--production` flag is ignored.

This might very well be a bug in yarn that they will fix in a next release, I don't know.

SISheogorath:
I don't think we want to remove yarn, but just use npm for for the production start (since in yarn it simply doesn't work as with npm). When recently someone went through all files and replaced npm with yarn, it was unnoticed that `npm start --production` is fundamentally different from `yarn start --production`.

ccoenen:
ok, got it :-)

Issue #148, opened 2019-08-21
Script for managing users only working with SQLite?


morpheus-87:
It seems that the script `bin/manage_users` is only working with SQLite. I would like to use a MySQL database and get the following error, when I run the script (`./bin/manage_users --pass foo --add 'foo@bar'`):

```
(node:28539) UnhandledPromiseRejectionWarning: SequelizeDatabaseError: SQLITE_ERROR: no such table: Users
```

What could be the reason for that issue?

SISheogorath:
There are multiple possible reasons. Starting with using the wrong configuration (you usually run CodiMD in production mode therefore you need to use this for the management script as well), this can be solved by prefixing the command with `NODE_ENV=production`.

Then again missing migrations: Make sure you configured the `.sequelizerc` properly and run `npm start` or `npm start --production` the migrations will run and CodiMD will start, this makes sure the tables are in the desired state for the CLI to work.

morpheus-87:
It was the first thing you mentioned, when I prefixed the command with `NODE_ENV=production` everything worked. Thanks for your help!

PR #244, opened 2019-12-24
[DOC] Run `manage_users` with `NODE_ENV=production` set.

savaroskij:
`manage_user` script defaults to `development` environment.
Add to doc the solution suggested by @SISheogorath in #148 .


SISheogorath:
Thanks a lot!

PR #250, opened 2020-01-08
Fix #249 - Focus user field after opening login modal

### LINKED ISSUE #249: wishlist: when opening the login form, the username field should be focused
When I click the "sign in" button, the login form pops up, but unlike in many login forms ([github's](https://github.com/login) for example), no input field is focused, so I can't type my login/password straight away, I have to click on the username field and then I can type.

Of course, it's minor, it's a little detail, but I bet many users will appreciate it!

ErikMichelson:
When the login dialog is opened, the focus will be set to either the LDAP-username field, the email field or the OpenID field in this order.

### DIFF
--- a/public/js/cover.js
+++ b/public/js/cover.js
@@ -428,3 +428,17 @@ $('.ui-use-tags').on('change', function () {
 $('.search').keyup(() => {
   checkHistoryList()
 })
+
+// focus user field after opening login modal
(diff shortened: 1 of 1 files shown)


hydrargyrum:
I'm not knowledgeable on the subject so my opinion has no real value but it looks good to me.

SISheogorath:
LGTM! Thanks for your work and sorry for the late response.

PR #257, opened 2020-01-15
Fix font loading when useCDN is false

### LINKED ISSUE #253: Font paths broken if use CDN is false
See reports [here](https://github.com/codimd/server/commit/23c7b5b0a6a4bb6ba79eb8ee61346de623f7aced#commitcomment-36762656): my PR to fix `urlPath` support, #242, broke the paths of certain fonts in some installations, e.g. in `build/cover-styles-pack.css`.

@MartB , @Techassi, how can one reproduce the problem? demo.codimd.org does not seem to be affected, although I remember @SISheogorath mentioning the he also saw some broken paths.

I can spend some time on this (seems right since I broke it and I'd rather have `urlPath` keep working rather than revert #242 :smile: ).

bluehood:
It should resolve #253.

It would be great if @MartB or @Techassi could try out the changes and confirm they fix their issue.

### DIFF
--- a/public/css/font.css
+++ b/public/css/font.css
@@ -3,7 +3,7 @@
   font-family: 'Source Code Pro';
   font-style: normal;
   font-weight: 300;
-  src: local('Source Code Pro Light'), local('SourceCodePro-Light'), url('/fonts/SourceCodePro-Light.woff') format('woff');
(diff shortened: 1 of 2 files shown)


MartB:
I pulled in your changes, testing currently will report here in an edit.

Techassi:
Can confirm this works. The path now is `.../build/3ab8[...].eot` 👍

Thank you for the quick fix!

SISheogorath:
Awesome people you are! Thanks for your work!

PR #173, opened 2019-09-08
Extended config.js documentation on login methods and libravatar

ErikMichelson:
Replaced gravatar with libravatar (as in #153).
Extended the documentation on login methods with links to the guides etc.

PR #261, opened 2020-01-22
Changed Gravatar to Libravatar in privacy-template

ErikMichelson:
Very tiny PR: In #173 the documentation was updated regarding the usage of Libravatar instead of Gravatar. This PR does the update for the privacy template document.

Btw: The privacy statement at demo.codimd.org still refers to Gravatar too.


SISheogorath:
Good catch! Thanks for getting back to this. when I was about to fix this, I first had to work with the libravatar people, to publish a privacy policy and that took a while. :+1:

PR #264, opened 2020-02-07
Update mermaid

Belphemur:
I wanted to be able to use the latest feature of mermaid and an older version was in use.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,7 @@
     "markdown-pdf": "^10.0.0",
     "mathjax": "~2.7.6",
     "mattermost": "^3.4.0",
-    "mermaid": "~8.2.3",
+    "mermaid": "~8.4.6",
     "meta-marked": "git+https://github.com/codimd/meta-marked#semver:^0.4.5",
     "method-override": "^2.3.7",
     "minimist": "^1.2.0",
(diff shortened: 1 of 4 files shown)


ccoenen:
Travis reports that some dependency is now incompatible with Node.JS 8. [The Dependency in question is reveal.js](https://travis-ci.org/codimd/server/jobs/647490985#L219), which is also upgraded in `package-lock.json` (3.7.0->3.9.2)

Reveal is [not a direct or indirect dependency of mermaid](http://npm.anvaka.com/#/view/2d/mermaid), so if you could try to [keep reveal.js to a version smaller than 3.8](https://github.com/hakimel/reveal.js/commit/4a8899e52068ee303291b9b9b01cf6cb984cd08e) it might run through fine.

ccoenen:
Node 8 is [fresh out of maintenance](https://nodejs.org/en/about/releases/), but I would like to avoid dropping it on such short notice.

Belphemur:
It's already broken on master if you check the current status. 

Sent from Nine
________________________________

ccoenen:
you're right, I'm sorry. I'll talk to @SISheogorath and we'll get back to you. If you wish, you can join us in our [matrix chat](https://riot.im/app/#/room/#codimd:matrix.org).

Belphemur:
@ccoenen Thanks :)
I'll drop when I have the time.

Also he forgot to update the yarn.lock, currently master isn't releasable (I tried on heroku).

SISheogorath:
Nice work, I would love to merge it, but we don't do yarn.lock updates in PRs, because it simply causes merge-conflicts. We'll update the yarn.lock in master as soon as it's merged. (the reveal.js was missing out on the yarn.lock update, due to unstable WiFi on FOSDEM)

Belphemur:
What would be the process for a PR to update the dependencies? 

Because if we don't update the lock file, yarn complain about the state of the project. 

Moreover it makes the project undeployable on heroku because they check that the package.json and lock file match. 

Sent from Nine
________________________________

SISheogorath:
The usual process is, that we merge changes and update yarn.lock in master from time to time. (There are always update coming in, but it saves PRs, which are sometimes hanging out for a bit longer from the need to rebase)

For the next big release, we are [talking about moving to a 2.0](https://matrix.to/#/!ybaSlrrMNLeuoQuUuR:matrix.org/$158098505996IXuAh:shivering-isles.com?via=matrix.org&via=hackerspaces.be&via=kif.rocks), we should consider if we can make it to make travis updating the yarn.lock automatically. But I'm not completely sure about that yet.

Belphemur:
@SISheogorath Make sense :)

I've rebased my MR and only changed the version in the package.json. It's not mergeable.

SISheogorath:
As I just noticed the CDN versions are outdated, this should be fixed as well. Please check:

- `public/views/codimd/foot.ejs`
- `public/views/slide.ejs`
- `public/views/pretty.ejs`

for the updated SHA256 etc, please refer to cdnjs.com

Belphemur:
@SISheogorath CDN links updated.

SISheogorath:
Great thanks!

PR #266, opened 2020-02-09
Update CDN defaults

SISheogorath:
As we noticed in our poll about CDN usage, that most people
intentionally turn it off, but very little intetionally turn it on or
leave it on. https://community.codimd.org/t/poll-on-cdn-usage/28

There is also strong indicators that CDNs don't really provide any
benefits in loading time and due to the small deployments of CodiMD,
there is no big savings due to CDNs either. https://csswizardry.com/2019/05/self-host-your-static-assets/

Therefore this patch changes the CDN default settings to off in order to
reduce the exposed user data.


SuperSandro2000:
Can you also update https://github.com/codimd/server/blob/master/docs/configuration-config-file.md ?

SISheogorath:
Good catch, I only check the env-vars description and there was no default mentioned. I updated the docs now

PR #268, opened 2020-02-11
Adjust description of CMD_ALLOW_ANONYMOUS_EDITS

stefandesu:
`CMD_ALLOW_ANONYMOUS_EDITS` is only applied when `CMD_ALLOW_ANONYMOUS` is `false`, see [here](https://github.com/codimd/server/blob/9c1665ae5bf3cf38210ffd862d10de9b13bfacdd/lib/config/index.js#L71-L73).


stefandesu:
I think one of the Travis build stalled and needs to be restarted: https://travis-ci.org/codimd/server/jobs/648740735

I mean I only changed documentation, so the code should be good.

SISheogorath:
Thanks!

PR #284, opened 2020-02-25
Fix broken redirect on login

### LINKED ISSUE #272: logging in while editing note should redirect to note, not history
**Steps to reproduce:**

* be logged out.
* go to any note that requires being logged in to edit. For example https://demo.codimd.org/TCQO0GQ1QCiuag2TuMoTfg?edit
* type (which brings up the login modal
* log in

**Expected result:**

successful login should bring me back to that document

**Actual result:**

successful login lands me at https://demo.codimd.org - the history.

### LINKED ISSUE #273: Wrong redirection happens after login when CMD_URL_PATH is specified
Under below conditions, wrong redirection will happen:

**Required config to reproduce:**
- `CMD_URL_PATH` is specified (example: codimd)
- Reverse proxied to CodiMD like below:
```
nginx (https://www.example.com/codimd/ , Docker container)
  |
   -- CodiMD (http://0.0.0.0:3000/ , Docker container)
```

**Steps to reproduce:**
1. User attempts to access restricted page, and see 'You are not allowed to access this page' error first.
2. Then, user sign in to CodiMD.
3. User will be redirected to '/(noteid)' and nginx error will be shown.

**Expected behavior:**
User should be redirected to '/(CMD_URL_PATH)/(noteid)'

**Actual behavior:**
User is redirected to '/(noteid)'

This behavior is occurred because errors.js (https://github.com/codimd/server/blob/8ce7b285636c02b4215c838640bb748947c8317f/lib/errors.js#L10) has 2 issues:

1. I think `serverUrl` is typo. Correct one is `serverURL`.
2. `req.originalUrl` doesn't have directory of `CMD_URL_PATH` because proxied request doesn't contain it.

SISheogorath:
This patch fixes the currently broken redirect on login when people try
to access a site they have no access to, they are redirected to the main
page to log in. After a successful login they should be redirected to
the original note, but instead are redirect to the index page again.

This aptch fixes the typo that causes the behavior and brings people
back to the note they edited.

Thanks to @clvs7-gh on Github, who submitted the patch via email.

On their behalf I hereby submit the change.

Fixes #272 and fixes #273

### DIFF
--- a/lib/errors.js
+++ b/lib/errors.js
@@ -7,7 +7,7 @@ module.exports = {
       responseError(res, '403', 'Forbidden', 'oh no.')
     } else {
       if (!req.session) req.session = {}
-      req.session.returnTo = req.originalUrl || config.serverUrl + '/'
+      req.session.returnTo = config.serverURL + (req.originalUrl || '/')
       req.flash('error', 'You are not allowed to access this page. Maybe try logging in?')
       res.redirect(config.serverURL + '/')
     }


dalcde:
This fix doesn't appear to work for me on release/2.0.x with OAuth2. Instead, I get redirected back to
`https://$DOMAIN/auth/oauth2/undefined/$ORIGINAL_PATH`

davidmehren:
I didn't test this yet, but noticed the changes are different from d389f45818b39e95fb22d3e7d06f7ba061d6f34a. Was that intentional @SISheogorath ?

SISheogorath:
When I tried to apply the changes back then I noticed it would only redirect me to the 403 page. That's why the changeset in d389f45818b39e95fb22d3e7d06f7ba061d6f34a is different. I'll port those changes over. Also: I'm not sure how `https://$DOMAIN/auth/oauth2/undefined/$ORIGINAL_PATH` can happen :thinking:

davidmehren:
Closing, as this won't be necessary with the new frontend.

PR #290, opened 2020-03-03
Split frontend and backend build

SISheogorath:
It should be possible to run a backend build without the need to start
the application and without the need to fiddle around with PATH or
fiddling with `node_modules`. Therefore this patch splits the build
command into `build-frontend` and `build-backend`, which will allow
those builds to be done independent and have a separate command `build`
which will provide the combination of both of them.


package.json
```diff
@@ -12,8 +12,10 @@
     "standard": "echo 'standard is no longer being used, use `npm run eslint` instead!' && exit 1",
     "dev": "webpack --config webpack.dev.js --progress --colors --watch",
     "heroku-prebuild": "bin/heroku",
-    "build": "webpack --config webpack.prod.js --progress --colors --bail",
-    "start": "tsc && sequelize db:migrate && node built/app.js"
+    "build": "npm run-script build-backend && npm run-script build-frontend",
+    "build-backend": "tsc",
+    "build-frontend": "webpack --config webpack.prod.js --progress --colors --bail",
+    "start": "npm run-script build-backend && sequelize db:migrate && node built/app.js"
```
SuperSandro2000: Why does start compile both things? I would think it only starts my already compiled stuff.

SISheogorath: Mainly because we did it before. I agree it's reasonable to change this.

package.json
```diff
@@ -12,8 +12,10 @@
     "standard": "echo 'standard is no longer being used, use `npm run eslint` instead!' && exit 1",
     "dev": "webpack --config webpack.dev.js --progress --colors --watch",
     "heroku-prebuild": "bin/heroku",
-    "build": "webpack --config webpack.prod.js --progress --colors --bail",
-    "start": "tsc && sequelize db:migrate && node built/lib/app.js"
+    "build": "npm run-script build-backend && npm run-script build-frontend",
+    "build-backend": "tsc",
+    "build-frontend": "webpack --config webpack.prod.js --progress --colors --bail",
+    "start": "[ ! -e built/ ] && npm run-script build-backend; sequelize db:migrate && node built/lib/app.js"
```
SuperSandro2000, line 18:
```suggestion
    "start": "[ -d built/ ] && npm run-script build-backend; sequelize db:migrate && node built/lib/app.js"
```
Just to be sure it is the directory.

Also I am not sure if start should call build-frontend or build-backend should be removed.

SISheogorath: I think it makes sense to try to build the application if it's not already done. In best case we run successful, in worst case we die anyway.

SISheogorath: Also we should not check if it's a directory. The trailing slash makes sure that a directory is available. And in worst case we'll fail to call the binary anyway.

davidmehren, line 18: Suppose someone is already running a version of CodiMD with this commit and then upgrades. If `tsc` is not run manually after the upgrade, the old code from before the upgrade will continue to be used. Is that a risk we want to take?

I just tried enabling `tsc`s `incremental` flag, which roughly halves compilation time on subsequent runs, so enabling that and always running `tsc` might be an idea?

SISheogorath: If upgrades are the concern, then I would prefer to not run tsc at all in npm start to prevent container builds to run it unnecessarily.

davidmehren: Containers can just ship a custom start-script which does not run `tsc` at all as this is done at container-build-time, can't they?

But for everyone who doesn't use a container, the issue remains: we can either document that one must run `tsc`/delete `built` after an upgrade or always run `tsc` on startup.

SISheogorath: Yeah, I think the whole situation just resolved itself by the realization that we don't install typescript in production builds as it's a devDependency. Therefore running `tsc` on start is a bad idea. Also in the long run I could imagine us packaging CodiMD and then it would be also contra intuitive to have to run tsc when you get a pre-built version.

mrdrogdrog: Tbh.. why do you want to run the compiler on start up? When I'm executing "yarn start" I don't expect tsc to run at all. Regardless if `build` is existing or not.

davidmehren: The call to `tsc` was added in the first Typescript PR to preserve the current way of starting CodiMD with `yarn start`. If we stop that way of starting CodiMD from working, we need to document that. I propose to not change the startup semantics in this PR but to do this in a separate PR that also adds corresponding docs on when to recompile.

davidmehren: I just noticed that we probably don't have to change our docs, but just update `bin/setup` to also run `tsc`. We already document to run `bin/setup` after an upgrade, so everything should be solved after adjusting that.

PR #291, opened 2020-03-04
Fixes problem with relative path for themes

### LINKED ISSUE #279: Can't choose theme in slideOptions
When adding a theme in `slideOptions` in YAML metadata, it doesn't appear in the generated slides. 
It comes from the way codimd detects if a theme exists or not in [lib/web/note/util.js](https://github.com/codimd/server/blob/master/lib/web/note/util.js#L104), the filesystem path used is not the right one.

Cpavrai:
Fix #279 

Signed-off-by: Charles Parmentier <charles.parmentier@hotmail.com>

### DIFF
--- a/lib/web/note/util.js
+++ b/lib/web/note/util.js
@@ -102,7 +102,7 @@ exports.getPublishData = function (req, res, note, callback) {
 }
 
 function isRevealTheme (theme) {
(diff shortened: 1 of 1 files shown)


SISheogorath:
Can you please base this on the `release/2.0.x` branch and file it against the same branch? We want to push 2.0 forward and might avoid another release. Backports can always be done if we decide 2.0 takes too long.

Cpavrai:
Moved into https://github.com/codimd/server/pull/293

PR #309, opened 2020-03-23
Make "transform-style: preserve-3d;" screen-only.

margau:
Should fix codimd#308.
Not 100% sure if it may breaks other stuff or if there is a more elegant way.

Signed-off-by: Marvin Gaube <dev@marvingaube.de>


margau:
Hi, any updates on this?

SISheogorath:
Sorry for the long delay. Seems like it got between the things that I had on my list and then was left.

Issue #301, opened 2020-03-19
Buttons for diagrams in toolbar of editor


PascalBru:
Are you planing to add buttons for all diagrams which are supported?
With a short example for every diagram, this would be very helpful.

SISheogorath:
We haven't planned that yet, but it sounds very doable :)

Let me provide the right pointers for someone who wants to implement this:

https://github.com/codimd/server/blob/8ce7b285636c02b4215c838640bb748947c8317f/public/js/lib/editor/toolbar.html#L46-L48

https://github.com/codimd/server/blob/8ce7b285636c02b4215c838640bb748947c8317f/public/js/lib/editor/index.js#L149-L230

You might need to check out the CodeMirror docs as well, which can be found here:
https://codemirror.net/doc/manual.html#api

PascalBru:
If you provide me the icons I will try to implement it, if you like?

PascalBru:
Or is a dropdown with the names better an more clear?

SISheogorath:
I would leave that up to you, to get a PoC ready. I think details like "is it a dropdown, or everything an own icon" is a solvable issue.

Available icons can be found here: https://forkaweso.me/Fork-Awesome/icons/

PascalBru:
@SISheogorath 
Would that UI be a good solution?
<img width="470" alt="codimd_addDiagrams" src="https://user-images.githubusercontent.com/6459785/77236020-ed724700-6bba-11ea-835a-02443e973691.png">

SISheogorath:
That looks nice to me, but let me summon our UXpert @ccoenen :D

ccoenen:
This is fine for me. It's a lot better than adding 5 different icons (that would likely not be decipherable). I also like the symbol you chose. It's generic enough to fit most, yet it's pretty clear that something is being embedded.

One tiny thing: at least in the case of mermaid, this is not "one" diagram, but three or four, maybe we should add one example for each, and have seperate items in that dropdown list?

Here's an example document with all types of mermaid charts, with charts from their documentation.

https://demo.codimd.org/as5gkYwKQOWIc6Osp2Kxcg?both

SISheogorath:
Thanks!

Fixed by #312

PR #312, opened 2020-03-25
add icon to toolbar with list of diagrams

PascalBru:
Fix for #301


SISheogorath:
Nice feature, lovely execution! :+1: Awesome work!

PR #316, opened 2020-03-31
WIP: permission -> viewableBy, editableBy

petabyteboy:
### Motivation for this change

The permissions model of codimd is _very_ confusing. With 6 different permission types that have very similar names I had to read all of the descriptions and still wasn't entirely sure what each type meant. This also simplifies the logic, since the old types were just combinations of the new ones.
Still working on the UI, migration and details


lib/models/note.js
```diff
@@ -42,9 +42,13 @@ module.exports = function (sequelize, DataTypes) {
       type: DataTypes.STRING,
       unique: true
     },
-    permission: {
+    viewableBy: {
       type: DataTypes.ENUM,
-      values: permissionTypes
+      values: userTypes
```
SISheogorath, line 47: Careful with this change, it'll require a database migration to be handled properly for existing setups

lib/models/note.js
```diff
@@ -42,9 +42,13 @@ module.exports = function (sequelize, DataTypes) {
...
+    },
+    editableBy: {
+      type: DataTypes.ENUM,
+      values: userTypes
```
SISheogorath, line 51: Same for this database change.

petabyteboy:
Is the Travis CI broken or can I see the result somewhere?

SISheogorath:
I'm not sure why it's not properly linked but yes, you can see them on https://travis-ci.org/github/codimd/server

You specific run should be this one: https://travis-ci.org/github/codimd/server

ErikMichelson:
> I'm not sure why it's not properly linked [...]

I guess that originates in the migration from travis-ci.org to travis-ci.com. Is there any option to change the URL in the GitHub integration?

petabyteboy:
Will open a new PR against the 2.0.x branch

PR #322, opened 2020-04-14
permission -> viewableBy, editableBy

petabyteboy:
##### Motivation for this change

The permissions model of codimd is very confusing. With 6 different permission types that have very similar names I had to read all of the descriptions and still wasn't entirely sure what each type meant. This also simplifies the logic, since the old types were just combinations of the new ones.
Still working on the UI, migration and details

### DIFF
--- a/docs/configuration-config-file.md
+++ b/docs/configuration-config-file.md
@@ -90,7 +90,8 @@ these are rarely used for various reasons.
 | `allowAnonymous` | `true` or `false` | Set to allow anonymous usage (default is `true`). |
 | `allowAnonymousEdits` | `true` or `false` | If `allowAnonymous` is `true`: allow users to select `freely` permission, allowing guests to edit existing notes (default is `false`). |
 | `allowFreeURL` | `true` or `false` | Set to allow new note creation by accessing a nonexistent note URL. This is the behavior familiar from [Etherpad](https://github.com/ether/etherpad-lite). |
-| `defaultPermission` | `freely`, `editable`, `limited`, `locked`, `protected` or `private` | Set notes default permission (only applied on signed-in users). |
+| `defaultEditableBy` | `owner`, `signedIn` or `anyone` | Set notes default permission (only applied on signed-in users). |
+| `defaultViewableBy` | `owner`, `signedIn` or `anyone` | Set notes default permission (only applied on signed-in users). |
 | `sessionName` | `connect.sid` | Cookie session name. |
 | `sessionLife` | `14 * 24 * 60 * 60 * 1000` (14 days) | Cookie session life time in milliseconds. |
 | `sessionSecret` | `secret` | Cookie session secret. If none is set, one will randomly generated on each startup, meaning all your users will be logged out. |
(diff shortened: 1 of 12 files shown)


SISheogorath:
Generally speaking, I'm not really sure about this change. I thought about it multiple times, but I don't think it's really an improvement. Currently we have 6 possible permissions which already cause confusion as you pointed out. But when this changes we have 9 possible permission scenarios of which 3 make no sense. I think we have to rework permissions more fundamentally and probably by associating them with users and notes. Or even go full blown RBAC, but I'm not completely sure about that yet.

SuperSandro2000:
I need to agree with sheo that the current permissions are very confusing. I personally would prefer RBAC in a matrix style which should be easier to understand for everyone.

petabyteboy:
I don't think this solution is perfect, especially the user interface. My idea for a better UI is using something similar to [Material Design Chips](https://material.io/components/chips).
But I think this change brings us much closer to something like RBAC. Maybe we should talk about this in a Jitsi meeting and think it through before I continue to work on changes in the permissions system.

petabyteboy:
Some notes on what we discussed in the community call today:

- The backend part should be designed to support giving single users access to a note
  - In the database there should be a new table with the fields 'user' (user id, primary key), 'note' (note id, primary key) and 'canEdit' (boolean)
  - 'anyone' / 'anonymous' and 'signed-in users' / 'authenticated' should be fake users with a special meaning
- The frontend could work like Nextcloud's share feature
  - A list of users with view access, when a user was granted view access there will be a new entry in the list with a checkbox to grant edit access in addition and an X to remove the user from the list
  - Search with autocompletion feature, should be configurable for user privacy
  - For the UI to make sense, we need a unique username (related: https://github.com/codimd/server/pull/34), but work on this can start with the non-unique display name

dalcde:
Many authentication providers have the concept of organizations/groups a user belongs to, e.g. GitHub. It would be helpful to be able to limit access to a certain organization/group.

JeroenAdam:
Interesting development.

I have a use case of multiple teams sharing one CodiMD instance and I would like each user to only view notes from his/her own team.

Could I already play with this? Assuming I manually build CodiMD with latest .js files related to this? I'm OK with adding new rows in the database for now,

pirate:
I greatly support @petabyteboy's note about following NextCloud's approach to permissions/sharing. Permission are complex and difficult to design from scratch, it's much easier to follow an existing project that has figured it out and that people are familiar with.

FWIW permissions are still by far the biggest pain point for us, causing an issue at least once a week at our company where someone is not able to share a note and have it be editable to only specific people easily. It's the one thing preventing us from moving away from google docs (along with comments), at the moment I think no other new feature could come close to having the level of impact/improvement on our lives as a permissions overhaul.

To that end I'm happy to beta test / review / brainstorm any new permissions model PRs, we're always excited about seeing movement in that direction and want to help in any way we can.

InnayTool:
Thank you for your contribution. Because of the new back-end design from scratch and the deletion of the release/2.0.x branch, the PR was closed by the automation. The new brach where we develop for 2.0 is develop. There we plan to implement a new permission system.

PR #324, opened 2020-04-16
Limit DB schema changes to migrations

SISheogorath:
Database migrations should be in charge of changing and modifying the
database schema. Therefore this breaking change removes the automatic
model synchronisation from the application startup, that we have
practised way too long and that always caused problems for us.


SISheogorath:
The failed test is unrelated to the changes. Therefore I think we can merge this change :)

PR #328, opened 2020-04-18
Adding webmanifest to work as a PWA

### LINKED ISSUE #327: Progressive Web App
Through [progressive web app manifests](https://developer.mozilla.org/en-US/docs/Web/Manifest) it should be possible to allow CodiMD or single note to be easily added to homescreens of mobile devices with an App-like user experience.

PWAs have more features, but if I'm not mistaken, for [adding to home screen](https://developer.mozilla.org/en-US/docs/Web/Progressive_web_apps/Add_to_home_screen#How_do_you_make_an_app_A2HS-ready) all we would need is to generate a json file containing things like a link and links to icons in various sizes.

ErikMichelson:
This pull request adds the webmanifest, so that certain browsers allow the installation of CodiMD as a PWA (progressive web app).

I also removed some path prefixes as they're not needed anymore because of the base-tag.
I'm still a bit unsure about the serviceworker-cached paths - maybe somebody wants to give feedback for that?

Fixes #327

### DIFF
--- a/public/js/common.js
+++ b/public/js/common.js
@@ -1,0 +1,3 @@
+if ('serviceWorker' in navigator) {
+  navigator.serviceWorker.register('serviceworker.js')
+}
--- a/public/js/cover.js
+++ b/public/js/cover.js
@@ -442,3 +442,7 @@ $('.signin-modal').on('shown.bs.modal', function () {
     fieldOpenID.focus()
   }
 })
+
+if ('serviceWorker' in navigator) {
+  navigator.serviceWorker.register('js/serviceworker.js')
+}
--- a/public/js/cover.js
+++ b/public/js/cover.js
@@ -444,5 +444,5 @@ $('.signin-modal').on('shown.bs.modal', function () {
 })
 
 if ('serviceWorker' in navigator) {
-  navigator.serviceWorker.register('js/serviceworker.js')
+  navigator.serviceWorker.register('serviceworker.js')
 }
--- a/public/js/cover.js
+++ b/public/js/cover.js
@@ -442,7 +442,3 @@ $('.signin-modal').on('shown.bs.modal', function () {
     fieldOpenID.focus()
   }
 })
-
-if ('serviceWorker' in navigator) {
-  navigator.serviceWorker.register('serviceworker.js')
-}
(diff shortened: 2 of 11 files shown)


SISheogorath:
@ErikMichelson do you have this running somewhere? Would love to try it :)

ErikMichelson:
Currently not but in about an hour I will have the time to start my instance for that branch again.
Would be https://codimd-1.test.fsbremen.de/ then.

ErikMichelson:
It's up and running! :)

DerMolly:
I just tried it and it looks great. One thing that could may be changed is that currently at least for iOS the login does not carry to the PWA-Mode of the App and you need to login again. That can be changed, but would need extra code in the login session handling as far as my light google search told me.

SISheogorath:
Here is some feedback:

lighthouse recommends to add a [theme-color](https://web.dev/themed-omnibox) as well as ["custom splash screen"](https://web.dev/splash-screen/). 

I also noticed that currently the PWA is only installable from the cover page, might be useful to have it in the editor as well.

So far, I think it looks quite fine. To be really useful, we'll have to work a bit more with the service worker, but I think it would provide some great potential over time.

ErikMichelson:
> lighthouse recommends to add a [theme-color](https://web.dev/themed-omnibox) as well as ["custom splash screen"](https://web.dev/splash-screen/).

Both result from me forgetting to add the theme color to the manifest too. In the HTML itself it already is added. I'll fix that. Btw I used the cover-page's background color (#333) as theme color or is there any better suited color?

> I also noticed that currently the PWA is only installable from the cover page, might be useful to have it in the editor as well.

Yeah, I wasn't sure whether that should be included into the already big index.js - therefore I just added it to cover.js only. But I can change that too.

> So far, I think it looks quite fine. To be really useful, we'll have to work a bit more with the service worker, but I think it would provide some great potential over time.

I know it's very basic, but at least it's a first step.

SISheogorath:
@DerMolly How does this go together with the new frontend?

mrdrogdrog:
Very. The new frontend is built to be a pwa

mrdrogdrog:
You won't need a service worker in the Backend Code.

davidmehren:
Closing, as the react-frontend does that already.

PR #333, opened 2020-04-24
Fixed wrong locales search path after app.js move

ErikMichelson:
The locales search path needs to be updated because the main app.js was moved into a sub-directory (/lib).


SISheogorath:
Can we get an absolute path here as we have with the docs directory and alike? This should self it more permanent.

SISheogorath:
Awesome! Thanks!

PR #334, opened 2020-04-24
Add 'nofooter' option for slide presentation mode

### LINKED ISSUE #88: Hide footer in presentation mode
Hello,
I'd like to know wether it is possible to remove the footer in the presentation mode.

Regards
Andi

ErikMichelson:
The slide presentation mode has a footer with metadata about the presentation that optionally may host disqus. If you scroll down by accident in a presentation, the footer might be irritating.
This PR adds a new option "nofooter" that may be added as a parameter to the URL.
So when "?nofooter" (or "?anythingelse&nofooter") is applied to the URL, the footer will be hidden.

Fixes #88

### DIFF
--- a/public/js/slide.js
+++ b/public/js/slide.js
@@ -137,3 +137,8 @@ Reveal.addEventListener('slidechanged', renderSlide)
 const isWinLike = navigator.platform.indexOf('Win') > -1
 
 if (isWinLike) $('.container').addClass('hidescrollbar')
+
+if (window.location.search.match(/.*[?&]nofooter(?:&.*)?$/)) {
+  $('.footer').addClass('hidden')
+  $('.container').addClass('hidescrollbar')
+}


SISheogorath:
Might be a better option to be implemented as note metadata? The URL parameter easily makes it a hidden feature and is not really accessible, especially since there is not documentation about it either.

ErikMichelson:
I don't think that an implementation via note metadata is better. Maybe someone wants to share the presentation with the footer online but the same presentation is used during an event on stage where the footer should be hidden. Editing the note metadata always for this is unpractical.
There is another not well-known presentation URL parameter "print-pdf" that also has no documentation at the moment here. I'm also unsure whether another metadata parameter is more accessible.
Regarding documentation an idea could be to add the slide URL-parameters to the URL-scheme documentation (see #335).

SISheogorath:
For the print-pdf option, scroll into the footer and press the printer icon :wink:

We should definitely put some thought intothe UX here. I don't think it's overly useful to just have a URL parameter which will get lost over time. (Like we had it happening with the slide preview in the editor.)

I agree that slide options are not ideal, but at least people are familiar with it already and have to use it anyway in order to make the footer more useful.

Either way, then let's find a better way for this to be exposed to users. (Maybe it would be an idea to detect full screen mode and then hide it?

davidmehren:
Moved to the new frontend: https://github.com/codimd/react-client/issues/325

PR #338, opened 2020-04-25
Migrate Backend to TypeScript

davidmehren:
This pull request migrates all the backend code (excluding migrations) to TypeScript. A few notes:
- Huge thanks to @DerMolly, @InnayTool, @ragon000 and @nzbr for helping!
- We fixed a few bugs that became apparent after turning on types :smiley: 
- Besides that, this is *mostly* a straightforward conversion of the JavaScript code. In some places that was not possible, because `tsc` complained too much.
- Yes, this is a very big PR and to be honest, we could not test all features.
- After rebasing onto release/2.0.x, a few things broke: The `Session` table did not get created automatically (because we don't `sync()` the db on startup anymore) and jQuery was upgraded and broke Bootstrap. This PR includes fixes for both issues.
- I did not cleanup the commit history too much, because I am lazy 😅. I you prefer a cleaner history, I can squash things a bit more.
- The config-code is *really* confusing, we should find a simpler way for managing the configuration :wink:  
- We extended the ESLint config to lint `.ts` files and removed the deprecated TSLint.
- This PR includes a fix from @DerMolly for the `sequelize-cli` config. The normal `config.json` is now parsed inside `.sequelizerc.example` to create a `database.json` which is then used by `sequelize-cli`.


SISheogorath:
Just looking at the changes (I only got half way through yet) it looks nice! You people did such a great job here! I'll make sure we merge this very either later tonight or tomorrow. Cleaning up the commits doesn't seem super needed. You made an enormous amount of changes and having them a bit better broken down, is always preferred over one big commit.

About the configs I'll definitely reorganize them, but I wanted to wait for you guys, because this would have been a mess to rebase because we would literally touch every single file.

I'm really looking forward to give it a testdrive and fix a few more bugs. But overall I'm quite positive about how it looks and how it'll stabilize our backend a lot. And honestly speaking, it would have taken me months to even remotely accumulate this amount of code changes. A very big thank you, once more! :heart: YOU ROCK! :rocket:

PR #339, opened 2020-04-25
Add night mode detection

SISheogorath:
Firefox and Safari power a new CSS media query which allows to detect
user's dark mode. This patch adds a JS implementation that toggles the
our night mode automatically when a user's theme requests a dark mode.

Resources:
https://hacks.mozilla.org/2019/05/firefox-67-dark-mode-css-webrender/
https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme

### DIFF
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -446,8 +446,15 @@ $(document).ready(function () {
     })
   }
 
+  // Use the new CSS selector for dark theme requests to automatically turn on night mode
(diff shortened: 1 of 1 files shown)


SISheogorath:
@DerMolly Might want to look into this for the new frontend. Don't think it makes lots of sense to provide any new frontend code for 2.0 into this repository.

mrdrogdrog:
Please create an issue in the react client repo

davidmehren:
I think we can close this here, as there is already a corresponding issue for the new frontend.

PR #350, opened 2020-04-30
allow to set a saml client certificate

n0emis:
For some SAML-Providers it is necessary, to have a client certificate. This PR extends CodiMD, to have such a config option. This is related to a keycloak-saml-guide, that will follow shortly...


lib/web/auth/saml/index.js
```diff
@@ -17,6 +17,7 @@ passport.use(new SamlStrategy({
   entryPoint: config.saml.idpSsoUrl,
   issuer: config.saml.issuer || config.serverURL,
   cert: fs.readFileSync(config.saml.idpCert, 'utf-8'),
+  privateCert: config.saml.clientCert === undefined ? undefined : fs.readFileSync(config.saml.clientCert, 'utf-8'),
```
SuperSandro2000, line 20: I am not sure but shouldn't this default to default cert bundle if undefined?

n0emis: But then you would need another config-option to disable signing.

SISheogorath: This should NOT(!) fallback to the the default cert bundle. That's not how SAML works.

n0emis:
Sorry for the git mess. We tried to sign @em0lar's commit on my machine, which went wrong...

docs/guides/auth/saml-keycloak.md
```diff
@@ -0,0 +1,113 @@
...
+9. Move or copy this key (in this case named `key.pem`) and save it to the file specified in `saml.clientCert` property of the CodiMD configuration or in the enviroment-variable `CMD_SAML_CLIENTCERT`
+
+
+### Convert Private Certificate generated by KeyCloak
+Instead if generating you own certificate, you can also use the one generated by keycloak.
+
+1. Select "Client" in left sidebar
+2. Go to your CodiMD-Client
+3. Select the "SAML Keys" tab
+![keycloak_saml_export_cert](https://md.em0lar.de/uploads/upload_ac3343ef86dad2f10155ba9707bea042.png)
```
SISheogorath, line 76: The  images should point to the ones uploaded to the repository.

lib/web/auth/saml/index.js
SISheogorath: We should verify that the file exists before loading it and provide a proper error instead of crashing with a stacktrace.

docs/guides/auth/saml-keycloak.md
n0emis: Sorry, that was the plan. Did those things late at night and forgot to hit save...

n0emis:
Is there a reason for this comment to be resolved? https://github.com/codimd/server/pull/350#discussion_r418377312

This would also have to be fixed with the `idpCert`...

lib/web/auth/saml/index.js
n0emis: Now there is error handling with logging...

lib/web/auth/saml/index.js
```diff
@@ -16,7 +16,21 @@ passport.use(new SamlStrategy({
   callbackUrl: config.serverURL + '/auth/saml/callback',
   entryPoint: config.saml.idpSsoUrl,
   issuer: config.saml.issuer || config.serverURL,
-  cert: fs.readFileSync(config.saml.idpCert, 'utf-8'),
+  privateCert: config.saml.clientCert === undefined ? undefined : (function () {
+    try {
+      return fs.readFileSync(config.saml.clientCert, 'utf-8')
+    } catch (e) {
+      logger.error('saml client certificate not found at: ' + config.saml.clientCert)
```
SISheogorath, line 23: This is an almost silent error drop. We shouldn't do this. This would print `saml client certificate not found at: /path/top/cert.file` even when the certificate is in place but the permissions are wrong.

SISheogorath:
Can we rework this towards 2.0? I think that's more appropriated there.

lib/web/auth/saml/index.js
n0emis: Would it be okay to change the error message to something like "There was an error reading the client certificate at: /path/top/cert.file"/ Or should this have a more complex checking?

SISheogorath: I think you should also print `e.message`. Maybe something like

```js
logger.error(`SAML client certificate: ${e.message}`)
```

The logger already sets the log level to error, therefore we don't need to tell this in the error message. The prefix provides the needed pointer and the error message itself should provide the reason why things fail.

PR #355, opened 2020-05-09
Migrate tests to typescript

davidmehren:
This PR migrates the mocha tests to typescript, enables linting of .ts files and contains a few fixes to make ESLint a bit happier.

I had to replace `mock-require` with `ts-mock-imports` as it seems to not be compatible with TypeScript.


test/letter-avatars.ts
```diff
@@ -24,21 +25,19 @@ describe('generateAvatarURL() gravatar enabled', function () {
   })
 
   it('should return correct urls for names with spaces', function () {
-    assert.strictEqual(avatars.generateAvatarURL('Daan Sprenkels'), 'http://localhost:3000/user/Daan%20Sprenkels/avatar.svg')
+    assert.strictEqual(avatars.generateAvatarURL('Daan Sprenkels'), 'https://cdn.libravatar.org/avatar/4c274a80fba9203139aa2908da7832a5?s=400')
```
SISheogorath, line 28: Was this test changed intentionally?

lib/config/index.ts
```diff
@@ -14,7 +14,7 @@ import { environment } from './environment'
 import { dockerSecret } from './dockerSecret'
 import deepFreeze = require('deep-freeze')
 
-const appRootPath = path.resolve(__dirname, '../../')
+const appRootPath = process.cwd()
```
SISheogorath, line 17: Please don't. This breaks every CodiMD installation that doesn't explicitly specifies a work directory. (Which means probably all installations running on Systemd with minimal service files)

The app root directory is intentionally not the work directory.

test/letter-avatars.ts
davidmehren: Good catch! I think I was confused by the failing test, but it turns out it should fail 😅.
@DerMolly seems to have inadvertently changed the behaviour of `generateAvatarURL` with the TS-migration. I will do a PR to fix that shortly.

lib/config/index.ts
davidmehren: The problem I tried to solve with that change was that we can't reliably get the app root path relative to the current file anymore. If we are running in a node process that understands TypeScript directly (`ts-node` typically) we would have to go "2 directories up". When the complied code is running (so a normal NodeJS after running `tsc`) we would have to go "3 directories up", because the complied code resides in the `built` subdirectory.

So either CodiMS would not work in one of those cases or we have to build some kind of heuristic ("if there is a `built` directory, go 3 dirs up, else 2"). Personally, I would prefer to use the working directory so we don't have to do that, but I agree that we probably should discuss this.

Proposal: Use `process.cwd()` for now, as it works in both use-cases and discuss this topic in a separate issue before release.

SISheogorath: In order to move this forward: How about we use the current working directory for now and add a TODO hint, to fix this in a better way in the long term? Then this shouldn't block your progress.

davidmehren: I'll resolve this, as we found a solution in #360. As soon as #365 is merged I'll rebase this PR and we can merge :tada:

davidmehren:
Rebased and ready :tada:

src/lib/web/note/util.ts
```diff
@@ -35,39 +70,11 @@ export function findNoteOrCreate (req, res, callback: (note: any) => void, inclu
...
-    owner = req.user.id
-  } else if (!config.allowAnonymous) {
-    return errors.errorForbidden(res)
-  }
-  if (config.allowFreeURL && noteId && !config.forbiddenNoteIDs.includes(noteId)) {
-    req.alias = noteId
-  } else if (noteId) {
-    return req.method === 'POST' ? errors.errorForbidden(res) : errors.errorNotFound(res)
+function isRevealTheme (theme: string) {
+  if (fs.existsSync(path.join(__dirname, '..', '..', '..', 'public', 'build', 'reveal.js', 'css', 'theme', theme + '.css'))) {
```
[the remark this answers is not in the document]
davidmehren: Fixed :heavy_check_mark:

SISheogorath:
Thanks!

src/test/csp.ts
```diff
@@ -31,22 +32,11 @@ describe('Content security policies', function () {
     }
   })
 
-  afterEach(function () {
```
DerMolly, line 34: Don't we need these Functions anymore?

PR #361, opened 2020-05-16
Tests and Fixes for Auth Middlewares

davidmehren:
This PR fixes a bug in the Dropbox auth middleware (broken import) and adds tests to make sure we avoid this type of bug in the future.

It also removes the custom handling of the Dropbox profile object in `parsePhotoByProfile`. The mail-address is now unwrapped in the `DropboxMiddleware` and `Profile.mail` is always a `string[]`.

Finally as `email` is always a string, we can get rid of a (broken) check in `generateAvatarURL`.

I tried to follow @SISheogorath s example and wrote longer commit messages to explain the changes :wink:

### DIFF
--- a/test/auth.ts
+++ b/test/auth.ts
@@ -1,0 +1,110 @@
+import assert from 'assert'
+import { ImportMock } from 'ts-mock-imports'
+import * as configModule from '../lib/config'
+import { DropboxMiddleware } from '../lib/web/auth/dropbox'
+import { EmailMiddleware } from '../lib/web/auth/email'
+import { FacebookMiddleware } from '../lib/web/auth/facebook'
+import { GithubMiddleware } from '../lib/web/auth/github'
+import { GitlabMiddleware } from '../lib/web/auth/gitlab'
+import { GoogleMiddleware } from '../lib/web/auth/google'
+import { LdapMiddleware } from '../lib/web/auth/ldap'
+import { OAuth2Middleware } from '../lib/web/auth/oauth2'
+import { OPenIDMiddleware } from '../lib/web/auth/openid'
+import { TwitterMiddleware } from '../lib/web/auth/twitter'
+
+describe('AuthMiddlewares', function () {
+  // We currently exclude the SAML Auth, because it needs a certificate file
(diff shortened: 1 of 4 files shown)


test/auth.ts
```diff
@@ -0,0 +1,123 @@
...
+        clientSecret: 'foobar'
+      }
+    }
+  }, {
+    name: 'OpenID',
+    middleware: OPenIDMiddleware,
+    config: {
+      openid: {
+        clientID: 'foobar',
+        clientSecret: 'foobar'
```
SISheogorath, line 108: :thinking: OpenID should not need a clientID or ClientSecret. OpenID is not oauth.

test/auth.ts
```diff
@@ -0,0 +1,123 @@
...
+      }
+    }
+  }, {
+    name: 'LDAP',
+    middleware: LdapMiddleware,
+    config: {
+      ldap: {
+        clientID: 'foobar',
+        clientSecret: 'foobar'
```
SISheogorath, line 77: ldap has a little bit different idea of how config should look like :D

test/auth.ts
```diff
@@ -0,0 +1,123 @@
...
+        scope: 'foobar'
+      }
+    }
+  }, {
+    name: 'Email',
+    middleware: EmailMiddleware,
+    config: {
+      email: {
+        clientID: 'foobar',
+        clientSecret: 'foobar'
```
SISheogorath, line 99: Same as with the other two.

test/auth.ts
davidmehren: Yep, my copypasting was a bit too fast ^^

test/auth.ts
davidmehren: Well, actually it doesn't :laughing: 

As long as `config.ldap` is defined (but it can be an empty object), the `LDAPStrategy` constructor works and we don't test any "deeper" ^^

test/auth.ts
davidmehren: also fixed

SuperSandro2000:
Can this branch be deleted?

Issue #360, opened 2020-05-16
Relative paths in CodiMD 2


davidmehren:
In https://github.com/codimd/server/pull/355 the following problem came up:

In some cases (like [the config](https://github.com/codimd/server/blob/master/lib/config/index.js#L13)) we get the app root path on the filesystem relative to the current file. This will not work anymore with TypeScript, as there are two ways of running CodiMD:
- With a 'normal' NodeJS, after running the TypeScript compiler `tsc`: The compiled JavaScript runs from the `built` subfolder. This is the default in our `npm start` script.
- With a NodeJS that can run TypeScript directly (`ts-node`). The code runs directly, without another subfolder in the path. This is handy during development, because you don't have to run `tsc` all the time.

I can come up with 3 ways so solve that issue:
1. Only support running the compiled version: This would entail adding one "path level" to find the app root directory. Consequently, running under `ts-node` would break.
2. Add some kind of heuristic to get the correct app root ("if there is a built directory, go 3 dirs up, else 2").
3. Don't use any relative paths to get the app root directory, but require the user to run CodiMD with the correct working directory.

I prefer option 3 because that eliminates potential breakage when refactoring (currently, if you move a file to another "folder level" you have to change any relative folder accesses in it and I don't think IDEs auto-fix our `path.resolve` calls?) and generally seems the cleanest to me. We are going to have breaking changes anyway so why not require the user so set the working directory to the CodiMD folder? We could do a check on startup if we find our `package.json` in the working directory and fail if that's not the case.

What do you think?

DerMolly:
I think the time to change this is the 2.0 release. People are most likely to read the how to update docs and catch, that they should configure a working dir in their setup. Especially if we make it very clear in the migrate document. Maybe we should start writing one now?

SISheogorath:
To me option 1 seems most appealing for a few simple reason: 

1. It's our decision to declare if `ts-node` is supported or not. 
2. We haven't supported `ts-node` yet, so it seems to be that we have to decide if the hassle is worth it
3. When we can solve a problem that requires manual interaction from a user otherwise, we should solve it.
4. `ts-node` seems to be a third-party solution of a problem that TypeScript itself doesn't create/have. And an easy workaround is the `tsc` watch mode + nodemon.

We should try to make configuring CodiMD correctly easier, not harder. And working directories are always tricky.

SuperSandro2000:
[...]

PR #365, opened 2020-05-20
moved lib and test into src directory

DerMolly:
as suggested by @davidmehren in #360


davidmehren:
Alright, I fixed the remaining path issues and cleaned up the commits a bit. From my pov this is now ready to merge :tada:

PR #368, opened 2020-05-24
Add a hotkey to hard-wrap the current paragraph or the selection

inducer:
This makes use of the `wrap` add-on that already comes with CodeMirror.

### DIFF
--- a/public/js/lib/editor/index.js
+++ b/public/js/lib/editor/index.js
@@ -9,6 +9,7 @@ import '../../../css/ui/toolbar.css'
 const isMac = CodeMirror.keyMap.default === CodeMirror.keyMap.macDefault
 const defaultEditorMode = 'gfm'
 const viewportMargin = 20
+const hardWrapColumn = 80
 
 const jumpToAddressBarKeymapName = isMac ? 'Cmd-L' : 'Ctrl-L'
 
@@ -121,30 +122,6 @@ export default class Editor {
       },
       'Shift-Ctrl-Backspace': cm => {
         utils.wrapTextWith(this.editor, cm, 'Backspace')
-      },
-      'Ctrl-\\': cm => {
-        var wrapOptions = {
-          wrapOn: /\s\S/,
-          column: 80
-        }
-
-        if (cm.somethingSelected()) {
-          var sels = cm.listSelections()
-          for (var i = 0; i < sels.length; ++sels) {
-            var head = sels[i].head
-            var anchor = sels[i].anchor
-
-            if (head.line < anchor.line) {
-              var temp = head
-              head = anchor
-              anchor = temp
-            }
-
-            cm.wrapRange(anchor, head, wrapOptions)
-          }
-        } else {
-          cm.wrapParagraph(cm.getCursor(), wrapOptions)
-        }
       }
     }
     this.eventListeners = {}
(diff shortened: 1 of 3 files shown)


public/js/lib/editor/index.js
```diff
@@ -121,6 +121,30 @@ export default class Editor {
       },
       'Shift-Ctrl-Backspace': cm => {
         utils.wrapTextWith(this.editor, cm, 'Backspace')
+      },
+      'Ctrl-\\': cm => {
```
SuperSandro2000, line 125: Is this <kbd>Ctrl</kbd> + <kbd>\\</kbd> ? That will most likely no work on German keyboard cause we don't have <kbd>\\</kbd> as a primary key.

inducer: All the more reason to switch to a US layout. :) I know what I'm talking about, I grew up with the German layout. It's terrible.

On a more serious note, got any suggestions? I'm happy to switch this to something that's a bit more globally compatible.

SuperSandro2000: If you can convince my muscle memory to change. :) If you know all the keys by heart you won't change anytime soon if you not really like change.

What about <kbd>Shift</kbd> + <kbd>Enter</kbd>?

inducer: OK with me, but that's widely used (Word uses it for a non-paragraph-breaking line break). If my suggestion was too off-Broadway, maybe Shift-Enter is too on-Broadway? <kbd>Ctrl</kbd>+<kbd>.</kbd> maybe?

inducer:
Does anyone know why Travis isn't completing?

SuperSandro2000:
It is green but status is not reported back to github. https://travis-ci.com/github/codimd/server/builds/167981182

public/js/lib/editor/index.js
SuperSandro2000: I am not sure about this.

SISheogorath:
I may ask the usually obvious question: Why? Like what problem do we try to solve here? And have we checked that this doesn't cause other issues (i.e. formatting issues)?

inducer:
> I may ask the usually obvious question: Why? Like what problem do we try to solve here?

I'll preface this by saying that I try to stick to 80 chars/line in any text document. I find long, wrapped lines difficult to navigate. Vertical cursor motion becomes more coarse-grained, while horizontal cursor motion remains inefficient. Plus indentation levels are less apparent. (Needless to say I'm a `breaks: false` person.)

This change helps me maintain that invariant in my documents without the need to do manual word wrapping.

Another reason to like short-ish lines: `diff` produces better output.

> And have we checked that this doesn't cause other issues (i.e. formatting issues)?

Not sure what you're looking for here. It's not perfect. It doesn't know about list items or blockqoute markers and will simply wrap them. But having this still beats re-wrapping a bunch of lines by hand.

InnayTool:
@DerMolly and I consider to merge this into version 1.7. The CI says the commits by @kaushikcfd don't comply with the DCO, so we can't merge. Because of the rewrite in 2.0, this will be the last release were it can be merged it. Do you think it's apart from DCO ready to merge?

InnayTool:
Thank you for your contribution! Since there was no further activity in this PR we decided to release 1.7rc1, which implies that this PR will not be merged. Unfortunately this means that there will be no release to merge it and I will close this PR.

PR #369, opened 2020-05-25
Make ESLint happy again

davidmehren:
This contains mostly more types at various places. We also excluded the migrations from eslint and fixed a bug in the note history updates

### DIFF
--- a/src/lib/utils.ts
+++ b/src/lib/utils.ts
@@ -4,6 +4,16 @@ import { logger } from './logger'
 import { Revision } from './models'
 import { realtime } from './realtime'
 
+/*
+Converts a map from string to something into a plain JS object for transmitting via a websocket
+ */
+export function mapToObject<T> (map: Map<string, T>): object {
+  return Array.from(map).reduce((obj, [key, value]) => {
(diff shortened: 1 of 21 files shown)


davidmehren:
After an evening of testing, a few more bugs surfaced and were fixed. @InnayTool and I are now mostly sure that the realtime stuff still works.

src/lib/utils.ts
```diff
@@ -4,6 +4,16 @@ import { logger } from './logger'
 import { Revision } from './models'
 import { realtime } from './realtime'
 
+/*
+Converts a map from string to something into a plain JSO for transmitting via a websocket
```
SISheogorath, line 8: Typo? Shouldn't it be JSON?

SISheogorath, line 8: Can we write it out? JSO was mistakenly considered a typo first, should be JS object.

davidmehren: done

SISheogorath:
@davidmehren feel free to press merge on this one :)

davidmehren:
:rocket:

PR #373, opened 2020-05-27
Update to mermaid 8.5.1

codingHahn:
I need to use the hide today marker in the gantt view of mermaid. This is supported since 8.5.0


SISheogorath:
LGTM!

codingHahn:
Thanks for the quick merge!
I think I forgot to update the checksums for the cdn scripts. Should I fix this or just leave it be, now that we have chosen to deprecate using a cdn?

SuperSandro2000:
Please update them.

PR #384, opened 2020-06-02
Run database migrations automatically on startup

SISheogorath:
Instead of using sequelize-cli and ensure migrations by shellscript,
this patch automates database migrations properly to the umzug library.
The sequelize CLI becomes a dev dependencies as it's still useful for
generating migrations.

This should eliminate the need for crude generating of database config
files and alike. Instead we utilize the pre-configured sequelize
connection that CodiMD will use anyway.

### DIFF
--- a/src/lib/models/index.ts
+++ b/src/lib/models/index.ts
@@ -1,5 +1,6 @@
 import { Sequelize } from 'sequelize-typescript'
 import { cloneDeep } from 'lodash'
+import * as path from 'path'
 import { Author } from './author'
 import { Note } from './note'
 import { Revision } from './revision'
@@ -7,6 +8,8 @@ import { Temp } from './temp'
 import { User } from './user'
 import { logger } from '../logger'
 import { config } from '../config'
+import Umzug from 'umzug'
+import SequelizeTypes from 'sequelize'
 
 const dbconfig = cloneDeep(config.db)
 dbconfig.logging = config.debug ? (data): void => {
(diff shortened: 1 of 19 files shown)


src/lib/models/index.ts
```diff
@@ -22,6 +26,47 @@ if (config.dbURL) {
...
+    logger.error(error)
+    logger.error('Database migration failed. Exciting…')
+    process.exit(1)
+  })
+  dbMigrationsComplete = true
+  logger.info('All migrations performed successfully')
+})()
+
+export async function waitForDBReady (): Promise<boolean> {
+  while (true) {
```
davidmehren, line 62: I really don't like waiting for a global variable that's set by an anonymous, self-executing async function. Can't we move the whole `await umzug.up().catch(...` codeblock directly into this function ?

src/lib/models/index.ts
```diff
@@ -22,6 +26,47 @@ if (config.dbURL) {
...
+  }
+})
+
+// Semicolon is required to prevent `TypeError: (intermediate value) is not a function`-error
+;(async () => {
+  // checks migrations and run them if they are not already applied
+  // exit in case of unsuccessful migrations
+  await umzug.up().catch(error => {
+    logger.error(error)
+    logger.error('Database migration failed. Exciting…')
```
davidmehren, line 54:
```suggestion
    logger.error('Database migration failed. Exiting…')
```

InnayTool: Very exciting Typo btw. ^^

src/lib/models/index.ts
SISheogorath: Good point… Not sure why I didn't come up with this one :D

PR #388, opened 2020-06-06
Refactor User model & profile-parsing

davidmehren:
This is really just moving around some code to separate the concerns of 'database model' and 'parsing passport profiles'.
I'm aware that we now have a separate class `PhotoProfile` that is not a model in the model folder, but I could not come up with a better place for it yet :smile:


SISheogorath:
Let's move the ProfileParser into some sort of util directory. I don't think it makes a lot of sense to keep it in the models directory.

davidmehren:
I created a new `utils` directory and moved `PhotoProfile` and the old `utils.ts` there.

SISheogorath:
Thanks a lot :)

PR #393, opened 2020-06-09
Update de.json

kleinhti:


SISheogorath:
Please use the official translation platform to submit translations: https://translate.codimd.org (you'll end up on POEditor)

Using direct PRs makes keeping everything in sync significantly harder. Thanks for your understanding :)

PR #396, opened 2020-06-09
Added serverVersion to status

InnayTool:

### DIFF
--- a/docs/dev/openapi.yml
+++ b/docs/dev/openapi.yml
@@ -782,6 +782,20 @@ components:
     ServerStatus:
       type: object
       properties:
+        serverVersion:
+          type: object
+          properties:
+            major:
+              type: integer
+            minor:
+              type: integer
+            patch:
+              type: integer
+            preRelease:
+              type: string
+            commit:
+              type: string
+          description: Version of the server specified according to SemVer specification
         onlineNotes:
           type: integer
           description: How many notes are edited at the moment.


docs/dev/openapi.yml
```diff
@@ -782,6 +782,9 @@ components:
     ServerStatus:
       type: object
       properties:
+        serverVersion:
+          type: string
+          description: Version of the server (commit or stable version number)
```
DerMolly, line 787:
```suggestion
          description: Version of the server (commit hash or stable version number)
```

ErikMichelson: Why commit hash or stable version number? If some tool/client wants to process this data, it's rather bad if the commit hash is returned for some instances and the version number for other ones.

SISheogorath: After consulting the git manual, I think commit itself is the correct way. As the hash is just a way to identify the commit object.

SISheogorath: @ErikMichelson the main problem is: Not every commit has a version, but every version has a commit. From that perspective it makes sense to either split it, or fallback to the more correct commit. This is what we do right now already when to figure it if clients are still compatible after a reconnect.

InnayTool: Maybe we should use `serverVersion` for the commit and add an attribute `serverVersionName` which contains the version string.

davidmehren:
Regarding the version number I googled around a bit and found that SemVer supports both pre-release info and build metadata: https://semver.org/#spec-item-9

I therefore propose a single version field with content as follows:
- For a release a 'version core': `1.6.0`
- For a pre-release a 'version core' with 'pre-release': `2.0.0-alpha1`
- For a non-released development version a 'version core', the 'pre-release' `next` and the commit as 'build' metadata: `1.6.0-next+deadbeef`.[1]

The 'quoted' identifiers are defined as in the [backus-naur form grammar](https://semver.org/#backusnaur-form-grammar-for-valid-semver-versions).

This way a client can easily differentiate the version of CodiMD it talks to.

[1] The 'version core' should also follow SemVer here. E.g. if we introduce a breaking change in the `master` branch without a release, the version should be *immediately* incremented accordingly. That way clients can still work out if they are compatible or not, even for development deployments.

SISheogorath:
@davidmehren `1.6.0-next+deadbeef` would be higher than `1.6.0-alpha1`. When we start to increment version numbers immediately, that would mean that a development version has a higher version number than the resulting alpha. It would also mean, we would change our "release management" significantly.

We could use `rc` instead, but since we have a stable master, I actually don't see the need for such a strict interpretation of SemVer. Clients should reload when a new version of the backend is detected. And for our created stable releases, we can decided based on the expected receiving for a user, if it's actually a breaking change or not.

davidmehren:
> `1.6.0-next+deadbeef` would be higher than `1.6.0-alpha1`

Hmm, good point. I just thought about the "master is always stable/always deployable" thing. Doesn't that imply that master should always follow the SemVer semantics? So every PR or commit that introduces a SemVer-relevant change also changes the version number? That way we could avoid having a state in master that has a version not representative of the actual behavior and get around my proposed "development releases/versions". I stumbled upon [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) that proposes a commit-message structure that helps automating just that.

If we don't want to have an always-correct SemVer in master, IMO it is necessary to mark 'in-between-releases' code. We could e.g. use `alpha` for that, so `alpha` would imply development code and would always be followed by a commit id. Another way (which would admittedly change our branching model a bit) would be to use a non-SemVer string (like `dev-<commit>`) in master and set a SemVer version only in an actual release branch.

> Clients should reload when a new version of the backend is detected.

That implies that server and client versions are always tightly coupled and that the reload always results in a client version that is completely compatible with the server.

In a situation where this is not the case (e.g. a third-party client or a separately hosted frontend), the client needs to detect if it supports the server version. Therefore the strict SemVer semantics are necessary.

SISheogorath:
> In a situation where this is not the case (e.g. a third-party client or a separately hosted frontend), the client needs to detect if it supports the server version.

To me this is an explicit non-goal. I think in order to keep our life simply, we should consider the CLI and maybe some sort of minor automations as third-party clients, but not a full featured frontend. The only actual supported frontend would be, what we ship with the backend. This takes a way a ton of headache.

This doesn't mean that one can't deliver the client separately, but I don't think we really make our lives easier when we make this a goal as it introduces a ton of headache and a API contract I'm actually not willing to be hold accountable for (i.e. forward or backwards compatibility for clients).

> I stumbled upon Conventional Commits that proposes a commit-message structure that helps automating just that.

Yes, conventional commits are a nice way to handle that. But personally I don't think they are well-suited for a free software project that is mainly developed by drive-by contributions. As we notice the sign-off already overwhelms the majority of contributors and conventional commits don't make this easier.

> If we don't want to have an always-correct SemVer in master, IMO it is necessary to mark 'in-between-releases' code.

I agree that we should clarify that. The most obvious solution would be to use the short-ref of a commit as build number (which is not entirely correct from a SemVer perspective but hey, we are old enough to make some decisions ourselves). But it actually raises one important question to me: Who is this version for? The frontend (i.e. for machine reability) or is it a string that should be displayed? In case of the latter, I don't think it's that important if it's 100% correct SemVer. In case of the latter: We should maybe create a version-type/object which looks like this:

```json
"version": {
 "major": 1,
 "minor": 2,
 "patch": 3,
 "commit": "<(optional) commit>"
}
```

This could solve both problems right away :)

davidmehren:
> The only actual supported frontend would be, what we ship with the backend.

I think this is a really important discussion to have for 2.0. To not bury it in this PR I opened #399 to gather more feedback :grinning:

InnayTool:
Should we merge this to the API? I think it should work even after the reworking we did recently.

PR #400, opened 2020-06-12
Fix allow anonymous edit documentation

david-sawatzke:
The default is false and the option only matters if allowAnonymous is false


SISheogorath:
Good catch! Thanks for this fix!

Issue #387, opened 2020-06-04
Release 1.7


davidmehren:
We accumulated a few fixes in `master` which we probably should push out in a patch-release before moving 2.0-development to `master`. To do that we should:

- [x] Bump the version
- [x] Write release notes
- [x] Make sure `yarn.lock` is up to date
- [x] Make sure translations are up to date
- [x] Run the tests described in the release test document (internal)
- [x] anything else?

SISheogorath:
The release test document is not really internal, as there aren't really any internal processes:

https://demo.codimd.org/dGE6iySoScy2qrkHMaBFaQ#

The idea is to have as many people as possible, grabbing a copy of this document, test it against the latest master and once this is done, giving feedback if things were broken.

davidmehren:
Test results for the `cookies_same_site` branch from #389:

- "Creating and working with documents": works! :heavy_check_mark: 
- "Database migrations": I don't have an 1.2 database :cry: 
- Testing `/features` in Firefox 78:
	- UML, MathJax, Codeblocks, TOC and embeddings work :heavy_check_mark: 
	- Still, I got some errors, which seem all to be CORS-stuff with Slideshare:

<details>
<summary>Details</summary>

```
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at https://public.slidesharecdn.com/fonts/fontawesome-webfont.woff2?v=4.3.0?cb=1590689192. (Reason: CORS header ‘Access-Control-Allow-Origin’ missing).

downloadable font: download failed (font-family: "FontAwesome" style:normal weight:400 stretch:100 src index:1): bad URI or cross-site access not allowed source: https://public.slidesharecdn.com/fonts/fontawesome-webfont.woff2?v=4.3.0?cb=1590689192
Cross-Origin Request Blocked: The Same Origin Policy disallows reading the remote resource at about:blank. (Reason: CORS request not http).

downloadable font: failed to start download (font-family: "MathJax_Blank" style:normal weight:400 stretch:100 src index:0): bad URI or cross-site access not allowed source: about:blank
```

</details>

- YAML Metadata: not tested
- GDPR features: Account deletion works :heavy_check_mark: , but a note that was being edited while the account was deleted still exists :red_circle:

oupala:
I'm not sure if this is the right place to mention a few proposal I mage, but I wanted to improve the documentation about configuration of codimd.

[...]

Issue #348, opened 2020-04-29
default values for environment variables


oupala:
The documentation is listing [environment variables](/codimd/server/blob/master/docs/configuration-env-vars.md) but does not provide the default value of each variable.

The default value, and a **required** flag should be specified for each variable.

oupala:
As my previous pull request #404 has been merged, are you interested in adding the default value for each variable?

oupala:
@ccoenen

ccoenen:
I'm currently not able to put in any meaningful amount of time. I'm really sorry, but this won't change any time soon.

oupala:
I was just asking for an opinion.

I offered to do it myself.

SISheogorath:
I don't think we should add another row to the tables. It's already super dense.

Also there is currently only the database config we require. Everything else can be left on default and CodiMD would work. It's recommended to set some more things, like `CMD_DOMAIN`, but technically it works without.

oupala:
@SISheogorath Your point is about a **required** column.

What about a **default value** column?

SISheogorath:
I don't think a defaults value row would make the tables less crammed. Looking at the GitHub rendering I already think that readability is questionable and would only decrease even further.

Therefore I advise against more column of any kind.

oupala:
My opinion is that readability worth nothing if mandatory information are not there.

Most of the configuration parameters are optional. But to guess if you want to override them or not, you have to know their default value, otherwise it is useless to offer default value without mentioning them in the documentation.

And if people have to look at the doc, it means that the documentation is missing some important informations. You can be a user of codimd, an admin of codimd and not being able to look at js code and understand it.

SISheogorath:
I'm not saying we shouldn't provide this data, but maybe such a table is simply the wrong representation. Maybe we should create a short overview table for config option instead? 

Or be creative in another way. I'm open for ideas here, I just don't see how another column would actually improve the documentation.

oupala:
Trying to be creative, I tought that we could add the default value as the first example on each `example` cell.

davidmehren:
I think we could even replace the "example value" column with a "default value"  column. The possible options could be mentioned in the description.

oupala:
I think that removing example values would not serve the users. I then decided to change the name of the *example* column to change its name from `example value` to `**default** and example value`.

It allows to get the default value in bold, and then one or multiple other example values. This solution does not add another column.

Please find this proposal in the following pull request #464.

PR #404, opened 2020-06-15
improve documentation about configuration

### LINKED ISSUE #349: Document CMD_S3_ENDPOINT ENV variable
The new environment variable``CMD_S3_ENDPOINT`` from #32 didn't get documented in the PR and will probably be missed by many people. Found by @oupala https://github.com/codimd/server/pull/32#issuecomment-621254222

oupala:
As discussed in #387 and with @davidmehren, here is a pull request to address some painpoints:

- I find it difficult to understand how to configure codimd, switching from environment variable to configuration file
  - there is also some inconsistency between both configuration solution (see #400 that fixes such inconsistency)
  - having to update two files while adding a configuration parameter is error prone (see #349 where a parameter hos not been documented)
- I am then proposing to merge two markdown files in one: configuration-config-file.md and configuration-env-vars.md
- I also fixed #349 and taken in account #400
- and finally I've added some linting default rule to have a consistent way of writing markdown

If the pull request seems ok, I'd like to imprive it by adding the default value of each parameter. This would ease the user experience while trying to figure out which parameter should be overidden or not. I've token about that in #348.

I'm now waiting for your feedback and I'm willing to discuss and change anything.


ccoenen:
I think the `.remarkrc` should be discussed seperately. (I'm not opposed to it, I just think it should be a separate thing, maybe even including it in our lint / test run on CI if it's good :-D)

ccoenen:
I'm somewhat reluctant about bunching the two files up into one, though. This was where we came from, and I don't think it helps readability to be honest. I would however wait for other opinions on this one.

oupala:
As I have linted a markdown file, I wanted to join the linting rules that I used for linting, in order to avoid potential linting-wars.

But I completely agree that markdown linting is a much wider subject than this pull request. I'm ok to remove the commit about markdown linting and to open another pull request about that.

I'm just waiting for other opinions.

By the way, why is the merging blocked? Is there some manual reviewing of the travis build?

SISheogorath:
For whatever reason the old travis integration is still coming up in some PRs. I haven't fully figured out yet, why this is the case. CI thinks your work is all good :)

docs/configuration.md
```diff
@@ -0,0 +1,267 @@
...
+### Lutim
+
+| config file | environment     | example value               | description                                                                 |
+| ----------- | --------------- | --------------------------- | --------------------------------------------------------------------------- |
+| `lutim`     |                 | `{"url": "YOUR_LUTIM_URL"}` | When `imageUploadType` is set to `lutim`, you can setup the lutim url       |
+|             | `CMD_LUTIM_URL` | `https://framapic.org/`     | When `CMD_IMAGE_UPLOAD_TYPE` is set to `lutim`, you can setup the lutim url |
+
+**Note:** *Due to the rename process we renamed all `HMD_`-prefix variables to be `CMD_`-prefixed. The old ones continue to work.*
+
+**Note:** *relative paths are based on CodiMD's base directory*
```
davidmehren, line 267: Maybe move this to the 'CodiMD paths stuff' section as it seems to be most relevant there? And then remove the footnote superscripts in that section.

docs/configuration.md
```diff
@@ -0,0 +1,267 @@
...
+| ----------------------------- | ------------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------ | ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- |
```
davidmehren, line 72: This line seems to break the table rendering, at least in Githubs preview.
Is a horizontal line even necessary here?

oupala: Note, this is a typo. I'll fix that.

docs/configuration.md
oupala: I moved the note about path from ` docs/configuration.md` to `docs/url-scheme.md` in b9c75a83f779a531bae666ce1b1afdb80cbee578.

I hope it was the file that you were talking about.

docs/configuration.md
oupala: Fixed in e46d520ca234c05cfe2431c5b7ab61b96194962e.

oupala:
> I think the `.remarkrc` should be discussed seperately. (I'm not opposed to it, I just think it should be a separate thing, maybe even including it in our lint / test run on CI if it's good :-D)

I removed the `.remarkrc` file in order to discuss it in another pull request.

oupala:
> I'm somewhat reluctant about bunching the two files up into one, though. This was where we came from, and I don't think it helps readability to be honest. I would however wait for other opinions on this one.

I'm not sure readability on the main point to look at. I think the current organization of the configuration document (2 separate files) is oriented on the **how** to configure codimd.

As a user, I was confronted to the problem of configuring codimd: some values can only be set by config file, some other can be configured by environment variables. It hard to guess what can be configured because of the way the documentation is organized: **how**. The content is splitted between two different files, for the same goal: configuring codimd.

With this pull request, I wanted to change the paradigm of the documentation to switch from **how** to configure codimd, to **what** can be configured on codimd.

Once you know **what** can be configured in codimd, each table tells you about **how** to configure this option.

I find this way of documenting more consistent for any user or admin of codimd.

In this comment, I wanted to explain why I created this pull request. Let discuss it further if needed.

docs/configuration.md
davidmehren: No, I was actually talking about [this section](https://github.com/oupala/server/blob/master/docs/configuration.md#L36) in the same file.

oupala: Ok, fixed in 46e86b1be3e81956dd9e42604f370f380fa1b63b.

SISheogorath:
Thank you very much!

PR #405, opened 2020-06-16
Support setting scope and state in Generic OAuth2

dalcde:
The scope option lets us set a custom scope to retrieve the name, email
and username.

Setting a state protects against CSRF and is required by some OAuth2
providers (e.g. ORY Hydra)

Signed-off-by: Dexter Chua <dalcde@yahoo.com.hk>

### DIFF
--- a/docs/configuration-config-file.md
+++ b/docs/configuration-config-file.md
@@ -152,7 +152,7 @@ these are rarely used for various reasons.
 
 | variables | example values | description |
 | --------- | ------ | ----------- |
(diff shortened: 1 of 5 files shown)


dalcde:
I just realized there is a release/2.0.x branch which already contains the scope option. Should I close this PR and write a new PR that adds `state` to `release/2.0.x`?

SISheogorath:
It would be a good idea to bring those changes to 2.0, yes.

But would you say the changes as they are here are a bug fix or more like a minor feature?

dalcde:
It's hard to tell; I believe generic oauth is unusable with most oauth providers otherwise without being able to set the scope. I don't have a strong stake in this though, as I'll probably end up using the 2.0 branch.

PR #407, opened 2020-06-16
Generic OAuth2: Set state: true

dalcde:
The OAuth2 specification RECOMMENDS setting the state to protect against
CSRF attacks. Some OAuth2 providers (e.g. ORY Hydra) refuse to
authenticate without the state set.

Signed-off-by: Dexter Chua <dalcde@yahoo.com.hk>


DerMolly:
@codimd/developers does someone already use the generic OAuth2 Provider and could test that this change (specifically the adding of `state: true`) won't break (your) existing setup(s)?

SISheogorath:
I do use it for my private setup. I'll see if I can run some testing tonight.

SISheogorath:
Thanks!

PR #410, opened 2020-06-17
doc: add a markdown linting configuration file

oupala:
Add a markdown linting configuration file. The configuration file is very similar to [remark-preset-lint-recommended](https://github.com/remarkjs/remark-lint#list-of-presets). This preset contains rules that prevent mistakes or syntaxes that do not work correctly across vendors.

If there is a consensus about this pull request and adding a markdown linting configuration file, I am willing to complete this pull request in order to lint all markdown files from the repo.


DerMolly:
How would one use the `.remarkrc` file or better which editor honors this file? This would be useful to know for future edits of the documentation and such...

oupala:
The `.remakrc` file is a configuration file for [remark-lint](https://github.com/remarkjs/remark-lint).

Most of modern editor can install remark-lint as an extension and use the config file to adjust its configuration.

This information would be welcomed in the `CONTRIBUTING.md` file.

ccoenen:
in case someone is looking for documentation on the rules themselves: they are in the individual readmes of these packages: https://github.com/remarkjs/remark-lint/tree/master/packages

ccoenen:
> The configuration file is very similar to [remark-preset-lint-recommended](https://github.com/remarkjs/remark-lint#list-of-presets). This preset contains rules that prevent mistakes or syntaxes that do not work correctly across vendors.

as far as I can see, one can "inherit" (for lack of a better word) rules from a preset like this: 

```js
"plugins": [
  "preset-lint-recommended",
  ["remark-lint-maximum-line-length", 8000] // this would override one of the settings
]
```

If we're based on one of the styles, I think this would make our changes to that much clearer.

oupala:
In fact, my configuration file was almost similar to remark-preset-lint-markdown-style-guide.

I've updated the file so that I directly use this preset.

We can now discuss this preset and override some rules if needed, as mentioned by @ccoenen.

I personally think that we should stick to the defaults of the preset.

But as codimd is based on markdown for writing, I'm pretty sure that one can have a different opinion about that topic.

ccoenen:
would you like to also introduce a `npm run mdlint` or `npm run remarklint` or some similar command to the `package.json`? This will also need some hints in the documentation on how to use it as a newbie.

Naming is of course up to you, I was trying to keep it in line with the existing linters we already have in there:

https://github.com/codimd/server/blob/c07a107b1f0b2e3b3f29870a852c2e05d2dbb9d6/package.json#L08-L11

If I understand things correctly, remark and the preset should probably be added as a dev-dependency. (I may well be wrong on this one, I'm still learning about this tool, I didn't know about it two days ago!)

oupala:
If we had to change some default rules, here is what I would suggest:

```js
{
  "plugins": [
    "preset-lint-markdown-style-guide",
    "preset-lint-recommended",
    ["lint-maximum-line-length", false],
    ["lint-list-item-indent", "space"]
  ]
}
```

oupala:
> would you like to also introduce a `npm run mdlint` or `npm run remarklint` or some similar command to the `package.json`? This will also need some hints in the documentation on how to use it as a newbie.

Yes, this would be needed.

> If I understand things correctly, remark and the preset should probably be added as a dev-dependency. (I may well be wrong on this one, I'm still learning about this tool, I didn't know about it two days ago!)

Again, you're right.

If you're ok with the idea, I'm gonna add the required changes.

ccoenen:
Why the `lint-list-item-indent: space`?

oupala:
> Why the `lint-list-item-indent: space`?

Mainly because I don't understand the meaning of the default `mixed` value of [lint-list-item-indent](https://github.com/remarkjs/remark-lint/tree/master/packages/remark-lint-list-item-indent).

But maybe someone can explain it to me.

DerMolly:
I guess that with the `mixed` the tool is okay with both 4 (`tab-size`) and 2 (`space`) space indentation and does not enforce one or the other. With `lint-list-item-indent: space` suggested by @oupala, remark should complain about all indendation of lists with more than 2 spaces. At least that's what I understand of the linked documentation. If I have time this evening I'd like to test my hypothesis (if that's still necessary by then).

oupala:
@DerMolly Please note that `remark-lint-list-item-indent` is about "spacing between a list item’s bullet and its content", not about indentation between a list a its sublist.

DerMolly:
I think that the `tab-size` options means 4 spaces in the continuation and that the initals item is as indented as that so.
```
1...bullet point one
....bullet point one (2nd line)
```

oupala:
@DerMolly As far I have tested, `["lint-list-item-indent", "tab-size"]` does not allow:

```md
-...multi
....line
....item
```

but it does allow:

```md
-.multi
..line
..item
```


It might be because of my `.editorconfig` file that says:

```
indent_style = space
indent_size = 2
```

Please, *do not dive* into space/tabs war... :smile:

I still think that `space` is a better value for `lint-list-item-indent` as *one simple* rule is more consistent than a *mixed* rule, and **1** space is always shorter than multiple spaces.

DerMolly:
I think `space` is fine. And since nobody else had anything to say about it, we should probably merge, @SISheogorath?

DerMolly:
Maybe we want to add the `markdownlint` command to the CI, too?

oupala:
> Maybe we want to add the `markdownlint` command to the CI, too?

[It is!](https://github.com/codimd/server/pull/410/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R11)

SISheogorath:
Please add a section to `travis-ci.yml` like we do with ESlint or jsonlint:

https://github.com/codimd/server/blob/c789b71cc3080a242858d30b5711aa7c43d41b2a/.travis.yml#L8-L12

Otherwise it looks good :)

oupala:
I'm not used to Travis CI and nor to pipeline management with GitLab. But I think that I've added the required code, and I hope this will be ok.

Do not hesitate to tell if something gets wrong...

.travis.yml
```diff
@@ -10,6 +10,12 @@ jobs:
         - 10
       script:
         - yarn run eslint
+    - stage: Static Tests
```
SISheogorath, line 13: You  don't need the `stage statement here, otherwise :+1:

oupala: Why is the `stage` statement useless here?

oupala:
I fixed the pull request according to your feedback.

oupala:
Should I also fix all the warnings in markdown documents in this pull request either?

oupala:
After disabling 2 useless rules (see discussion above), I fixed all markdown files to remove warnings.

I did not modified markdown files in `public/docs` directory as I don't know if it has any impact on codimd. Let me know if I should lint them also. In particular, I don't know how the release note `public/docs/release-notes.md` is maintained: it is automatically generated, or manually updated?

docs/guides/minio-image-upload.md
```diff
@@ -16,31 +15,31 @@ Minio Guide for CodiMD
...
 
    ![minio default view](../images/minio-image-upload/default-view.png)
 
-4. Create a bucket for CodiMD
+1. Create a bucket for CodiMD
 
    ![minio create bucket](../images/minio-image-upload/create-bucket.png)
 
-5. Add a policy for the prefix `uploads` and make it read-only.
+1. Add a policy for the prefix `uploads` and make it read-only.
```
ccoenen, line 34: Am I the only one who does not care for `1.`-everywhere for ordered lists?

I know why it is done, but I think it partly defeats the purpose of Markdown - being well-readable even without a renderer.

As before, I'd love to hear other comments, of course, and will gladly shut up if I'm the only one with that particular opinion.

ccoenen: (same comment applies to all files where numbered lists were swapped to be `1.` lists)

DerMolly: totally agree. making every entry `1.` defeats the purpose. Starting them with `1.` and having the correct followup numbers would be great though. Can that be done with remark?

ccoenen: https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-ordered-list-marker-value

this should be: [`remark-preset-lint-markdown-style-guide: ordered`](https://github.com/remarkjs/remark-lint/tree/main/packages/remark-lint-ordered-list-marker-value), the readme claims that this would be the default, but our base ruleset [`preset-lint-markdown-style-guide`](https://github.com/remarkjs/remark-lint/tree/main/packages/remark-preset-lint-markdown-style-guide#rules) overrides this.

oupala: See the [documentation about this rule](https://unifiedjs.com/explore/package/remark-lint-ordered-list-marker-value/).

The default value is `one`:

> When set to 'one', bullets should always be 1.

I think you'd like to set it to `ordered`?

I have a split opinion about the default value of `one`. The cons is that it makes the markdown less readable. But the pro is that it is very easy to maintain as you don't have to renumber the list when adding or removing an item from the list.

I personally prefer the `one` value because it is the best solution to avoid inconsistency in lists.

ccoenen: Their readme is somewhat weirdly structured, I also first thought it was `one` by default, but it's actually `ordered`:

grafik

The table below that lists the value that's overridden inside `remark-preset-lint-markdown-style-guide` (which we're basing off of).

oupala: There is the default value of the rule, and the default value of the preset. Both default values are not the same.

ccoenen: yes, this is exactly what I wrote here: https://github.com/codimd/server/pull/410#discussion_r448458446

docs/guides/providing-terms.md
```diff
@@ -1,23 +1,22 @@
-Setup your terms of use
-===
+# Setup
```
ErikMichelson, line 1: "Setup" sounds unfitting in my opinion as this file is explicitly only about setting up legal files like the terms.

docs/guides/minio-image-upload.md
DerMolly: I'm strongly against `one` as our devs and possibly our (admin) users will see these files in a plaintext editor rather than with a markdown preview.

docs/guides/providing-terms.md
oupala: Do you have any proposal?

Maybe "legal setup" or "legal terms setup"?

SISheogorath: Since this falls into the how-to guide category of documentation, it should be "How to set up your terms of use"

oupala: I've changed the title.

oupala: I've changed the title according to @SISheogorath's proposal.

docs/guides/minio-image-upload.md
oupala: I've added 2 commits to change `ordered-list-marker-value` from `one` to `ordered`. I can now squash or not the commits depending on the final decision about this default value.

I just want to repeat that I personally prefer the one value because it is the best solution to avoid inconsistency in list numering. There is already some inconsistency, such as the [keycloak page](/codimd/server/blob/master/docs/guides/auth/keycloak.md) where item 6 follows item 4 without any item 5:

> 4. Go to the client you set up in the previous steps using the Clients page, then choose the Client Scopes tab. Apply the scope you've created. This should mitigate errors as seen in [codimd/server#56](https://github.com/codimd/server/issues/56), as the `/userinfo` endpoint should now bring back the user's ID under the `id` key as well as `sub`.
> 
> ---
> 
> 6. In the `docker-compose.yml` add the following environment variables to `app:` `environment:`

This manual renumbering is also a pain, and is no more a pain when  `ordered-list-marker-value` is set to `one`.

davidmehren:
@oupala Could you run `yarn install` and commit the updated `yarn.lock`?
After that, Travis should be happy and we can merge this. :rocket:

oupala:
Why does the lockfile need to be updated? I have not modified the `package.json` file...

And besides, I want to delete or to squash commits 9cbdc69 and 1cdf87c before merging the pull request, depending on the conclusion of the thread above.

davidmehren:
> Why does the lockfile need to be updated? I have not modified the `package.json` file...

You added
```
"remark-cli": "^8.0.0",
"remark-preset-lint-markdown-style-guide": "^3.0.0",
```
in 3d0cdc91 :smiley: 

> _Note:_ I don't use yarn (I use npm) so don't know how yarn is working...

As this project uses only `yarn` (note that we only have a `yarn.lock` in the repo and no `package-lock.json`) I'm afraid you will need to also use that to update the lockfile. Otherwise our CI will refuse to run properly.

> And besides, I want to delete or to squash commits [9cbdc69](https://github.com/codimd/server/commit/9cbdc69694ae8cad818dc8cbd7e480a4a27aac69) and [1cdf87c](https://github.com/codimd/server/commit/1cdf87c2e3fe9f0a6946380e497b68b5b55323df) before merging the pull request, depending on the conclusion of the thread above.

I think most people (myself included) would prefer the `ordered` setting, because that way the raw markdown is more readable.

oupala:
The requested change has been changed. @SISheogorath can you please check that?

I also updated the yarn.lock file according to @davidmehren's request. You're right I had modified the `package.json` file but I could not remember it. ;-)

The only thing we have to decide is the value of `ordered-list-marker-value` : `one` (which is the default value of the preset) or `ordered` (which is the default value of the module). I have prepared both versions, I now have to know wich version I keep, by deleting or squashing some commit in order to keep a clean commit tree.

oupala:
Thanks @SISheogorath. Can you just tell me why is the `stage` statement useless on the travis file I changed?

Before merging, I'd like to know your answer on the two questions I asked yersterday.

*I don't know how to mark a pull request as a Work In Progress(WIP).*

oupala:
friendly ping @davidmehren

davidmehren:
Regarding `ordered-list-marker-value`: please keep the version using `ordered`.

Regarding `public/docs`: These are the documentation files we ship with CodiMD to explain some features. I don't see any harm in linting them, as long as that does not break the rendered version. As far as I can see the release notes are updated manually.

oupala:
I have left `ordered-list-marker-value` to the value `ordered`.

I have also linted the markdown files in the `public/docs` directory while keeping some syntaxe as the `features.md` is about showing all features, including some ways to write markdown that do not follow the markdown linter's best practices.

Feel free to merge this pull request whenever you want.

davidmehren:
Thanks!

PR #411, opened 2020-06-17
Allow for undefined email and displayName

dalcde:
OAuth2 allows the user to only consent to a subset of the scopes
requested. Previously, the Generic Oauth2 implementation assumes that
the `username`, `email` and `displayName` attributes are supplied, and
may crash if they are not defined.

This commit allows for `email` and `displayName` to not be defined,
either through the user refusing consent or the OAuth2 configuration
not asking for them in the first place (by not setting
`userProfile*Attr`).

If `email` is not provided, the `emails` property is simply left empty.
If `displayName` is not provided, it is left undefined, and CodiMD uses
the `username` whenever the `displayName` is expected.

This does not deal with the case where `username` is not provided. Since
usernames are not unique in CodiMD, it is possible to deal with this by
setting a dummy username. This can be added in a future commit if
desired.

### DIFF
--- a/src/lib/web/auth/oauth2/oauth2-custom-strategy.ts
+++ b/src/lib/web/auth/oauth2/oauth2-custom-strategy.ts
@@ -1,6 +1,7 @@
 import { InternalOAuthError, Strategy as OAuth2Strategy } from 'passport-oauth2'
 import { config } from '../../../config'
 import { PassportProfile, ProviderEnum } from '../utils'
+import { logger } from '../../../logger'
 
 function extractProfileAttribute (data, path: string): string {
   // can handle stuff like `attrs[0].name`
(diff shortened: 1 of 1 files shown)


src/lib/web/auth/oauth2/oauth2-custom-strategy.ts
```diff
@@ -15,14 +15,30 @@ function extractProfileAttribute (data, path: string): string {
...
+  }
+
+  const emails: string[] = []
+  try {
+    const email = extractProfileAttribute(data, config.oauth2.userProfileEmailAttr)
+    if (email !== undefined) {
+      emails.push(email)
+    }
+  } catch (e) {
+    // Leave emails empty
```
SISheogorath, line 34: Please no empty catch statements. At least log on a debug level that no email was supplied.

src/lib/web/auth/oauth2/oauth2-custom-strategy.ts
```diff
@@ -15,14 +15,30 @@ function extractProfileAttribute (data, path: string): string {
...
   const username = extractProfileAttribute(data, config.oauth2.userProfileUsernameAttr)
-  const displayName = extractProfileAttribute(data, config.oauth2.userProfileDisplayNameAttr)
-  const email = extractProfileAttribute(data, config.oauth2.userProfileEmailAttr)
+  let displayName: string | undefined
+  try {
+    // This may fail if the config.oauth2.userProfileDisplayNameAttr is undefined,
+    // or it is foo.bar and data["foo"] is undefined.
+    displayName = extractProfileAttribute(data, config.oauth2.userProfileDisplayNameAttr)
+  } catch (e) {
+    displayName = undefined
```
SISheogorath, line 24: Some debug logging that logging that no display name was provided would be nice.

dalcde:
I'm not sure if this is log-worthy. It could be a conscious decision by
the host.

SISheogorath:
Well, people aware of it can easily ignore it, but people unaware of it will have a bad time figuring out the problem. Especially when it might be an unexpected issue like a HTTP request returning a 403 even when it should return a 200.

dalcde:
Fair enough. I'll make it log level debug

dalcde:
I considered logging the entire `data`, but I'm not a huge fan of leaving users' personal data in the logs (it can also get very long)

src/lib/web/auth/oauth2/oauth2-custom-strategy.ts
```diff
@@ -15,14 +16,33 @@ function extractProfileAttribute (data, path: string): string {
...
+  const emails: string[] = []
+  try {
+    const email = extractProfileAttribute(data, config.oauth2.userProfileEmailAttr)
+    if (email !== undefined) {
+      emails.push(email)
+    } else {
+      logger.debug('\'id_token[%s]\' is undefined. Setting \'emails\' to [].', config.oauth2.userProfileEmailAttr)
+    }
+  } catch (e) {
+    logger.debug('\'id_token[%s]\' is undefined. Setting \'emails\' to [].', config.oauth2.userProfileEmailAttr)
```
SISheogorath, line 38: We should also print `e.message ` into the debug log as the request actually failed.

src/lib/web/auth/oauth2/oauth2-custom-strategy.ts
```diff
@@ -15,14 +16,33 @@ function extractProfileAttribute (data, path: string): string {
...
-  const displayName = extractProfileAttribute(data, config.oauth2.userProfileDisplayNameAttr)
-  const email = extractProfileAttribute(data, config.oauth2.userProfileEmailAttr)
+  let displayName: string | undefined
+  try {
+    // This may fail if the config.oauth2.userProfileDisplayNameAttr is undefined,
+    // or it is foo.bar and data["foo"] is undefined.
+    displayName = extractProfileAttribute(data, config.oauth2.userProfileDisplayNameAttr)
+  } catch (e) {
+    displayName = undefined
+    logger.debug('\'id_token[%s]\' is undefined. Setting \'displayName\' to \'undefined\'.', config.oauth2.userProfileDisplayNameAttr)
```
SISheogorath, line 26: Same over here, we should keep error messages around.

dalcde:
Done!

PR #413, opened 2020-06-18
Fully dedicate our soul to yarn

dalcde:
This removes npm dependency, and fixes warnings about `npm` and `yarn` using different copies of `node`.


dalcde:
Turns out this doesn't actually remove the npm dependency; `yarn install` still calls `npm`, presumably through a dependency.

SISheogorath:
It still tells the direction :wink:

PR #421, opened 2020-06-23
Centralize permission checking

dalcde:
This makes it more convenient to modify the permission model, both for
future modifications and for custom installations.

This changes the `owner` property of NoteSession to `ownerId`, which is
a more accurate description anyway.


SISheogorath:
As we already talked in the community chat, I wonder, do we want to continue the expansion of this PR in order to complete the changes to the permission system or do we want to merge this and start from the `release/2.0.x` branch to go further?

dalcde:
I would prefer if this is merged first. Any further development will necessarily require a lot of discussion and design, and may have to wait till `react-client` is ready. Personally, this would let me more easily and cleanly implement a rudimentary (and hard-coded) custom permission model on my own instance, and [I'm not the only person looking for this](https://github.com/codimd/server/pull/322#issuecomment-648165049)

PR #424, opened 2020-06-26
Fix browser logo links

SuperSandro2000:
browser logos switched to use main as their default branch which breaks the direct links.


SISheogorath:
Thanks!

SISheogorath:
This should also be files to master

PR #425, opened 2020-06-27
Improve handling of termination signals

### LINKED ISSUE #408: CodiMD prints lots of error messages with invalid postgres config
Steps to reproduce:
1. Install CodiMD
2. Set `CMD_DB_URL` to a postgres database, but with incorrect password
3. Run `node app.js`
4. The server panics because it fails to connect to the database.
5. Ctrl-C to kill the process
6. The process keeps printing lines of the form
```
2020-06-16T10:31:40.143Z error:         password authentication failed for user "codimd"
```
around 20 such messages per second, non-stop.

dalcde:
Previously, upon receiving a termination signal, the process tries to
flush all changes to the database, retrying every 0.1s until it
succeeds. However, if the database is not set up properly, this always
fails, and spams the terminal/logging with the error message 10 times a
second.

If the user sends another termination signal, the handleTermSignal
function is called once again, and we get twice the number of error
messages.

This commit changes the behaviour in various ways.

(1) It lowers the retry rate to 0.5s, and aborts after 30 seconds.

(2) If the write to the database errored, the error message explains
    that this is due to us flushing the final changes.

(3) We replace realtime.maintenance with realtime.state, which is an
    Enum with three possible states --- Starting, Running, and Stopping.
    If a termination signal is received in the starting state, the
    process simply aborts because there is nothing to clean up. This is
    the case when the database is misconfigured, since the application
    starts up only after connecting to the databse. If it is in the
    Stopping state, the handleTermSignal function returns because
    another instance of handleTermSignal is already running.

Fixes #408

### DIFF
--- a/src/lib/realtime.ts
+++ b/src/lib/realtime.ts
@@ -20,6 +20,11 @@ export type SocketWithNoteId = Socket & { noteId: string }
 
 const chance = new Chance()
 
+export enum State {
+  Starting,
+  Running,
+  Stopping
+}
 /* eslint-disable @typescript-eslint/no-use-before-define */
 const realtime: {
   onAuthorizeSuccess: (data, accept) => void;
@@ -27,7 +32,7 @@ const realtime: {
   io: SocketIO.Server; isReady: () => boolean;
   connection: (socket: SocketWithNoteId) => void;
   secure: (socket: SocketIO.Socket, next: (err?: Error) => void) => void;
-  getStatus: (callback) => void; maintenance: boolean;
+  getStatus: (callback) => void; state: State;
 } = {
   io: SocketIO(),
   onAuthorizeSuccess: onAuthorizeSuccess,
@@ -36,7 +41,7 @@ const realtime: {
   connection: connection,
   getStatus: getStatus,
   isReady: isReady,
-  maintenance: true
+  state: State.Starting
 }
 /* eslint-enable @typescript-eslint/no-use-before-define */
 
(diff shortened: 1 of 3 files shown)


src/lib/utils/functions.ts
```diff
@@ -51,8 +51,14 @@ export function processData<T> (data: T, _default: T, process?: (T) => T): T | u
 }
 
 export function handleTermSignals (io): void {
+  if (realtime.state === State.Starting) {
+    process.exit(0)
+  } else if (realtime.state === State.Stopping) {
```
SISheogorath, line 56: Let's split this into two independent ifs as the else block makes it just harder to read :)

src/lib/realtime.ts
```diff
@@ -20,14 +20,19 @@ export type SocketWithNoteId = Socket & { noteId: string }
...
+}
 /* eslint-disable @typescript-eslint/no-use-before-define */
 const realtime: {
   onAuthorizeSuccess: (data, accept) => void;
   onAuthorizeFail: (data, message, error, accept) => void;
   io: SocketIO.Server; isReady: () => boolean;
   connection: (socket: SocketWithNoteId) => void;
   secure: (socket: SocketIO.Socket, next: (err?: Error) => void) => void;
-  getStatus: (callback) => void; maintenance: boolean;
+  getStatus: (callback) => void; state: State;
```
SISheogorath, line 35: Can you please check if this has an impact on the Frontend? If I recall correct the frontend uses maintenance for some sort of state change. (Like making the note read-only.)

src/lib/utils/functions.ts
dalcde: Done

src/lib/realtime.ts
dalcde: This should be fine. We don't send the `realtime` object to the frontend, but an explicit `maintenance` message, which we still do. It's in the `handleTermSignal` function.

davidmehren:
Thanks! The `State` enum is really way cleaner than the `maintenance` boolean.

PR #434, opened 2020-07-03
Fix dmpWorker logging

### LINKED ISSUE #433: workers/dmpWorker ignores loglevel setting
This is tested on `release/2.0.x`. `workers/dmpWorker.ts` logs a lot of messages at the `debug` level, and this is always printed regardless of the `loglevel` setting. The reason is that this code runs as a child process instead of in the main thread, and uses a different global instance of the logger. This is a fresh global instance with the default `loglevel`, which is `debug`.

This can be reproduced by looking at notes revisions. This leaks notes contents into the logs, which is Bad.

dalcde:
dmpWorker is run as a childProcess, which is a completely separate
nodejs instance. As such, the `logger` it obtains is a separate instance
than the one in the parent. The parent reads the config file to
determine the log level, but the childProcess does not. So the log level
used in dmpWorker is always `debug`, regardless of the configuration
options.

In addition to polluting the logs, this is potentially a privacy issue,
because `dmpWorker` logs the diffs of notes at the `debug` level, which
will then enter the system logs.

This commit fixes this by making `dmpWorker` send any messages back to
the parent, who is responsible for logging. This also avoids any
potential race conditions between the two loggers writing to the same
output.

Fixes #433

### DIFF
--- a/src/lib/models/revision.ts
+++ b/src/lib/models/revision.ts
@@ -21,6 +21,7 @@ class Data {
   cacheKey
   error
   result
+  level
 }
 
 function createDmpWorker (): ChildProcess {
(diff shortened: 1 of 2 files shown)


davidmehren:
Thanks!

PR #437, opened 2020-07-04
Set lang-metadata from yaml as document language

ErikMichelson:
This PR fixes the problem that published CodiMD notes have a hardcoded `lang=en` html-attribute and therefore won't be correctly identified by external parsers or translation browser extensions.
If you set the `lang` [yaml-metadata](https://demo.codimd.org/yaml-metadata#lang), that value will be used instead of 'en' with this PR.


hexabinaer:
Reads like this hasn't been merged yet. **I'd be so happy to see that support coming.** We have a blind colleague who was about starting to work with CodiMD massively - until we found out that the screen reader will mess it up.

New to the CodiMD issue queue - please forgive me if I'm misunderstanding ...

SISheogorath:
@hexabinaer As far as I can tell it is merged, but not yet part of a stable release. If you are willing to take the risk (it's not a high risk) and run the latest master branch, you can get all this goodness already. Otherwise you have to wait for 1.6.1/1.7.0 as upcoming release.

hexabinaer:
Thanks for the reply :-)

Seems like I'll have to wait for the next stable because we're using a package through cloudron.io (at least that will be an auto-update that way).

Issue #313, opened 2020-03-26
Kubernetes installation docs point to hackmdio/codimd


dvob:
The Kubernetes installation documentation (docs/setup/kubernetes.md) points to the Helm chart stable/hackmd which uses the Docker image hackmdio/hackmd, which in turn is based on hackmdio/codimd.
I think it would be best to remove that documentation until there is a new chart for codimd/server?

SISheogorath:
I would actually love to see a new chart appearing, maybe based on the one linked, but also maybe not in the official helm repository, because I had to notice that upgrades to the official repository are rather slow.

dsbrng25b:
The official helm stable and incubator repository will soon be depracted either way.
The new way is to host your own chart repository (e.g. on github pages / releases). Then you can add your repository to https://github.com/helm/hub that you get listed on https://hub.helm.sh/

SISheogorath:
Would you like to get involved with this? It sounds like you are enough into K8s to develop the chart and I'm sure there would be other members in the community willing to help. I'm explicitly thinking of @unteem :wink:

dsbrng25b:
Yes, I can help with this. I think we need to decided the following things:
* **Git Repository**: In which repository do we want to have the helm chart source
* **Helm Repository**: We need a HTTP server to host the index.yaml and the versioned charts.
  * Option 1: Use Github Pages. This would result in a repository named https://codimd.github.io/<repo> or you could use a CNAME like helm.codimd.org. You can see an example on https://github.com/dsbrng25b/charts. In that case the index.yaml is created in the gh-pages branch and the versioned charts are uploaded as releases.
  * Any other HTTP server, depending on what kind of infrastructure you have

The chart itself will be pretty simple I think. We could either reuse the old one from stable/hackmd or maybe even better recreate it from scratch with `helm create` to have the newest best practices from helm in the chart.

SISheogorath:
@dsbrng25b I created a bare-bones repository which we can make available at helm.codimd.org later.

I invited you to have the right to work in this and if you need anything, feel free to reach out :)

Looking forward to your work!

unteem:
Sorry just catching up on this issue now...

On our side we do not use helm anymore, the chart we used was pretty much the same that the original one in `helm/stable`. For now we just use kustomize. But I'm currently developing kubernetes operators for rocketchat and nextcloud, codimd is the next one on the list ;). I'll let you know when I'm done, should be in the next couple of weeks.

[...]

PR #438, opened 2020-07-09
Update kubernetes.md with some more info

cfstras:
The kubernetes documentation used to point to stable/hackmd, which does not use CodiMD, but rather **hackmd**'s open source variant.
Since that is a different piece of software, add some pointers to the docs on how to select codimd and add configurations.

### DIFF
--- a/docs/setup/kubernetes.md
+++ b/docs/setup/kubernetes.md
@@ -3,4 +3,26 @@ Kubernetes
 
 To install use `helm install stable/hackmd`.
 
-For all further details, please check out the offical CodiMD  [K8s helm chart](https://github.com/kubernetes/charts/tree/master/stable/hackmd).
+In your `values.yaml`, set this to use this repo instead of HackMD:
+```yaml
+image:
+  repository: quay.io/codimd/server
+  tag: 1.6.0
+```
+Replace the version with the [latest release listed here](https://github.com/codimd/server/tags)
+
+Configuration values can be set using `extraVars`:
+```yaml
+extraVars:
+  - name: CMD_DOMAIN
+    value: pad.mydomain.com
+  - name: CMD_PROTOCOL_USESSL
+    value: "true"
+  - name: CMD_URL_ADDPORT
+    value: "false"
+  - name: CMD_LOGLEVEL
+    value: "info"
+  ...
+```
+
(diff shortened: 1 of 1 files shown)


SISheogorath:
Thanks! We wanted to fix that for a quite long time but no one seat down to fix it yet. :+1:  Great solution to the problem, looking forward to this being merged.

SISheogorath:
If you are into K8s and would like to help us out with an own helm chart, that would be also very appreciated.

SISheogorath:
Okay, so I did some testing and it went off-rail quite easily. Newer versions of helm no longer come with the stable repository pre-installed, means one has to run `helm repo add stable https://kubernetes-charts.storage.googleapis.com` before going for anything.

Next step is to provide a parameter for the name, means it has to be either `helm install codimd stable/hackmd` or `helm install -g stable/hackmd`.

We should also make it clear that people have to create the values file themselves. Either using the minimal version from your example or by using `helm show values stable/hackmd > values.yaml`

Otherwise this seems fine. An addition request for help for an own helm chart would be a bonus but wanted. Especially since this way of deployment will be EOL in November of this year.

SISheogorath:
I'm actually coming closer to the point of thinking that removing the docs for K8s entirely is the better idea and start from scratch with a new helm chart in future will provide everyone a better experience than piggybagging on the stone-age implementation of HackMD that was never properly maintained, sadly. No experience and recommending to use something else is better than recommending a broken solution.

cfstras:
The HackMD chart still gets updated... I was fairly happy that you had some docs, if there hadn’t been any I would have probably stuck with hackmd even if they had oauth bugs at the time...

We‘re are happily running Codi with the chart from stable. I can adapt their chart for Codi, but I can’t promise to keep it updated. However, I think you’d need some hosting for a chartmuseum because I doubt helm/stable will accept a chart for Codi if they already have hack and it’s „almost the same“.

SISheogorath:
I still think getting an own chart is better than referring to "someone else's chart", saying "just change the images and you'll be fine".

I'll probably look into writing an own helm chart for HedgeDoc during the weekend.

InnayTool:
Thanks again for your contribution. As there is an issue #313 for that and we won't merge this into 1.7.1, I close this PR. More work on kubernetes docs should be done on the develop branch.

PR #461, opened 2020-07-29
Add note about renaming

SISheogorath:
To make it more obvious for people, a note in the README as recommended
by @danyork seems like a good idea.

Signed-off-by: Sheogorath <sheogorath@shivering-isles.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -9,6 +9,8 @@
 CodiMD lets you create real-time collaborative markdown notes. You can test-drive
 it by visiting our [CodiMD demo server][codimd-demo].
 
+**Note**: CodiMD is in the process of [changing its name to HedgeDoc](https://github.com/codimd/server/issues/452).
+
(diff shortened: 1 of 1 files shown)


DerMolly:
maybe we should mention that this rename will happen in 1.6.1?

SISheogorath:
I think that's something people will notice :wink:

SuperSandro2000:
I would put the note in the first text line below the badges or make it bold. Right now you could easily miss it if you revisit the readme and don't read everything twice.

davidmehren:
I also think we should at least make the note bold, if not move it directly below the heading.

SISheogorath:
@davidmehren done

SISheogorath:
I think moving it more up doesn't necessarily help. People will get it at latest when the renaming is happening. Otherwise it's just a beforehand information anyway.

Issue #275, opened 2020-02-20
Localization issue: moment.js locale is not set


numero-six:
When `format` is called on moment, the locale is always the default (`en`). It break the UI localization when, for instance, there is a call like:

```js
moment(time).format('llll')
```

It just seems the `moment` is not confiured: calling

```js
moment.locale('fr');
```

to set the locale seems to fix that.

ccoenen:
Would you say this is a setting per user, or per server or should we just default to the selected language ... (or or or :-) )

numero-six:
@ccoenen i don't know but it must be consistent with the locale used when calling `__`.

```
<%= __('Export') %>
```

Most of the time, `format` is called with some translated string around, ie `'Started at ' + moment().format('llll');`, so locale used to translate 'Started at' should be the same.

numero-six:
I don't understand why the issue is marked as `enhancement`, it's clearly a `bug` one. The interface is broken, the user can be exposed to an UI being half localized, half not-localized.

It's not a minor issue: as you may know, a date formatted as "04/02/2020" may be interpreted as February 4th, or April 2nd whereas the reader is english or not. That could lead to serious misunderstandings for non-english speaking people and a negative UX.

> Would you say this is a setting per user, or per server or should we just default to the selected language...

I don't know what is the good choice but that's not the issue here. The locale can be chosen guessed/per user/per instance/whatever. In every case, we should never be confronted to a UI with some buttons translated and some not: the locale chosen should be applied everywhere.

ccoenen:
Well, we're all just humans doing this in our spare time. If it makes you happy I'll add "bug" instead.

ErikMichelson:
In the [frontend for version 2.0](https://github.com/codimd/react-client) this issue is fixed already. For the next patch version of 1.x I wrote a quick fix.

PR #468, opened 2020-08-13
[1.x] Fix: Set moment.js locale to users' locale (#275)

ErikMichelson:
This PR fixes the fact that moment.js was not set to the language the user selected.
This was reported as an issue by @numero-six in #275.

### DIFF
--- a/locales/en.json
+++ b/locales/en.json
@@ -124,5 +124,7 @@
     "Limited - Signed-in people can edit (forbid guests)": "Limited - Signed-in people can edit (forbid guests)",
     "Locked - Only owner can edit": "Locked - Only owner can edit",
     "Protected - Only owner can edit (forbid guests)": "Protected - Only owner can edit (forbid guests)",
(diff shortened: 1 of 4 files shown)


davidmehren:
I just tested this fix, but sadly there are a few problems:
- The `locale` cookie the editor uses to get the current language only gets updated by the "home page" (where the language dropdown is), when the chosen language is *actively changed*. When I normally visit a CodiMD instance, the home page automatically detects my language from the browser and uses that, but the cookie doesn't get set without me scrolling down and choosing another language from the dropdown.
- In https://github.com/codimd/server/blob/master/public/js/extra.js#L52 the words "created" and "changed" are hardcoded anyway, so the translation we archive with this PR is incomplete. It now shows a rather weird mixed translation (like "created vor 15 Minuten") at the top of notes.

Both issues are certainly fixable, I just don't know if it's worth the effort, if we already have the feature working in 2.0.

davidmehren (approved):
Nice, the status is even localized now :heart_eyes_cat:

PR #473, opened 2020-08-17
Api2.0 to develop

InnayTool:


SISheogorath:
It would be nice if there would be an explanation why this PR was closed (even when it might be obvious to people involved).

InnayTool:
I tried to merge, the new spec to the develop branch, but saw that I forgot the signed of by and decided to rebase the api spec to the develop branch in the git cli.

PR #486, opened 2020-08-27
[1.x] Add config option for cookie SameSite policy

ErikMichelson:
This PR adds a new config option "`cookiePolicy`" to set the SameSite value of cookies. This feature was requested in the matrix chat by `@rjsparks:matrix.org`.


lib/config/default.js
```diff
@@ -27,6 +27,7 @@ module.exports = {
     upgradeInsecureRequests: 'auto',
     reportURI: undefined
   },
+  cookiePolicy: 'strict',
```
SISheogorath, line 30: After reading my own commit we might want to go for lax by default. otherwise people might run CodiMD locally and end up with  non functional setups leaving a bad impression.

https://github.com/codimd/server/commit/3d1fab0512a828927f277796a0e23e8133230be5

ErikMichelson: Done in 387e6682

SISheogorath (approved):

davidmehren (approved):

PR #490, opened 2020-09-01
Document the backup process

autra:
This commit adds a short documentation detailing how to backup one's instance.

It looks obvious, but I'm still wondering whether I forgot something or not. This PR is a mean to answer this question for me and others in the future.

Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,6 +66,16 @@ To use CodiMD, your browser should match or exceed these versions:
 - ![Opera](https://raw.githubusercontent.com/alrra/browser-logos/HEAD/src/opera/opera_24x24.png) Opera >= 34, ![Opera Mini](https://raw.githubusercontent.com/alrra/browser-logos/HEAD/src/opera-mini/opera-mini_24x24.png) Opera Mini not supported
 - ![Android Browser](https://raw.githubusercontent.com/alrra/browser-logos/HEAD/src/android-webview-beta/android-webview-beta_24x24.png) Android Browser >= 4.4
 
+## Backup/restore your instance
+
+To backup codimd, you should:
+
+- backup your database
+- backup your custom config file if you have one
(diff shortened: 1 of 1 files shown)


autra:
Aaaannnd I guess I forgot about the configuration file

ccoenen:
Thanks for the writeup :-) that is certainly something that has come up every now and then and which we _should_ document.

I would like to suggest giving it its own file in `docs/guides/backup.md`.

README.md
```diff
@@ -66,6 +66,16 @@ To use CodiMD, your browser should match or exceed these versions:
...
 
+## Backup/restore your instance
+
+To backup codimd, you should:
+
+- backup your database
+- backup your custom config file if you have one
+- backup the upload folder (see the [uploadsPath](./docs/configuration.md#codimd-paths-stuff) config directive) 
+
+Restoring an existing instance of codimd is then just a matter of restoring these two things.
```
InnayTool, line 77: I think it has to be "three" things. ;)

InnayTool (changes requested):

README.md
autra: You're right! I removed the quantifier, it will scale better ;-)

I'm still not entirely happy with this sentence though...

InnayTool: I think it's okay like that.

InnayTool (approved):

autra:
> I would like to suggest giving it its own file in `docs/guides/backup.md`.

I don't know... For me it's about the same importance as the configuration section for instance, and I felt it was overkill to create a new page and link it here for such a small section. What do you think ?

ErikMichelson:
Although I like the idea of having a separate page, I'm also fine with the integration into the readme page as the section really isn't too long.

ccoenen:
Ok, fine as well :-)

autra:
Thanks!

PR #496, opened 2020-09-08
[1.x] Update matrix chat link from riot.im to app.element.io

### LINKED ISSUE #495: Update matrix chat  mentioned from riot to element
can someone change the matrix hyper link to

https://app.element.io/#/room/#hedgedoc:matrix.org

instead of riot? on the main page at the bottom but also on the info you can reach by clicking the question mark icon

ErikMichelson:
This PR changes the former riot.im links to new app.element.io while also changing the alias link from "codimd:matrix.org" to "hedgedoc:matrix.org".

Fixes #495


SuperSandro2000 (approved):

public/views/index/body.ejs
```diff
@@ -158,7 +158,7 @@
...
+                        <%- __('Follow us on %s and %s.', '<a href="https://github.com/codimd/server" target="_blank" rel="noopener"><i class="fa fa-github"></i> GitHub</a>, <a href="https://community.codimd.org" target="_blank" rel="noopener"><i class="fa fa-users" aria-hidden="true"></i> Discourse</a>, <a href="https://app.element.io/#/room/#hedgedoc:matrix.org" target="_blank" rel="noopener"><i class="fa fa-comments"></i> Riot</a>, <a href="https://social.codimd.org/mastodon" target="_blank" rel="noopener"><i class="fa fa-mastodon"></i> Mastodon</a>', '<a href="https://translate.codimd.org" target="_blank" rel="noopener"><i class="fa fa-globe"></i> POEditor</a>') %>
```
SISheogorath, line 161: We should also replace "Riot" with "Matrix" or "Element"

ErikMichelson: Good point. As matrix is the ecosystem and element just a client, I prefer replacing it with "matrix". Any other opinions?

ccoenen: One thing that would speak for element: everyone who knows the details about Matrix will know what element is and how to convert an element link to something they can use with their clients. Someone who does _not_ know what matrix is has a pretty shitty first encounter with it (my opinion). So I'd prefer Element over Matrix as it is easier to get started with.

DerMolly: As far as I understand the Link would be to element.io, but the discussion is about what the text should be. And there Matrix is a fine choice in my opinion.

ErikMichelson: The alternative would be to include both names like "Element (matrix)".

davidmehren: I think we could actually write "Matrix" and use the [matrix.to link](https://matrix.to/#/#hedgedoc:matrix.org?via=matrix.org&via=kif.rocks&via=chat.weho.st), as that got redesigned recently and is now way more usable than before.

SISheogorath: After seeing the redesign, yes please. We should use the matrix.to link.

davidmehren (approved):

PR #499, opened 2020-09-22
Adds convenience script to migrate to minio.

pierreozoux:


bin/migrate_from_fs_to_minio
```diff
@@ -0,0 +1,71 @@
...
+    });
+  });
+});
+
+// You can only select to see what it will do
+// sequelize.query(`SELECT regexp_replace(content, '${minioUrl}', '${config.serverURL}', 'g') FROM "Notes"`).then(([results, metadata]) => {
+// 	console.log(metadata);
+// })
+
+sequelize.query(`UPDATE "Notes" SET content = regexp_replace(content, '${config.serverURL}', '${minioUrl}', 'g')`); // if something goes wrong, then revert '${config.serverURL}', '${minioUrl}' by swapping their place in the query.
```
SISheogorath, line 71: That's a bit plain. It'll break all internal links. you should replace the uploads path:

`sequelize.query(`UPDATE "Notes" SET content = regexp_replace(content, '${config.serverURL}/uploads/', '${minioUrl}/uploads/', 'g')`);`

davidmehren:
Thanks for the script!
Could you also write a few words in our documentation about it?
Maybe just expand https://github.com/codimd/server/blob/master/docs/guides/minio-image-upload.md?

pierreozoux:
@davidmehren I added a bit of documentation with a big warning.

Hope it is enough.

(Sorry I can't do a proper script and take into account all edge cases, I just think it would have been useful for me as it is, that's why I'm sharing :) )

PR #500, opened 2020-09-22
Implement /notes route

davidmehren:
Closes #474

PR #502, opened 2020-09-26
Use custom logger

davidmehren:
Depends on #500

[The Nest docs](https://docs.nestjs.com/techniques/logger#use-external-logger) recommend using an external logging module for production.

Pino is a JSON-based logger, that allows for structured logging in production and pretty logging during development. `nestjs-pino` automatically configures `pino-http` to intercept requests and adds a unique id to every request, which aids traceability.

As we currently have no config system, pretty printing and other settings are hardcoded.

### DIFF
--- a/src/logger/console-logger.service.ts
+++ b/src/logger/console-logger.service.ts
@@ -1,0 +1,119 @@
+import { Injectable, Optional, Scope } from '@nestjs/common';
+import { isObject } from '@nestjs/common/utils/shared.utils';
+import * as clc from 'cli-color';
+
+@Injectable({ scope: Scope.TRANSIENT })
+export class ConsoleLoggerService {
+  private classContext;
+  private lastTimestamp: number;
+
+  constructor(@Optional() context?: string) {
+    this.classContext = context;
+  }
+
+  setContext(context: string) {
+    this.classContext = context;
+  }
+
+  error(message: any, trace = '', functionContext?: string) {
+    this.printMessage(
+      message,
+      clc.red,
+      this.makeContextString(functionContext),
+      false,
+    );
+    this.printStackTrace(trace);
+  }
+
+  log(message: any, functionContext?: string) {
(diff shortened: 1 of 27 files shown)


InnayTool (approved):

ccoenen:
In case of the http logger: the default should be to _not_ log the full ip address. Doing that is pointless most of the time, and on the other hand storing personally identifiable data you don't need/use is problematic.

Could you check if this is the case? And change it if neccessary?

davidmehren:
Very good point! Pino has [support for redactions](https://github.com/pinojs/pino/blob/master/docs/redaction.md), so I will look into redacting the request IP. Do we want to make that configurable, in case the IP is needed for debugging?

ccoenen:
as long as it's off by default, I think this would be the best way to do that :-)

SISheogorath:
Note: I would try to avoid JSON-based structured logging. As far as I know most people here don't run large logging systems that would benefit from it. But a lot of people run either docker or a systemd service, those types of simple logging systems, prefer just straight textline-based logging. We should try to keep that as default.

davidmehren:
Alright, change of plans. After playing around with Pino a bit more today, I got more and more unhappy, as features I wanted were not possible or only possible with implementing a JSON-to-Text prettifier. Therefore I decided to scrap JSON-based logging for now and go with a slightly modified version of the default Nest logger.

InnayTool (approved):

src/logger/console-logger.service.ts
```diff
@@ -0,0 +1,119 @@
...
+
+    const localeStringOptions = {
+      year: 'numeric',
+      hour: 'numeric',
+      minute: 'numeric',
+      second: 'numeric',
+      day: '2-digit',
+      month: '2-digit',
+    };
+    //TODO make timestamp optional
```
ccoenen, line 90: this TODO might be easily resolved.

PR #503, opened 2020-09-26
finished NoteUtil class

DerMolly:


src/notes/note.utils.ts
```diff
@@ -1,18 +1,53 @@
...
 
-  public static parseTags(note: Note): string[] {
-    // TODO: Implement method
-    return ['Hardcoded note tag'];
+  private static parseMetadata(content: string) {
+    const matches = content.match(/^---\n((?:.|\n)*)\n---.*/)
+    if (!matches || matches.length <= 1) {
+      return ''
+    }
+    return safeLoad(matches[1])
```
InnayTool, line 51: Do we need this functions for other things than parsing from note? Or why did you change the signature?

DerMolly: Because it would be much more cumbersome to fetch the current text from the `Note` object than just get it via the parameters. We would need to add the `getLastRevision()` function to `NoteUtils` or the `Note` object itself…

InnayTool:
We (@DerMolly, @davidmehren, @InnayTool ) have decided that it would be better to save title, tags and description as database fields and change it with an API call.
The frontend or a other client should parse the title, tags and description and send them as separate fields with the note content to the server.

PR #504, opened 2020-09-26
Added test functions for /me routes

InnayTool:
Note that they don't contain the functions from the services yet.


DerMolly:
all test are skipped. Was this done on purpose or did you forget to revert some debug changes?

InnayTool:
Yes, it was done on purpose because I realized that other things should be implemented first. I want to make them visible, so nobody implement them again without knowing that there was work done before.

davidmehren (approved):

PR #507, opened 2020-09-27
Routes/notes/plain body parsing

InnayTool:
Moved plain body parsing to decorator

### DIFF
--- a/src/api/public/notes/notes.controller.ts
+++ b/src/api/public/notes/notes.controller.ts
@@ -1,5 +1,4 @@
 import {
-  BadRequestException,
   Body,
   Controller,
   Delete,
@@ -8,14 +7,12 @@ import {
   Param,
   Post,
   Put,
-  Req,
 } from '@nestjs/common';
-import { Request } from 'express';
-import * as getRawBody from 'raw-body';
 import { ConsoleLoggerService } from '../../../logger/console-logger.service';
 import { NotePermissionsUpdateDto } from '../../../notes/note-permissions.dto';
 import { NotesService } from '../../../notes/notes.service';
 import { RevisionsService } from '../../../revisions/revisions.service';
+import { MarkdownBody } from '../../utils/markdownbody-decorator';
 
 @Controller('notes')
 export class NotesController {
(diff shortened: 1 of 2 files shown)


src/api/public/notes/notes.controller.ts
```diff
@@ -2,48 +2,51 @@ import {
...
+ * Implementation inspired by https://stackoverflow.com/questions/52283713/how-do-i-pass-plain-text-as-my-request-body-using-nestjs
+ */
+export const PlainBody = createParamDecorator(async (_, context: ExecutionContext) => {
+  // we have to check req.readable because of raw-body issue #57
+  // https://github.com/stream-utils/raw-body/issues/57
+  const req = context.switchToHttp().getRequest<import('express').Request>();
+  if (req.readable) {
+    return (await getRawBody(req)).toString().trim();
+  }
+  throw new BadRequestException('Body is not text/plain');
```
ErikMichelson, line 32: text/plain? I thought of text/markdown or is this a step forward for making the backend content-unaware.

InnayTool: I've changed it to two more precise error messages.

src/api/utils/plainbody-decorator.ts
```diff
@@ -0,0 +1,26 @@
...
+ */
+export const PlainBody = createParamDecorator(async (_, context: ExecutionContext) => {
+  // we have to check req.readable because of raw-body issue #57
+  // https://github.com/stream-utils/raw-body/issues/57
+  const req = context.switchToHttp().getRequest<import('express').Request>();
+  if (req.readable) {
+    return (await getRawBody(req)).toString().trim();
+  }
+  if (req.get('Content-Type') === 'application/json') {
+    throw new BadRequestException('Body must not be json!');
```
SISheogorath, line 22: Please correct me if I got that wrong: this code prevents clients from uploading `application/json` content but allows all others?

If that's the case, we might want to take an denied-unless-allowed approach instead.

InnayTool: We don't want to prevent from uploading something, but if you upload json with Content-Type application/json, NEST parses it and getRawBody will fail. There is an [issue](https://github.com/stream-utils/raw-body/issues/57) which describes that.

davidmehren: Maybe "we only accept one content-type" (like `text/plain`) is better than "we accept everything but `application/json`"?

davidmehren (approved):

src/api/utils/plainbody-decorator.ts
InnayTool: changed to text/markdown

src/api/public/notes/notes.controller.ts
InnayTool: Now it's changed to text/markdown

PR #509, opened 2020-09-29
Makes the mime also work with upper case extension

pierreozoux:
If you upload images like .JPG, this fails without this fix.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -7,7 +7,7 @@ exports.isSQLite = function isSQLite (sequelize) {
 exports.getImageMimeType = function getImageMimeType (imagePath) {
   var fileExtension = /[^.]+$/.exec(imagePath)
 
-  switch (fileExtension[0]) {
+  switch (fileExtension[0].toLowerCase()) {
     case 'bmp':
       return 'image/bmp'
     case 'gif':


lib/utils.js
```diff
@@ -7,7 +7,7 @@ exports.isSQLite = function isSQLite (sequelize) {
 exports.getImageMimeType = function getImageMimeType (imagePath) {
   var fileExtension = /[^.]+$/.exec(imagePath)
```
SuperSandro2000, line 8: I think this would be a better place to apply the ``.toLowerCase`` but the 1.6 code is not in development as much anymore so maybe this also works.

davidmehren: I think it doesn't matter much. You would need to also move the `[0]` array access up, so the current version moves less code.

davidmehren:
@pierreozoux Thank you for you contribution!
Before I can merge this PR, you need to sign your commit. Have a look at https://github.com/codimd/server/blob/master/CONTRIBUTING.md#sign-your-work for details. In this case, `git commit --amend -s` and a force push should do the trick.

davidmehren:
@pierreozoux :wave: Are you still interested in working on this and #499?

pierreozoux:
I signed, sorry for delay :/

PR #535, opened 2020-10-19
fix(oauth2): add state parameter

haslersn:
This is a port of: https://github.com/hackmdio/codimd/pull/1512

### DIFF
--- a/lib/config/default.js
+++ b/lib/config/default.js
@@ -93,7 +93,8 @@ module.exports = {
     tokenURL: undefined,
     clientID: undefined,
     clientSecret: undefined,
-    scope: undefined
+    scope: undefined,
+    state: undefined
   },
   facebook: {
     clientID: undefined,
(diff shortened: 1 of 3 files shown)


davidmehren:
In #407 (merged to `release/2.0.x`) the `state` parameter was always set to `true`. Do you see any situation where someone might want to disable sending the state?

haslersn:
No, I don' see any situation where someone might want to disable it. I was not aware of #407. I created this PR, because my CodiMD (I think version 1.6.0) didn't work with ORY Hydra.

haslersn:
Why isn't #407 in the master branch?

davidmehren:
`master` contains (stable) code that will go into the next point release (probably 1.7). `release/2.0.x` as well as `develop` contain unstable code that will become the 2.0 release in the future.

As we already merged the same feature with `state` hardcoded to `true`, I'm open to either doing that here or at least defaulting to `true` instead of `undefined`.

haslersn:
I think that #407 should be considered as a bugfix and should be backported as far as you backport stuff. That is, it should definitely land in 1.7.

This PR on the other hand can be closed, except somebody else needs it. I don't need it.

PR #547, opened 2020-10-29
Upgrade various dependencies

davidmehren:
I read the changelogs of most of the minor upgrades to spot anything critical and tested locally, but please also test this yourself ;)

Note that I skipped the [codemirror](codemirror-spell-checker) upgrade, because it broke our vendored `codemirror-spell-checker`.


SISheogorath:
For codemirror the main concern seems to be a regex-dos. Regex-dos can be a problem as they might freeze your browser. But let's look at the context: It would be a DoS attack on a note in edit mode. Not great but more than fine for most cases. Therefore I'm not too concerned about it, the impact is quite small.

davidmehren:
@mrdrogdrog already took care of the codemirror upgrade in #551 :smile:

SISheogorath (commented):
I think we might need to update a lot of the pinning and continue to use `~` instead of `^` as if I recall correct we didn't drop CDN support.

https://github.com/codimd/server/pull/202

davidmehren:
I now also updated the CDN versions of our libraries to the version we use in the locally-built bundle.

SISheogorath (approved):

PR #548, opened 2020-10-30
Use URL constructor instead of regex to check for valid URL

### LINKED ISSUE #545: use of credentials in embeded PDF doesn't work
I have a PDF which I'm trying to embed using the syntax:

```
{%pdf https://myserver.net/public/PDF/Document.pdf %}
```

This works.  However, the PDF is stored on a WebDAV server which requires authentication.
When I try to embed the authentication information:

```
{%pdf https://public:public@myserver.net/public/PDF/Document.pdf %}
```

The PDF tag, then doesn't load.

I guess that something is failing in parsing the `user:password@` stanza.

davidmehren:
Fixes #545
Basic auth-regex from https://gist.github.com/dperini/729294

Co-authored-by: Yannick Bungers <git@innay.de>
Signed-off-by: David Mehren <git@herrmehren.de>

### DIFF
--- a/public/js/extra.js
+++ b/public/js/extra.js
@@ -178,16 +178,11 @@ function slugifyWithUTF8 (text) {
 }
 
 export function isValidURL (str) {
-  const pattern = new RegExp('^(https?:\\/\\/)?' + // protocol
-        '((([a-z\\d]([a-z\\d-]*[a-z\\d])*)\\.)+[a-z]{2,}|' + // domain name
-        '((\\d{1,3}\\.){3}\\d{1,3}))' + // OR ip (v4) address
-        '(\\:\\d+)?(\\/[-a-z\\d%_.~+]*)*' + // port and path
-        '(\\?[;&a-z\\d%_.~+=-]*)?' + // query string
-        '(\\#[-a-z\\d_]*)?$', 'i') // fragment locator
-  if (!pattern.test(str)) {
+  try {
+    const url = new URL(str)
+    return ['http:', 'https:'].includes(url.protocol)
+  } catch (e) {
     return false
-  } else {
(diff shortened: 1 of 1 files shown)


ErikMichelson (approved):

public/js/extra.js
```diff
@@ -179,6 +179,7 @@ function slugifyWithUTF8 (text) {
 
 export function isValidURL (str) {
   const pattern = new RegExp('^(https?:\\/\\/)?' + // protocol
+        '(?:\\S+(?::\\S*)?@)?' + // basic auth
```
SISheogorath, line 182: How about we simplify the whole function to a native browser interface (which is missing from IE11, but otherwise fine):

```
export function isValidURL (str) {
  try {
    new URL(str)
  } catch(e) {
    return false
  }
 
  return true
}
```

Because the regex you provide there will eat itself, probably ending up in yet another RegexDoS.

davidmehren: Great idea, done!

ErikMichelson:
I don't think that this change made it better. The URL-constructor allows just too much to be useful in this context. For example the following "URL" is perfectly valid with this approach: `somerandomwords:`.
Maybe at least add a check for the protocol to be either `http`, `https` or `ftp`.

SISheogorath:
@ErikMichelson Good point. But at least we can now use the object representation and no longer need to parse things ourselves :D

davidmehren:
I added a check for the used protocol, but decided to support `http` and `https` to not accidentally break someones HTTP-only deployment.

SISheogorath (approved):
LGTM much cleaner :)

ErikMichelson (approved):

Issue #96, opened 2019-05-30
Provide prebuilt installation


AndiLeni:
Is it possible to provide a prebuilt installation. Maybe even with a gui to install? 
I'm using a shared hosting service and the possibilities to built codimd on the server are very limited.

SISheogorath:
We currently support fully automated installations on Heroku or Cloudron as well as providing a container image that can run on any hosting platform that supports container images using `docker` or `podman`.

https://github.com/codimd/server#installation--upgrading

I don't think we have the resources to provide more installation methods. But we are welcoming people to join us and taking care of them :)

ccoenen:
I started a step by step guide, but I never really got around to finishing it:

https://demo.codimd.org/svq0JObCToqyt3ZBCZ58og#

Our current setup instructions only concern CodiMD itself. That document there would include database, a user and system service with SystemD.

Not automated, but maybe a step in that direction.

SISheogorath:
https://docs.travis-ci.com/user/deployment/releases/ <-- this should already help with this issue.

We also need to pack the repository without `.git` and we can ask ourselves if we want to upload `node_modules` or not (iirc we compile some things as part of the npm dependencies, so probably it's better not to go for it). And we should also only get the production dependencies.

ccoenen:
as soon as there's compiled stuff in `node_modules`, packaging them it is probably not a good idea. Also this might clash with their respective licenses if we pre-package them.

ErikMichelson:
Maybe it's worth to have a look at this GitHub action: https://github.com/actions/upload-release-asset#example-workflow---upload-a-release-asset

The advantage is that it integrates directly with GitHub and does not need a 'machine user account' on GitHub. ([1])

[1]: https://docs.travis-ci.com/user/deployment/releases/#authenticating-with-an-oauth-token

SuperSandro2000:
> does not need a 'machine user account' on GitHub like travis does.

Not one we need to provide. It technically stills uses one.

DerMolly:
This is fixed by now, isn't it. Then I'll close this issue

PR #585, opened 2020-11-17
Added zipping and uploading of production builds

ErikMichelson:
### Component/Part
CI

### Description
This PR adds zipping and uploading of the production builds so that one can just download, extract, configure and run them without the need to build the code. This is especially useful for low-end devices that suffer from too less available memory for building.

The workflow is extended by the following jobs:
- install production dependencies => dev-deps get removed
- zip everything together except the .git directory
- Upload the zip file as artifact

### Related Issue(s)
#96

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -78,5 +78,9 @@ jobs:
             ${{ runner.os }}-yarn-
       - run: yarn --frozen-lockfile --prefer-offline
       - run: yarn run build
-
-
+      - run: yarn install --production --ignore-scripts --frozen-lockfile --prefer-offline
+      - run: zip -5 -r prebuild-node${{ matrix.node-version }}.zip ./ -x '*.git*'
+      - uses: actions/upload-artifact@v2
+        with:
+          name: Prebuild with node ${{ matrix.node-version }}
+          path: ./prebuild-node${{ matrix.node-version }}.zip


SISheogorath:
You should also spare node_modules. The frontend is the heavy part. Installing dependencies afterwards is fine, doing the frontend build, overwhelms some machines. As node_modules can and does contain binary content, I wouldn't pack that as it's most likely incompatible with the machine you unpack it on.

.github/workflows/node.js.yml
```diff
@@ -78,5 +78,8 @@ jobs:
             ${{ runner.os }}-yarn-
       - run: yarn --frozen-lockfile --prefer-offline
       - run: yarn run build
-
-
+      - run: zip -5 -r prebuild-node${{ matrix.node-version }}.zip ./ -x '*.git*' -x 'node_modules/*'
```
SISheogorath, line 81: I would use `tar -cvpzf prebuild-node${{ matrix.node-version }}.tar.gz  --exclude=prebuild-node${{ matrix.node-version }}.tar.gz  --exclude=node_modules  --exclude=.git ` as this preserves permissions such as the executable bit (which we use in `bin/`).

SISheogorath (dismissed):

.github/workflows/node.js.yml
ErikMichelson: Well, I'm open to use tar instead of zip, I just thought that maybe zip is even more well-known than tar but as we're targeting linux-platforms anyway, zip really doesn't make any sense. I'm gonna change it.

davidmehren:
I would actually *only* include the built frontend files (so `public/build` and `/public/views/build`) in the bundle. When we also include all other files, setups that cloned via Git might get confused.

SISheogorath:
@davidmehren How would an install of the bundle then look like?

(Currently it would be `tar -xvf prebuild-node${{ matrix.node-version }}.tar.gz && cd <dir> && npm install && npm start`)

SISheogorath:
I also think the release process should only run, on releases/git-tags, not on pushes and pull requests. (Just in case you didn't already plan to change it at the end of this PR.)

ErikMichelson:
> I also think the release process should only run, on releases/git-tags, not on pushes and pull requests. (Just in case you didn't already plan to change it at the end of this PR.)

I thought quite some time about that. My first intention was to run it just on git-tags, yes. But then I thought that there are probably people that want to run the latest master-branch directly and since we build it anyways and GitHub doesn't charge us for storage (at least not that I know of), I decided to keep the build results as artifacts.

davidmehren:
> @davidmehren How would an install of the bundle then look like?

Well, you would install normally (either by cloning the repo or downloading a .tar.gz) but instead of `yarn run build` you extract the prebuilt frontend bundle.

Granted, if you use a .tar.gz to install, you could also just extract a complete bundle. But we currently instruct people to preferably use Git to install, which needs a frontend-only tar.gz.

davidmehren:
We could also provide both variants: one frontend-only bundle for Git users and one full bundle for users not using Git.

InnayTool:
In respect to the deployment of version 2.0 it would be convenient to have a "pre-compiled" version of the whole software as a tarball. I think for devices which don't have enough RAM, the default deployment should be extracting the tarball.

davidmehren:
For simplicity, what about just stopping to recommend the "git clone deployment" for production environments? Then, nobody can get confused about which tarball to extract when/where.

davidmehren:
I made `tar` happy by creating the file it writes to first. Now the CI job works, but GitHub seems to always zip the artifacts, so the download now is a .zip with another .tar.gz inside :D

I think I'll just remove all `tar`ing of files and let GitHub handle the compression.

davidmehren:
Without `tar` the upload task now takes about 2m30s instead of 15s :disappointed: 
We could also `tar` all files (without gzip compression) so the upload is faster and we avoid double compression, but extracting a zip and then another tar is still not great...
Using `tar` would also preserve file permissions, although I'm not sure if we need execute bits anywhere...

davidmehren:
We probably want to restrict the artifact uploads to `push` on `master` and the latest Node release, so we don't generate ridiculous amounts of artifacts.

mrdrogdrog:
Imho: If you restrict the build to `push on master` then it doesn't matter if the packaging needs 15s or 2m

davidmehren (approved):

SISheogorath:
So if I got this correct, GitHub forces zipping for uploaded artefacts and therefore if you upload a tar, you'll get a `.tar` file in a `.zip` file?

And this action basically uploads the entire directory content except of `node_modules` and the `.git` directory.

davidmehren:
@SISheogorath Correct.

After merging this, I would update our installation docs to recommend installing with the new bundle (at least on memory-constrained systems).

SISheogorath:
Could we still get proper tar.gz bundles for actual releases? If I see it correctly, the forced zipping is a restriction of GitHub actions, while release artifacts are not subject to those.

mrdrogdrog:
Upload artifact is only for CI artifacts. If you want to use the github release then you must use an action like https://github.com/marketplace/actions/create-release (or handle the github api yourself)

davidmehren:
I don't think our release schedule warrants a CI job for creating a bundle on release tags :D
I'd be fine if we merged this now as is. Either somebody cares enough to add bundling for releases before we release 1.7 or we add a .tar.gz manually.

PR #595, opened 2020-11-21
Add oauth2 authorization

joachimmathes:
### Component/Part
oauth2 provider

### Description
The oauth2 provider is missing a basic authorization concept. That's why I propose this PR. The implementation is based on the saml provider: https://github.com/hedgedoc/hedgedoc/blob/929d4af2b5f9564c16074f4425a1ff693300f04e/lib/web/auth/saml/index.js#L37

1. I added an attribute to specify the claim which is supposed to hold an array of strings of role names: `CMD_OAUTH2_ROLES_CLAIM`.
2. I added an attribute to specify a dedicated role which has to be included in the role claim of the ID token to become authorized.
3. Besides that I decided to allow another optional user profile attribute as `id`, which overrides `userProfileUsernameAttr`, if set: `userProfileIdAttr`.
  I introduced that attribute because a `username` might be unique, but nevertheless it is also prone to change. Imagine usernames which are built from parts of your first and lastname, either automatically or due to company guidelines. If people want to change their name, because of marriage or whatever reason else, the username no longer provides a stable ID. Thus, it's not only uniqueness but also immutability, which makes a good ID, e.g. a `uuid` provided by another claim

**Example configuration**
```
CMD_OAUTH2_ROLES_CLAIM = roles
CMD_OAUTH2_ACCESS_ROLE = role/codimd
CMD_OAUTH2_USER_PROFILE_ID_ATTR = user_uuid
```

Please note:
* I am no JavaScript developer. So, the code style of my modifications is merely an educated guess. :slightly_smiling_face: 
* I'm sorry, according to the contribution documentation I should have created an issue first. But the code was already written. :shrug:

### Steps

- [x] Added implementation
- [ ] Added / updated tests

### Related Issue(s)
This is the analogous PR of https://github.com/hackmdio/codimd/pull/1626.


lib/web/auth/oauth2/index.js
```diff
@@ -31,6 +32,7 @@ class OAuth2CustomStrategy extends Strategy {
         return done(new Error('Failed to parse user profile'))
       }
 
+      checkAuthorization(json)
```
davidmehren, line 35: This call also needs the `done` parameter

lib/web/auth/oauth2/index.js
```diff
@@ -91,7 +106,7 @@ passport.use(new OAuth2CustomStrategy({
   callbackURL: config.serverURL + '/auth/oauth2/callback',
   userProfileURL: config.oauth2.userProfileURL,
   scope: config.oauth2.scope,
-  state: true
+  state: true,
```
davidmehren, line 109: This comma needs to be removed :D

davidmehren (changes requested):
Thank you for your contribution!
Just two small things I noticed

lib/web/auth/oauth2/index.js
joachimmathes: The curse of JSON. :smile:

davidmehren (approved):

lib/web/auth/oauth2/index.js
```diff
@@ -50,18 +52,30 @@ function extractProfileAttribute (data, path) {
...
     username: username,
     displayName: displayName,
     email: email
   }
 }
 
+function checkAuthorization (data, done) {
+  const roles = extractProfileAttribute(data, config.oauth2.rolesClaim)
+
+  if (config.oauth2.accessRole && roles) {
```
SISheogorath, line 71: I wouldn't check if roles is set. Otherwise one might be able to craft a Profile that doesn't contain roles, resulting in getting access.

```suggestion
  if (config.oauth2.accessRole) {
    if (!roles) {
      logger.error('oauth2: "accessRole" configured, but user profile doesn\'t contain roles attribute. Permission denied')
      return done('Permission denied', null)
    }
```

lib/web/auth/oauth2/index.js
```diff
@@ -50,18 +52,30 @@ function extractProfileAttribute (data, path) {
...
+    if (!roles.includes(config.oauth2.accessRole)) {
+      logger.error('oauth2 permission denied')
```
SISheogorath, line 73: I would adjust this error message a bit:

```suggestion
      logger.error('oauth2: user doesn't have the required role. Permission denied')
```

Without being aware of the quickest way to do that, I would also recommend to add the username into the server-side log in order to make debugging more pleasant.

SISheogorath (changes requested):
Some additional remarks on how to improve logging and secure fallback situations.

lib/web/auth/oauth2/index.js
ErikMichelson: But in that case the username should be logged to the `debug`-output only. Don't put privacy-related information into a non-debug log.

SISheogorath (approved):

ErikMichelson (approved):

PR #644, opened 2020-12-31
added enums to app.config.ts

DerMolly:
### Component/Part
config

### Description
This PR adds enums to the config and started to implement auth backends

### Steps

- [x] Added implementation


DerMolly:
For auth provider that should be able to be specified more than once, I adopted the following schema
The different configs are stored in their own env vars suffiexed with `_0` through `_4`. 
e.g: `HD_AUTH_LDAP_PROVIDER_NAME_0`, `HD_AUTH_LDAP_PROVIDER_NAME_1` and so on

I think having 5 of each should be sufficient for the time being. If we really need more in the future this can easily be extended up to `_9`

davidmehren:
I don't think its sensible to add all these config options at this time.
For many of them, it's questionable if we will even need them (like `linkifyHeaderStyle`, `useCDN`, `tooBusyLag`).
We might also want to structure all the top-level options a bit.
I would rather add options as we implement the code that needs them, then we can see which options are actually needed.

Regarding the auth provider env vars: I *really* hate this solution, but suppose there aren't many other ways, besides throwing JSON into an env var or only supporting one auth provider per type via env vars...

DerMolly:
> Regarding the auth provider env vars: I really hate this solution, but suppose there aren't many other ways, besides throwing JSON into an env var or only supporting one auth provider per type via env vars...

Yeah, I'm also not to happy. I thought long about if there is something better to do, but I wasn't able to come up with something better myself.

DerMolly:
> I don't think its sensible to add all these config options at this time.

I commented the new top level options out for the time being. They can be readded as we need them or deleted before the 2.0 release, which ever their fate might be.

davidmehren:
When I try to start the app with no config it crashes:
```
$ nest start
[Nest] 381953   - 06.01.2021, 19:49:09   [NestFactory] Starting Nest application...
[Nest] 381953   - 06.01.2021, 19:49:09   [ExceptionHandler] Cannot read property 'split' of undefined +64ms
TypeError: Cannot read property 'split' of undefined
    at Object.exports.toArrayConfig (/home/david/Projekte/hedgedoc/server/dist/src/config/utils.js:5:25)
    at InstanceWrapper.metatype (/home/david/Projekte/hedgedoc/server/dist/src/config/app.config.js:278:47)
    at Injector.instantiateClass (/home/david/Projekte/hedgedoc/server/node_modules/@nestjs/core/injector/injector.js:293:55)
    at callback (/home/david/Projekte/hedgedoc/server/node_modules/@nestjs/core/injector/injector.js:77:41)
    at async Injector.resolveConstructorParams (/home/david/Projekte/hedgedoc/server/node_modules/@nestjs/core/injector/injector.js:118:24)
    at async Injector.loadInstance (/home/david/Projekte/hedgedoc/server/node_modules/@nestjs/core/injector/injector.js:81:9)
    at async Injector.loadProvider (/home/david/Projekte/hedgedoc/server/node_modules/@nestjs/core/injector/injector.js:38:9)
    at async Promise.all (index 4)
    at async InstanceLoader.createInstancesOfProviders (/home/david/Projekte/hedgedoc/server/node_modules/@nestjs/core/injector/instance-loader.js:43:9)
    at async /home/david/Projekte/hedgedoc/server/node_modules/@nestjs/core/injector/instance-loader.js:28:13
```

DerMolly:
@mrdrogdrog suggested changing the array structure in this way:

Specify which instances are configured via e.g `HD_AUTH_GITLABS = "test1, test2"` and then you define the needed variables via `HD_AUTH_GITLAB_TEST1_PROVIDER_NAME` and so on.

This of course would yield an on the fly building of the config object. Schema and config interface are not affected, as they – just like now – only know about a list of configs.

What do you think @davidmehren, @InnayTool?

davidmehren:
That sounds way better than the numbered env vars.
Can we still validate these dynamic vars?

InnayTool (approved):

DerMolly:
> That sounds way better than the numbered env vars.
> Can we still validate these dynamic vars?

I think so.

InnayTool (approved):

PR #646, opened 2021-01-02
changed kubernetes setup doc

DerMolly:
### Component/Part
docs

### Description
This PR improves the kubernetes documentation.

### Steps

### DIFF
--- a/docs/setup/kubernetes.md
+++ b/docs/setup/kubernetes.md
@@ -2,7 +2,7 @@
 
 HedgeDoc currently does not support any deployment via Kubernetes.
 
-We would love help with that, so if you want to write this, please feel free to chat us up on [Matrix](https://chat.hedgedoc.org/), so we can provide you with a repo.
+If you want to help us creating a helm chart, then feel free to contact us on [Matrix](https://chat.hedgedoc.org/) or [Discourse](https://community.hedgedoc.org/).
 
 ~~To install use `helm install stable/hackmd`.~~
 
--- a/docs/setup/kubernetes.md
+++ b/docs/setup/kubernetes.md
@@ -3,7 +3,3 @@
 HedgeDoc currently does not support any deployment via Kubernetes.
 
 If you want to help us creating a helm chart, then feel free to contact us on [Matrix](https://chat.hedgedoc.org/) or [Discourse](https://community.hedgedoc.org/).
-
-~~To install use `helm install stable/hackmd`.~~
-
-~~For all further details, please check out the offical hackmd/CodiMD [K8s helm chart](https://github.com/kubernetes/charts/tree/master/stable/hackmd).~~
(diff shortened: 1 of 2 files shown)


docs/setup/kubernetes.md
```diff
@@ -1,5 +1,9 @@
 # Kubernetes
 
-To install use `helm install stable/hackmd`.
+HedgeDoc currently does not support any deployment via Kubernetes.
 
-For all further details, please check out the offical HedgeDoc  [K8s helm chart](https://github.com/kubernetes/charts/tree/master/stable/hackmd).
+We would love help with that, so if you want to write this, please feel free to chat us up on [Matrix](https://chat.hedgedoc.org/), so we can provide you with a repo.
```
mrdrogdrog, line 5: Well. I don't know. I would write something like.

```suggestion
If you want to help us creating a helm chart, then feel free to contact us on [Matrix](https://chat.hedgedoc.org/) or [Discourse](https://community.hedgedoc.org/).
```

mrdrogdrog (changes requested):

mrdrogdrog (approved):

docs/setup/kubernetes.md
```diff
@@ -1,5 +1,9 @@
 # Kubernetes
 
-To install use `helm install stable/hackmd`.
+HedgeDoc currently does not support any deployment via Kubernetes.
 
-For all further details, please check out the offical HedgeDoc  [K8s helm chart](https://github.com/kubernetes/charts/tree/master/stable/hackmd).
+If you want to help us creating a helm chart, then feel free to contact us on [Matrix](https://chat.hedgedoc.org/) or [Discourse](https://community.hedgedoc.org/).
+
+~~To install use `helm install stable/hackmd`.~~
```
davidmehren, line 7: I would completely remove these lines.
Why should we first write "we don't support Kubernetes" and then mention a helm chart for a different project?

davidmehren (changes requested):

docs/setup/kubernetes.md
DerMolly: I was not to sure if we want to remove it, so I opted to just strike through the old kubernetes documentation. But I'm more than willing to throw it out all together.

mrdrogdrog: In this case I would only remove the last line.

DerMolly: I think this would yield more confusion than be helpful

mrdrogdrog: Why?.. `We don't support kubernetes. If you want to help us then contact us.` Why should that be confusing?

DerMolly: the file has three lines and I guess you just saw the one I added and `
~~To install use 'helm install stable/hackmd'.~~`. But yes remove all the old documentation seems reasonable. Let me do that.

davidmehren (approved):

ErikMichelson (changes requested):
If we state here that we don't support Kubernetes, we should remove it from the list of supported ways in the README as well.

DerMolly:
Good point.

ErikMichelson (approved):

PR #703, opened 2021-01-09
NotesController: Do not crash on nonexistent notes

davidmehren:
### Component/Part
NotesController

### Description
This PR adds proper error handling and returns 404 when a note does not exist.
Previously, we leaked the `NotInDBError` and sent a 500 status code.

### Steps

- [x] Added implementation
- [x] Added / updated tests

### DIFF
--- a/src/api/public/notes/notes.controller.ts
+++ b/src/api/public/notes/notes.controller.ts
@@ -10,12 +10,13 @@ import {
   Delete,
   Get,
   Header,
+  NotFoundException,
   Param,
   Post,
   Put,
 } from '@nestjs/common';
+import { NotInDBError } from '../../../errors/errors';
 import { ConsoleLoggerService } from '../../../logger/console-logger.service';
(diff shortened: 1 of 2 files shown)


src/api/public/notes/notes.controller.ts
```diff
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ * SPDX-FileCopyrightText: 2020 The HedgeDoc developers (see AUTHORS file)
```
DerMolly, line 2: I think this is wrong

DerMolly (changes requested):

src/api/public/notes/notes.controller.ts
davidmehren: fixed

InnayTool (approved):

DerMolly (approved):

PR #709, opened 2021-01-10
Notes API: Improve spec compliance

### LINKED ISSUE #702: PUT /notes/{id} does not return any content
**Description**
Sending a PUT request to a /notes/{id} URL that gets accepted with a status code 200 does not contain any data inside the response even though the API documentation says it should return the updated note metadata.

**To Reproduce**
Send new content to a note via the PUT on URL /notes/{id} with a valid Note-ID.

**Expected behavior**
I get a response with content documented in the API-doc.

**Alternatives to consider**
Update the documentation to say that a PUT does not return any content. This would conflict with how any other "change" to the system returns content though and would just be inconsistent behaviour.

**Your Setup (please complete the following information):**
 - Host OS: Win10
 - NodeJS version: v14.15.0
 - HedgeDoc version: current dev branch (no frontend, so I did not look up any version number)

davidmehren:
### Component/Part
Notes API

### Description
This PR improves spec compliance of the note API by returning the changed note on `PUT` requests to `/notes/{noteid}`.
It also moves the route to update note permissions under the `metadata` path. That location seems more sensible, as the metadata object includes the permissions.

### Steps

- [x] Added implementation
- [x] Added / updated tests

### Related Issue(s)
Fixes #702

### DIFF
--- a/docs/content/dev/public_api.yml
+++ b/docs/content/dev/public_api.yml
@@ -346,7 +346,7 @@ paths:
           content:
             text/plain:
               example: my-note
-  /notes/{note}/permissions:
+  /notes/{note}/metadata/permissions:
     put:
       tags: [ note ]
       summary: Set permissions of a note
--- a/src/api/public/notes/notes.controller.ts
+++ b/src/api/public/notes/notes.controller.ts
@@ -115,7 +115,7 @@ export class NotesController {
     }
   }
 
-  @Put(':noteIdOrAlias/permissions')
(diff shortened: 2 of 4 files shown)


InnayTool:
I am not sure by now if it is the best solution to make the API route longer. Where do we need this permissions? Maybe it's better to have a more flat object structure and move the permissions out of the metadata object.

InnayTool (approved):

PR #733, opened 2021-01-15
config: split config up

DerMolly:
### Component/Part
config

### Description
This PR splits the config up. This is necessary to correctly work with nestjs.

### Steps

- [x] Added implementation
- [x] Added / updated tests


InnayTool (approved):

InnayTool (approved):

src/media/media.service.spec.ts
```diff
@@ -39,7 +40,7 @@ describe('MediaService', () => {
       imports: [
         ConfigModule.forRoot({
           isGlobal: true,
-          load: [appConfigMock],
+          load: [appConfigMock, mediaConfigMock],
```
davidmehren, line 43: `appConfigMock` can be removed

test/public-api/media.e2e-spec.ts
```diff
@@ -31,7 +32,7 @@ describe('Notes', () => {
       imports: [
         ConfigModule.forRoot({
           isGlobal: true,
-          load: [appConfigMock],
+          load: [appConfigMock, mediaConfigMock],
```
davidmehren, line 35: `appConfigMock` can be removed

test/public-api/notes.e2e-spec.ts
```diff
@@ -27,7 +28,7 @@ describe('Notes', () => {
       imports: [
         ConfigModule.forRoot({
           isGlobal: true,
-          load: [appConfigMock],
+          load: [appConfigMock, mediaConfigMock],
```
davidmehren, line 31: `appConfigMock` can be removed

davidmehren (changes requested):
In addition to the bits I commented on, this PR exacerbates an existing problem:  
The error messages when the validation of the config fails are unhelpful. Previously, they didn't reference the name of the environment variable, but you had an idea how that could look like when the error complained about `database.username` not being present.

Now, the error will only say that `username` is missing, which is completely not helpful.
We need to improve the messages here ASAP.

davidmehren (approved):

PR #736, opened 2021-01-16
auth: adds token-auth to public api 

DerMolly:
### Component/Part
public api

### Description
This PR adds token-auth to the public api

### Steps

- [x] Added implementation
- [x] Added / updated tests


DerMolly:
Firstly this PR should be merged when #733 was merged.
Secondly I'm not to sure how to get the tests to work in this PR. Maybe @davidmehren or @InnayTool could give me some pointers?

I'm especially puzzled by this: 

```
 FAIL  src/auth/auth.service.spec.ts (5.414 s)
  ● AuthService › should be defined

    Nest can't resolve dependencies of the IdentityRepository (?). Please make sure that the argument Connection at index [0] is available in the TypeOrmModule context.

    Potential solutions:
    - If Connection is a provider, is it part of the current TypeOrmModule?
    - If Connection is exported from a separate @Module, is that module imported within TypeOrmModule?
      @Module({
        imports: [ /* the Module containing Connection */ ]
      })

      at Injector.lookupComponentInParentModules (../node_modules/@nestjs/core/injector/injector.js:188:19)
      at Injector.resolveComponentInstance (../node_modules/@nestjs/core/injector/injector.js:144:33)
      at resolveParam (../node_modules/@nestjs/core/injector/injector.js:98:38)
          at async Promise.all (index 0)
      at Injector.resolveConstructorParams (../node_modules/@nestjs/core/injector/injector.js:113:27)
      at Injector.loadInstance (../node_modules/@nestjs/core/injector/injector.js:46:9)
      at Injector.loadProvider (../node_modules/@nestjs/core/injector/injector.js:68:9)
          at async Promise.all (index 5)
      at InstanceLoader.createInstancesOfProviders (../node_modules/@nestjs/core/injector/instance-loader.js:43:9)

  ● AuthService › should be defined

    expect(received).toBeDefined()

    Received: undefined

      26 | 
      27 |   it('should be defined', () => {
    > 28 |     expect(service).toBeDefined();
         |                     ^
      29 |   });
      30 | });
      31 | 

      at Object.<anonymous> (auth/auth.service.spec.ts:28:21)
```

src/auth/auth.service.ts
```diff
@@ -0,0 +1,16 @@
+import { Injectable } from '@nestjs/common';
+import { UsersService } from '../users/users.service';
+import { User } from '../users/user.entity';
+
+@Injectable()
+export class AuthService {
+  constructor(private usersService: UsersService) {}
+
+  async validateToken(token: string): Promise<User> {
+    const user = await this.usersService.getUserByAuthToken(token);
```
InnayTool, line 10: As I stated in #738 the token should be hashed and this has effect on this line because it has to be hashed here.

InnayTool (changes requested):
First a comment the requested change follows below.
@davidmehren and I thought about having a simple way to define the privilege level which is required to call a Route. This can be:
- read
- write
- create
- own/admin
and beside from that
- locked in
Maybe this levels should be specified as annotations or by calling the right permission check function.

src/auth/auth.service.ts
DerMolly: Which hashing algorithm should be used here? bcrypt?

DerMolly:
How would such a annotation work? Won't that need context from the function?

src/auth/auth.service.ts
DerMolly: fixed with bcrypt.

InnayTool: yes bcrypt should be fine I think.

InnayTool:
Not necessarily it can be induced in the function by adding it to the request or something but I think call the correct function is the better and cleaner solution.

src/auth/auth.service.ts
InnayTool: We will use sha256 because we don't need a salt and gives no benefit and makes it harder to get which user owns the auth token

InnayTool (dismissed):

DerMolly:
I'll close this PR and continue with #738, as it's better to combine those two instead of building on two PRs in parallel.

PR #738, opened 2021-01-17
Add token support

DerMolly:
### Component/Part
private api

### Description
This PR adds the private api and the token routes.

This can be merged after #733 was merged.

### Steps

- [x] Added implementation


src/api/private/tokens/tokens.controller.ts
```diff
@@ -0,0 +1,54 @@
...
+  @Get()
+  async getUserTokens(): Promise<AuthTokenDto[]> {
+    // ToDo: Get real userName
+    return (await this.usersService.getTokensByUsername('molly')).map((token) =>
+      this.usersService.toAuthTokenDto(token),
+    );
+  }
+
+  @Post()
+  async postToken(@Body() label: string): Promise<AuthTokenWithSecretDto> {
```
InnayTool, line 39: I think postToken is confusing because you don't post a token only a token request.
So it should be named postTokenRequest.

InnayTool (commented):
This PR adds some database fields, we have to update the db-schema.plantuml.

DerMolly:
> This PR adds some database fields, we have to update the db-schema.plantuml.

This PR is still a Draft 😉

InnayTool:
We should hash the auth token in database because it's working like a password.

InnayTool:
> > This PR adds some database fields, we have to update the db-schema.plantuml.
> 
> This PR is still a Draft wink
yes I only want to mention it because it can easiely be forgotten.

src/api/private/tokens/tokens.controller.ts
DerMolly: fixed

docs/content/dev/db-schema.plantuml
```diff
@@ -24,11 +24,13 @@ entity "User" {
   email : text
 }
 
-entity "AuthToken" as authToken{
+entity "AuthToken" {
   *id : number <<generated>>
   --
   *userId : uuid
   *accessToken : text
+  *identifier: test
```
InnayTool, line 32:
```suggestion
  *identifier: text
```

src/users/users.service.ts
```diff
@@ -26,8 +33,30 @@ export class UsersService {
...
+    const user = await this.getUserByUsername(userName);
+    let accessToken = '';
+    for (let i = 0; i < 100; i++) {
+      try {
+        const randomString = crypt.randomBytes(64).toString();
+        accessToken = await this.hashPassword(randomString);
+        await this.getUserByAuthToken(accessToken);
+      } catch (NotInDBError) {
+        const token = AuthToken.create(user, identifier, accessToken);
+        return this.authTokenRepository.save(token);
```
InnayTool, line 49: The random string has to be returned and send to the user which can be hashed if it is send back to the user. After that the server does not know the random string anymore only the hash of it.

InnayTool (changes requested):

src/users/users.service.ts
```diff
@@ -26,8 +33,30 @@ export class UsersService {
...
 
+  async createTokenForUser(
+    userName: string,
+    identifier: string,
+  ): Promise<AuthToken> {
+    const user = await this.getUserByUsername(userName);
+    let accessToken = '';
+    for (let i = 0; i < 100; i++) {
+      try {
+        const randomString = crypt.randomBytes(64).toString();
```
ErikMichelson, line 44: This will just convert random bytes to a string, including invalid character points and control characters. Even if the `hashPassword` method is resistent to this, we should keep in mind that an access token probably only should contain printable characters.

My suggestion is to use either the hexadecimal representation of these bytes or the base64 encoded version of these bytes.

For a comparison see my example snippet below:
image

src/users/users.service.ts
```diff
@@ -26,8 +33,30 @@ export class UsersService {
...
+        accessToken = await this.hashPassword(randomString);
+        await this.getUserByAuthToken(accessToken);
+      } catch (NotInDBError) {
+        const token = AuthToken.create(user, identifier, accessToken);
+        return this.authTokenRepository.save(token);
+      }
+    }
+    // This should never happen
+    throw new RandomnessError(
+      'You machine is not able to generate not-in-use tokens. This should never happen.',
```
ErikMichelson, line 54:
```suggestion
      'Your machine is not able to generate not-in-use tokens. This should never happen.',
```

src/users/users.service.ts
```diff
@@ -26,8 +33,30 @@ export class UsersService {
     return this.userRepository.save(user);
   }
 
+  async createTokenForUser(
+    userName: string,
+    identifier: string,
+  ): Promise<AuthToken> {
+    const user = await this.getUserByUsername(userName);
+    let accessToken = '';
+    for (let i = 0; i < 100; i++) {
```
ErikMichelson, line 42: Do you want to hard-code these 100 tries to find a token?

src/users/users.service.ts
```diff
@@ -53,6 +98,44 @@ export class UsersService {
...
 
+  async getTokensByUsername(userName: string): Promise<AuthToken[]> {
+    const user = await this.getUserByUsername(userName);
+    return user.authTokens;
+  }
+
+  async removeToken(userName: string, timestamp: number) {
+    const user = await this.getUserByUsername(userName);
+    const token = await this.authTokenRepository.findOne({
+      where: { createdAt: new Date(timestamp), user: user },
```
ErikMichelson, line 109: Could it be possible that a user generates two tokens with the same timestamp?

Maybe the identifier should be unique and the token will be deleted based on its identifier.

ErikMichelson (dismissed):

src/users/users.service.ts
DerMolly: Well to be frank I based that on your implementation in the frontend repo. I thought you had good reason to build it that way.

ErikMichelson: Okay, I built it without having any knowledge about the backend. Timestamps are okay IMO, if the backend ensures that they're unique.

docs/content/dev/db-schema.plantuml
DerMolly: done

src/users/users.service.ts
DerMolly: done

src/users/users.service.ts
DerMolly: done

src/users/users.service.ts
DerMolly: I thought what would be a good upper limit of tries. Maybe 10 would be better? I don't think this should configurable or anything.

src/users/users.service.ts
DerMolly: Maybe it would be best to just expose the id and use that for the deletion? What do you think @davidmehren?

src/users/users.service.ts
DerMolly: done

src/users/users.service.ts
ErikMichelson: Well, we've got a chance of 1/2^512 collisions. I think we can safely remove this loop.

InnayTool (approved):

src/users/users.service.ts
SISheogorath: Just in case this isn't done yet (didn't check the new code) please use base64url, not regular base64 as those /-characters can be beasts.

src/users/users.service.ts
```diff
@@ -44,6 +65,24 @@ export class UsersService {
     return user;
   }
 
+  async hashPassword(password: string): Promise<string> {
+    // hash the password with SHA-256
+    const hash = createHash('sha256');
+    hash.update(password);
+    return hash.digest('hex');
+  }
```
SISheogorath, line 73: Please don't hash passwords just with sha-256.

Please let a proper library handle password hashing including salting, stretching and time resistant hashing functions such as argon2.

src/users/auth-token.entity.ts
```diff
@@ -6,20 +6,40 @@
...
+  @CreateDateColumn()
+  createdAt: Date;
+
+  @Column({ unique: true })
   accessToken: string;
+
+  public static create(
+    user: User,
+    identifier: string,
+    accessToken: string,
```
SISheogorath, line 36: Good token implementations also provide a last-used field allowing better management of tokens. Another addition could be a TTL or expires-at field to ensure tokens don't survive by accident.

SISheogorath (changes requested):
A minor and a major remark that we should get sorted.

src/users/users.service.ts
InnayTool: It's no password in this way and salting is not needed for fully random byte strings because there can't be an effective rainbow table. The problem of using a hash function with salt is, that we don't know the user before we matched the key, so checking all keys would be needed.

SISheogorath: Then it shouldn't be called "hashPassword"? Also why not add a token identifier or use a JWT if you want to use the token itself as selector?

Maybe one needs to explain the concept of this to me, but for now, I'm not really seeing the goal of this token implementation.

InnayTool: Yes renaming would make it clearer and it has to be commented. I think it's as secure as password auth, but maybe it's worth it to use JWTs.

DerMolly: Initially the function was meant to be used for both passwords and tokens, but that changed in the mean time. A renaming will occur.

Regarding JWTs, maybe that would be an idea…

src/users/auth-token.entity.ts
DerMolly: I guess this would be rather easy to implement with JWTs.

davidmehren:
I feel using a JWT is a bit overcomplicated here.
Why not:
- generate some random bytes
- create a string using base64url -> the `secret`
- hash the string using bcrypt2 and save to the database
- hand the token to the user as `base64url(username).secret`

When we get a token from the user, we can then get the username and only check the database for tokens that correspond to the user.

The only downside of this approach is that check-time scales linearly with the amount of API tokens a user has, but we could just limit that to a reasonable number (20?).

ErikMichelson:
Inspired by @davidmehren's idea. What about having a key-secret pair?

I could imagine the following process:
 - generate two random strings using base64url
 - one of the strings will be inserted into the database directly as keyid, the other one will be hashed and inserted into the database as corresponding secret
 - the token consist of `KEYID.SECRET`

With that approach you won't have a linear scale on the process to find the user for a token.

DerMolly:
If we use the id of the token anyway (for deletion purposes), we could just use that instead of a new second random string.

SISheogorath:
Let's not invent an own session management. The OWASP has some quite clear guidelines: https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html

We should also see if there might be a library which just takes care of that, as it appears to be a quite common use case.

DerMolly:
> Let's not invent an own session management. The OWASP has some quite clear guidelines: [cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html](https://cheatsheetseries.owasp.org/cheatsheets/Session_Management_Cheat_Sheet.html)

Your linking to something talking about sessions, when we don't implement any sessions (in this PR). I'm a bit confused what we should take from that post?

DerMolly:
> If we use the id of the token anyway (for deletion purposes), we could just use that instead of a new second random string.

A point against this made by @InnayTool is that an incrementing id would tell every user exactly how many tokens were at one point issued on the instance. An information we may want to not disclose.

ErikMichelson:
To make it clear:
I had the following data-structure in my mind:

- id (autogenerated, may be counted up, is always non-public, needs to be unique)
- user-id (relation to users table)
- keyid (A, needs to be unique)
- secret (hash of B)

The token is made from `A.B`.
A and B are both each a base64url version of random bytes.

An API user authenticates via the `A.B` token. The database lookup for the user  of the token is O(1) as the keyid can be matched directly to the user.

This doesn't involve any session logic. It's just about the tokens for the API. A user has a list of their API keyids to delete them any time. After generation a user can not reclaim the secret as it's only stored hashed. 

In some terms this would be similar to GitHub's PATs.

InnayTool:
A decision to make is if a token built like `base64url(keyid) + "." + base64url(argon2(secret))` or `base64url(byteconcat(keyid, argon2(secret)))`

DerMolly:
I prefer the first one, since it's easier to split `A` and `B` and it's common practice to concat such things with `.`.

InnayTool:
An advantage of concatenating with a `.`is that a API key can recognised by the `keyid`.

SISheogorath:
> Your linking to something talking about sessions, when we don't implement any sessions (in this PR). I'm a bit confused what we should take from that post?

A token is, at the end of the day, the identifying bit for a session. Token management and Session management are, to a certain degree, identical as the goal of both is to allow a client to again certain privileges on a server and ensuring that this client is actually what it claims to be. That's one of the reasons the document talks about sessions and tokens almost all the time.

But honestly speaking, after re-reading the entire thread and also the code, I wonder what's the actual purpose of the PR? Like what are the tokens actually used for? Should they allow a user to run `curl` queries against the API and therefore be more long term? Or should they be used for the frontend, means they should be more short lived?

InnayTool:
You can use it for curl but the intention is that you can write a bot, script or other program which manipulates, creates or reads notes and you don't have to store your username and password on disk if you want to use it non-interactively, so I think more long term.

The frontend should be authenticated by username and password and than the cookie for the session.

SISheogorath:
I mean, this token is also just a username + password in disguise, but I get the idea.

But then we are literally talking about the main purpose of JWTs. Just that JWTs are thought out and don't require a lot of custom logic and build own cryptography, which we've already seen in the thread.

JWTs will also have other benefits and have good libraries in JS so no one has to think themselves, which is always good for this kind of problems.

ErikMichelson:
JWTs are designed for another purpose. They are useful when a client needs to check the integrity of a token payload without having other knowledge about the database.

For our purpose — authenticating users against the server — we don't really need a payload inside the token if the token (or part of it) is identifier enough, therefore it also doesn't need to be cryptographically signed.

Even the JS libraries that exist for JWTs probably won't help us implementing this. We would need to store token data in the database anyway to allow some kind of revocation for users. But if we store the tokens in plain-text in the database, a database leakage (sometimes people don't protect their backups enough for example) would allow others to use the JWTs for write-access.

All in all, the JWT-approach looks for me like a lot of overhead without resulting in major benefits.

DerMolly:
> Even the JS libraries that exist for JWTs probably won't help us implementing this.

I played around with the JWT Libs that are tightly integrated with nestjs (passport-jwt to precise) and as far as I can tell they won't help us implementing this.

SISheogorath:
> Even the JS libraries that exist for JWTs probably won't help us implementing this. We would need to store token data in the database anyway to allow some kind of revocation for users. But if we store the tokens in plain-text in the database, a database leakage (sometimes people don't protect their backups enough for example) would allow others to use the JWTs for write-access.

We don't need to store any JWT in the database. We would store a (plaintext) token id in the database, which is part of the JWT and therefore a database leakage wouldn't affect anything as long as the secret from the application configuration, that is used to sign the JWT isn't leaked as well.

src/api/private/tokens/tokens.controller.ts
```diff
@@ -0,0 +1,58 @@
...
+    @Body('label') label: string,
+    @Body('until') until: number,
+  ): Promise<AuthTokenWithSecretDto> {
+    // ToDo: Get real userName
+    const authToken = await this.authService.createTokenForUser(
+      'hardcoded',
+      label,
+      until,
+    );
+    return this.authService.toAuthTokenWithSecretDto(authToken);
```
InnayTool, line 49: can be removed if `createTokenForUser` returns `createTokenForUser`.

src/errors/errors.ts
```diff
@@ -15,3 +15,7 @@ export class ClientError extends Error {
 export class PermissionError extends Error {
   name = 'PermissionError';
 }
+
+export class TokenNotValid extends Error {
```
InnayTool, line 19:
```suggestion
export class TokenNotValidError extends Error {
```

InnayTool (approved):

InnayTool (changes requested):
the things specified above.

src/auth/auth-token.dto.ts
```diff
@@ -0,0 +1,18 @@
...
+export class AuthTokenDto {
+  @IsString()
+  label: string;
+  @IsNumber()
+  created: number;
+  @IsNumber()
+  validUntil: number | null;
+  @IsNumber()
+  lastUsed: number | null;
+}
```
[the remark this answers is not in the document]
DerMolly: added

src/auth/auth-token.entity.ts
```diff
@@ -0,0 +1,64 @@
...
+  user: User;
+
+  @Column()
+  identifier: string;
+
+  @CreateDateColumn()
+  createdAt: Date;
+
+  @Column({ unique: true })
+  accessToken: string;
```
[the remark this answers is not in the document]
DerMolly: changed

src/auth/auth.service.ts
```diff
@@ -0,0 +1,156 @@
...
+import { User } from '../users/user.entity';
+import { AuthToken } from './auth-token.entity';
+import { AuthTokenDto } from './auth-token.dto';
+import { AuthTokenWithSecretDto } from './auth-token-with-secret.dto';
+/*
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
```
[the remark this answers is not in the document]
DerMolly: fixed

src/auth/auth.service.ts
```diff
@@ -0,0 +1,156 @@
...
+      .replace('+', '-')
+      .replace('/', '_')
+      .replace(/=+$/, '');
+  }
+
+  async createTokenForUser(
+    userName: string,
+    identifier: string,
+    until: number,
+  ): Promise<AuthToken> {
```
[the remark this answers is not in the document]
DerMolly: I'd like to keep the DTO's and internal Classes separated. So I think while this would technically be possible, I'd rather not make this change.

src/auth/auth.service.ts
```diff
@@ -0,0 +1,156 @@
...
+
+  async setLastUsedToken(keyId: string) {
+    const accessToken = await this.authTokenRepository.findOne({
+      where: { keyId: keyId },
+    });
+    accessToken.lastUsed = new Date().getTime();
+    await this.authTokenRepository.save(accessToken);
+  }
+
+  async getUserByAuthToken(keyId: string, token: string): Promise<User> {
```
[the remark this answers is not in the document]
DerMolly: Fixed

src/auth/auth.service.ts
```diff
@@ -0,0 +1,156 @@
...
+    }
+    return {
+      label: authToken.identifier,
+      created: authToken.createdAt.getTime(),
+      validUntil: authToken.validUntil,
+      lastUsed: authToken.lastUsed,
+    };
+  }
+
+  toAuthTokenWithSecretDto(
```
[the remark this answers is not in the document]
DerMolly: As stated above I don't think this is a good idea to implement.

src/auth/auth.service.ts
```diff
@@ -0,0 +1,156 @@
...
+    let token;
+    if (until === 0) {
+      token = AuthToken.create(user, identifier, keyId, accessToken);
+    } else {
+      token = AuthToken.create(user, identifier, keyId, accessToken, until);
+    }
+    const createdToken = await this.authTokenRepository.save(token);
+    return {
+      ...createdToken,
+      accessToken: `${keyId}.${secret}`,
```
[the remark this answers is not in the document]
DerMolly: Not to sure about this one. One the one hand I understand that the reassignment of accessToken is a bit hacky here, on the other hand introducing a whole new class just for this function seems unnecessary.

src/errors/errors.ts
DerMolly: fixed

src/auth/auth.service.ts
```diff
@@ -0,0 +1,156 @@
...
+  }
+
+  async createTokenForUser(
+    userName: string,
+    identifier: string,
+    until: number,
+  ): Promise<AuthToken> {
+    const user = await this.usersService.getUserByUsername(userName);
+    const secret = this.randomBase64UrlString();
+    const keyId = this.randomBase64UrlString();
```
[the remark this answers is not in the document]
DerMolly: any recommendations? I just sticked with 64 as we already implement a function to create random strings of this length anyway.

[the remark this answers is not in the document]
InnayTool: I think 8 byte will be sufficient.

src/auth/auth.service.ts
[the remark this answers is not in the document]
InnayTool: It's no new class it's only the AuthTokenWithSecretDto class.

src/auth/auth.service.ts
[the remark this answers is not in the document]
InnayTool: I think it's really confusing, that you save the value in the wrong field and afterwards move it to another. I see no benefit in having an internal and external class. The class is only used for returning the access token. There is no internal use beside returning it from a function. And the important thing is that this token should be passed around in the program. Nobody should use it and only return it to the user.

src/auth/auth.service.ts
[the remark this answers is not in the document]
InnayTool: By the way I think this is much to hacky for a security critical section of the program.

src/auth/auth.service.ts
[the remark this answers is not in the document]
DerMolly: fixed

src/auth/auth.service.ts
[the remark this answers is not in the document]
DerMolly: fixed

src/auth/auth.service.ts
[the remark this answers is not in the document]
DerMolly: fixed

src/auth/auth.service.ts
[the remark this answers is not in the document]
DerMolly: fixed

src/api/private/tokens/tokens.controller.ts
DerMolly: fixed

InnayTool (dismissed):

src/auth/auth-token.entity.ts
```diff
@@ -0,0 +1,64 @@
...
+  })
+  lastUsed: number;
+
+  public static create(
+    user: User,
+    identifier: string,
+    keyId: string,
+    accessToken: string,
+    validUntil?: number,
+  ): Pick<AuthToken, 'user' | 'accessTokenHash'> {
```
davidmehren, line 52: That type looks wrong, it would only contain the `user` and `accessTokenHash` keys, but you set more properties.

src/auth/auth.service.spec.ts
```diff
@@ -0,0 +1,157 @@
...
+      .compile();
+
+    service = module.get<AuthService>(AuthService);
+  });
+
+  it('should be defined', () => {
+    expect(service).toBeDefined();
+  });
+
+  it('checkPassword', async () => {
```
davidmehren, line 98: The names of unit tests should generally form a (somewhat) complete sentence. In [the Nest example](https://docs.nestjs.com/fundamentals/testing#unit-testing) the `describe` and `it` calls result in `CatsController find all should return an array of cats`.
Could you change these names into something similar?

src/auth/auth.service.ts
```diff
@@ -0,0 +1,158 @@
...
+    private readonly logger: ConsoleLoggerService,
+    private usersService: UsersService,
+    @InjectRepository(AuthToken)
+    private authTokenRepository: Repository<AuthToken>,
+  ) {
+    this.logger.setContext(AuthService.name);
+  }
+
+  async validateToken(token: string): Promise<User> {
+    const parts = token.split('.');
```
davidmehren, line 32: What about `[keyId, token] = token.split('.')`?
Is a bit clearer than a `parts` array.

src/auth/auth.service.ts
```diff
@@ -0,0 +1,158 @@
...
+    if (user) {
+      await this.setLastUsedToken(parts[0]);
+      return user;
+    }
+    return null;
+  }
+
+  async hashPassword(cleartext: string): Promise<string> {
+    // hash the password with bcrypt and 2^16 iterations
+    return hash(cleartext, 12);
```
davidmehren, line 46: Any reason why you chose 12?

src/auth/auth.service.ts
```diff
@@ -0,0 +1,158 @@
...
+      .toString('base64')
+      .replace('+', '-')
+      .replace('/', '_')
+      .replace(/=+$/, '');
+  }
+
+  async createTokenForUser(
+    userName: string,
+    identifier: string,
+    until: number,
```
davidmehren, line 75: Same thing about renaming `until` and the type

src/auth/auth.service.ts
```diff
@@ -0,0 +1,158 @@
...
+
+  async setLastUsedToken(keyId: string) {
+    const accessToken = await this.authTokenRepository.findOne({
+      where: { keyId: keyId },
+    });
+    accessToken.lastUsed = new Date().getTime();
+    await this.authTokenRepository.save(accessToken);
+  }
+
+  async getAuthToken(keyId: string, token: string): Promise<AuthToken> {
```
davidmehren, line 100: This not only gets the token, but also validates it. It might make sense to either rename this method or split it into two.

src/auth/auth.service.ts
```diff
@@ -0,0 +1,166 @@
...
+      validUntil: authToken.validUntil,
+      lastUsed: authToken.lastUsed,
+    };
+  }
+
+  toAuthTokenWithSecretDto(
+    authToken: AuthToken | null | undefined,
+    secret: string,
+  ): AuthTokenWithSecretDto | null {
+    const tokeDto = this.toAuthTokenDto(authToken);
```
davidmehren, line 160: typo?

src/auth/auth.service.ts
```diff
@@ -0,0 +1,166 @@
...
+    const user = await this.usersService.getUserByUsername(userName);
+    const token = await this.authTokenRepository.findOne({
+      where: { keyId: keyId, user: user },
+    });
+    await this.authTokenRepository.remove(token);
+  }
+
+  toAuthTokenDto(authToken: AuthToken | null | undefined): AuthTokenDto | null {
+    if (!authToken) {
+      this.logger.warn(`Recieved ${authToken} argument!`, 'toAuthTokenDto');
```
davidmehren, line 144: Why do we even allow a nullable parameter here?

davidmehren (changes requested):
- Please write documentation, at least for the services
- Currently, both the private and the public API are shown in the same Swagger UI, we should change that. If you don't want to solve that now, please open an issue.
- Did we already discuss how to handle instances that allow anonymous interaction?

src/auth/token-auth.guard.ts
```diff
@@ -0,0 +1,11 @@
...
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+import { Injectable } from '@nestjs/common';
+import { AuthGuard } from '@nestjs/passport';
+
+@Injectable()
+export class TokenAuthGuard extends AuthGuard('token') {}
```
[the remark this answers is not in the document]
DerMolly: Correct

src/auth/auth.service.ts
DerMolly: fixed

src/auth/auth.service.ts
DerMolly: Not to sure. This was code I copied over from before. It should originate from either you or @InnayTool…

src/auth/auth.service.ts
DerMolly: fixed

src/auth/auth.service.ts
DerMolly: It's the default in many bcrypt libs. Also I read some discussion on the amount of rounds and it seemed long enough.

src/auth/auth.service.ts
DerMolly: fixed

src/auth/auth-token.entity.ts
DerMolly: fixed

src/auth/auth-token.dto.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export class AuthTokenDto {
+  @IsString()
+  label: string;
+  @IsString()
+  keyId: string;
+  @IsNumber()
+  created: number;
+  @IsNumber()
+  validUntil: number | null;
```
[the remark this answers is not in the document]
DerMolly: fixed

src/auth/auth-token.dto.ts
```diff
@@ -0,0 +1,20 @@
...
+
+import { IsNumber, IsString } from 'class-validator';
+
+export class AuthTokenDto {
+  @IsString()
+  label: string;
+  @IsString()
+  keyId: string;
+  @IsNumber()
+  created: number;
```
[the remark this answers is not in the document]
DerMolly: fixed

src/api/public/media/media.controller.ts
```diff
@@ -35,14 +40,16 @@ export class MediaController {
...
 
+  @UseGuards(TokenAuthGuard)
   @Post()
   @UseInterceptors(FileInterceptor('file'))
   async uploadMedia(
+    @Request() req,
     @UploadedFile() file: MulterFile,
     @Headers('HedgeDoc-Note') noteId: string,
   ) {
     //TODO: Get user from request
```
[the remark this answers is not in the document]
DerMolly: fixed

src/auth/auth.service.ts
DerMolly: fixed

src/auth/auth.service.spec.ts
DerMolly: fixed

src/api/private/tokens/tokens.controller.ts
```diff
@@ -0,0 +1,53 @@
...
+    // ToDo: Get real userName
+    return (
+      await this.authService.getTokensByUsername('hardcoded')
+    ).map((token) => this.authService.toAuthTokenDto(token));
+  }
+
+  @Post()
+  async postTokenRequest(
+    @Body('label') label: string,
+    @Body('until') until: number,
```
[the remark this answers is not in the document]
DerMolly: fixed

src/auth/auth.service.ts
davidmehren: Could you put that into a comment so the reference does not get lost?

davidmehren (changes requested):

src/auth/auth.service.ts
DerMolly: sure

DerMolly:
> Currently, both the private and the public API are shown in the same Swagger UI, we should change that. If you don't want to solve that now, please open an issue.

I created #759

DerMolly:
> Did we already discuss how to handle instances that allow anonymous interaction?

At least not in this PR

DerMolly:
> Please write documentation, at least for the services

Could you point me to what I should write and where to put it exactly?

src/auth/auth.service.ts
```diff
@@ -0,0 +1,186 @@
...
+
+  async getTokensByUsername(userName: string): Promise<AuthToken[]> {
+    const user = await this.usersService.getUserByUsername(userName, true);
+    if (user.authTokens === undefined) {
+      return [];
+    }
+    return user.authTokens;
+  }
+
+  async removeToken(userName: string, keyId: string) {
```
InnayTool, line 144: I think `userName` is not needed here because the `keyId` is unique.

InnayTool (approved):

src/auth/auth.service.ts
```diff
@@ -0,0 +1,185 @@
...
+  }
+
+  async validateToken(token: string): Promise<User> {
+    const [keyId, secret] = token.split('.');
+    const accessToken = await this.getAuthTokenAndValidate(keyId, secret);
+    const user = await this.usersService.getUserByUsername(
+      accessToken.user.userName,
+    );
+    if (user) {
+      await this.setLastUsedToken(keyId);
```
SISheogorath, line 39: Why do we only run this in case the previous query resulted in a user? 

Also I don't think we need to await this call since even if there is a race condition, the difference in time is basically meaningless.

src/auth/auth.service.ts
```diff
@@ -0,0 +1,185 @@
...
+
+  toAuthTokenDto(authToken: AuthToken): AuthTokenDto | null {
+    if (!authToken) {
+      this.logger.warn(`Recieved ${authToken} argument!`, 'toAuthTokenDto');
+      return null;
+    }
+    const tokenDto: AuthTokenDto = {
+      lastUsed: null,
+      validUntil: null,
+      label: authToken.identifier,
```
SISheogorath, line 159: Are we sure that we want to have different names for this? Wouldn't make sense to always call it either label or identifier?

SISheogorath (commented):
Some minor remarks. Additionally I think we should setup an interval that cleans up expired tokens, as they are useless anyway and don't need to clunk the database. Also I would recommend to add some arbitrary, but intentional limits on Token Lifetime and number, those will help to prevent potential abuse by rogue clients/users as well as people playing the "guess the secret"-game in the long run.

src/auth/auth.service.ts
ErikMichelson: I'd prefer the usage of label throughout the code to not mix it up with the key identifier.

DerMolly: fixed

DerMolly:
> Some minor remarks. Additionally I think we should setup an interval that cleans up expired tokens, as they are useless anyway and don't need to clunk the database.

Agreed and implemented.

> Also I would recommend to add some arbitrary, but intentional limits on Token Lifetime and number, those will help to prevent potential abuse by rogue clients/users as well as people playing the "guess the secret"-game in the long run.

Do you mean any token should only life a maximum of x days / years / whatever or do you mean if tokens have an expiration date it can only be x days / months / years in the future and not further in the future?

What would you suggest the maximum number of tokens to be? 10? 20?

src/auth/auth.service.ts
DerMolly: This was necessary before, but now isn't necessary anymore. I moved the call above the user search.
It should be fixed now.

SISheogorath:
> Do you mean any token should only live a maximum of x days / years / whatever or do you mean if tokens have an expiration date it can only be x days / months / years in the future and not further in the future?

Yes, as with certificates you need to rotate secrets. The sooner the better. For the simplicity sake I would go for 2-5 years as maximum age of a token, this should make a reasonable time with API rate-limits and alike.

> What would you suggest the maximum number of tokens to be? 10? 20?

I would actually go for 200. This is a rather high number, means only very, very few valid use-cases should ever hit it, while abuse is limited rather well. It'll definitely prevent someone from filling up the table and making queries unnecessary slow.

DerMolly:
> Yes, as with certificates you need to rotate secrets. The sooner the better. For the simplicity sake I would go for 2-5 years as maximum age of a token, this should make a reasonable time with API rate-limits and alike.

done
 
> I would actually go for 200. This is a rather high number, means only very, very few valid use-cases should ever hit it, while abuse is limited rather well. It'll definitely prevent someone from filling up the table and making queries unnecessary slow.

done

InnayTool (approved):

SISheogorath (approved):
All in call, LGTM

src/auth/auth.service.ts
```diff
@@ -0,0 +1,216 @@
...
+    const tokenDto = this.toAuthTokenDto(authToken);
+    return {
+      ...tokenDto,
+      secret: secret,
+    };
+  }
+
+  // Delete all non valid tokens  every sunday on 3:00 AM
+  @Cron('0 0 3 * * 0')
+  async handleCron() {
```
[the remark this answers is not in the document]
DerMolly: Any reason why after x minutes on not after x seconds (which I added with 5 seconds just now)? The startup process is so fast I think we can do after 5s.

src/auth/auth.service.ts
```diff
@@ -0,0 +1,216 @@
...
+    // This is necessary as the is no base64url encoding in the toString method
+    // but as can be seen on https://tools.ietf.org/html/rfc4648#page-7
+    // base64url is quite easy buildable from base64
+    if (length <= 0) {
+      return null;
+    }
+    return randomBytes(length);
+  }
+
+  BufferToBase64Url(text: Buffer): string {
```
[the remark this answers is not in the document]
DerMolly: There was one, but I forgot to move it from the function two lines above, when I split them up.

src/auth/auth.service.ts
```diff
@@ -0,0 +1,232 @@
...
+    userName: string,
+    identifier: string,
+    validUntil: TimestampMillis,
+  ): Promise<AuthTokenWithSecretDto> {
+    const user = await this.usersService.getUserByUsername(userName, true);
+    if (user.authTokens.length >= 200) {
+      // This is a very high ceiling unlikely to hinder legitimate usage,
+      // but should prevent possible attack vectors
+      throw new TooManyTokensError(
+        `User '${user.displayName}' has already 200 tokens and can't have anymore`,
```
SISheogorath, line 88: This should be username as the display name doesn't require uniqueness and therefore finding this user is more complicated.

src/auth/auth.service.ts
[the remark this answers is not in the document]
SISheogorath: The idea was to declutter the start process (for example a lot of stuff is going on during a freshly booted system. But 5 seconds should also work completely fine.)

SISheogorath (dismissed):

src/auth/auth.service.ts
DerMolly: good point

DerMolly: fixed

SISheogorath (dismissed):

InnayTool (dismissed):

davidmehren (approved):

InnayTool (approved):

Issue #754, opened 2021-01-22
Unintended note creation in FreeURL mode


nidico:
If FreeURL mode is enabled, a single `GET` request on a non-existing note immediately creates that note.

This may lead to many accidentally created notes. In addition, there are many bots which try out known paths (`/wp-admin` etc) in order to automate attack attempts or to just map the web. We just discovered quite a few notes which have been created by such bots on a relatively unknown HedgeDoc instance.

Possible solutions / mitigations:

- Create a modal dialog when opening a formally unknown note and require a POST on a given form (e.g. a button labeled "Yes, please create the note/pad"). This is what etherpad "classic" (not etherpad-lite) is doing. This would mitigate accidental creations as well as stupid bots which just randomly try out URLs and do not actively try to "DDOS" HedgeDoc.

- Allow to restrict FreeURL mode to authenticated users. This would avoid the "bot spam". Of course this is a limitation for unauthenticated users, but it's still better than disabling FreeURL mode entirely.

As the latter is easy to implement and backward compatible and at least solves the problem for us, I'm going to submit a PR for that option.

I'd propose to allow that simple addition in HedgeDoc 1.7.x (as it solves real problems *now*) and to do something like a modal dialog in HedgeDoc 2.x.

davidmehren:
> If FreeURL mode is enabled, a single GET request on a non-existing note immediately creates that note.

I'd call that "works as intended" :D
So yes, this behavior creates more rows in the database, but that doesn't look like a big problem to me.

Regarding your PR:
A new config option is a new feature, which calls for a new minor release, which we do not plan to do for 1.x.

nidico:
> I'd call that "works as intended" :D
> So yes, this behavior creates more rows in the database, but that doesn't look like a big problem to me.

I understand there may be diverging views on the criticality of the issue. We considered the issue *that bad* that we decided to disable FreeURL although we really much like it. Hence the PR, which we'll most likely deploy, even if the PR wouldn't be merged upstream (although this would create additional work)...

>  A new config option is a new feature, which calls for a new minor release, which we do not plan to do for 1.x.

I had hoped this could slip through in a 1.7.x release as it's non-breaking and partially fixes a bug (if you call it that way).

davidmehren:
> We considered the issue that bad that we decided to disable FreeURL although we really much like it.

Could you elaborate on why you think this is so bad?

[...]

PR #755, opened 2021-01-22
Add config option which requires authentication in FreeURL mode

nidico:
### Component/Part
Configuration / backend.

### Description
This PR mitigates the issue described in #754 by introducing a config option which - if enabled - only allows authenticated users to create new FreeURL notes.

This is implemented in a backward compatible way, i.e. if the default option is applied, nothing changes.

### Steps

- [x] Added implementation
- [ ] Added / updated tests

### Related Issue(s)

#754 -  Unintended note creation in FreeURL mode


lib/config/environment.js
```diff
@@ -29,6 +29,7 @@ module.exports = {
   allowAnonymous: toBooleanConfig(process.env.CMD_ALLOW_ANONYMOUS),
   allowAnonymousEdits: toBooleanConfig(process.env.CMD_ALLOW_ANONYMOUS_EDITS),
   allowFreeURL: toBooleanConfig(process.env.CMD_ALLOW_FREEURL),
+  requireFreeURLAuthentication: toBooleanConfig(process.env.CMD_REQUIRE_FREEURL_AUTH),
```
SISheogorath, line 32: Either the environment variable should be written out or the JS variable name should be cut. Let's try to keep the variable names in sync.

lib/config/hackmdEnvironment.js
```diff
@@ -24,6 +24,7 @@ module.exports = {
   allowAnonymous: toBooleanConfig(process.env.HMD_ALLOW_ANONYMOUS),
   allowAnonymousEdits: toBooleanConfig(process.env.HMD_ALLOW_ANONYMOUS_EDITS),
   allowFreeURL: toBooleanConfig(process.env.HMD_ALLOW_FREEURL),
+  requireFreeURLAuthentication: toBooleanConfig(process.env.HMD_REQUIRE_FREEURL_AUTH),
```
SISheogorath, line 27: This is not not needed. We don't add new config to a compatibly layer.

lib/config/environment.js
nidico: Okay, I have decided for the long version as *auth* could also mean *authorization*.

lib/config/hackmdEnvironment.js
nidico: Okay, removed. I honestly didn't even know what this file was good for, but I have assumed it's there for legacy reasons...

nidico:
(push-forced because of missing sign-off)

SISheogorath (approved):
No further concerns from my side

davidmehren (approved):

davidmehren:
Thanks!

I hope to release 1.8 with this change in the next few days.

nidico:
Thanks for reviewing and merging!

PR #768, opened 2021-01-25
Swagger: Split public and private API

### LINKED ISSUE #759: Split openApi docs
the openapi docs

Split public API and private API in the autogenerated openapi docs. Maybe we don't want to show docs for the private API on every instance like we do for the public API.

davidmehren:
### Component/Part
Swagger config

### Description
This PR splits the Swagger docs into the public and private API.

The docs for the public API continue to be available under `/apidocs`, the docs for the private API are now separated under `/private/apidocs`.

### Steps
- [x] Added implementation

### Related Issue(s)
Fixes #759

### DIFF
--- a/src/main.ts
+++ b/src/main.ts
@@ -8,11 +8,11 @@ import { ValidationPipe } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
 import { NestFactory } from '@nestjs/core';
 import { NestExpressApplication } from '@nestjs/platform-express';
-import { DocumentBuilder, SwaggerModule } from '@nestjs/swagger';
 import { AppModule } from './app.module';
 import { AppConfig } from './config/app.config';
-import { NestConsoleLoggerService } from './logger/nest-console-logger.service';
 import { MediaConfig } from './config/media.config';
+import { NestConsoleLoggerService } from './logger/nest-console-logger.service';
+import { setupPrivateApiDocs, setupPublicApiDocs } from './utils/swagger';
 
 async function bootstrap() {
   const app = await NestFactory.create<NestExpressApplication>(AppModule);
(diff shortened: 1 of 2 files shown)


InnayTool (dismissed):

src/main.ts
```diff
@@ -23,16 +25,28 @@ async function bootstrap() {
   const appConfig = configService.get<AppConfig>('appConfig');
   const mediaConfig = configService.get<MediaConfig>('mediaConfig');
 
-  const swaggerOptions = new DocumentBuilder()
-    .setTitle('HedgeDoc')
+  const publicApiOptions = new DocumentBuilder()
+    .setTitle('HedgeDoc Public API')
     .setVersion('2.0-dev')
```
DerMolly, line 30: Shouldn't this be extracted from the package.json or something (less chance of a missed change?)

src/main.ts
```diff
@@ -23,16 +25,28 @@ async function bootstrap() {
...
+
+  const privateApiOptions = new DocumentBuilder()
+    .setTitle('HedgeDoc Private API')
+    .setVersion('2.0-dev')
+    .build();
+
+  const privateApi = SwaggerModule.createDocument(app, privateApiOptions, {
+    include: [PrivateApiModule],
+  });
+  SwaggerModule.setup('private/apidoc', app, privateApi);
```
DerMolly, line 49: Do we want to publish the private API in this form? Don't we encourage people to use the private changing api and buy us more support tickets in the future?

DerMolly (changes requested):

src/main.ts
davidmehren: Yeah, but we currently don't have a global version or similar.
So I'd just keep it that way for now

src/main.ts
davidmehren: What about only publishing the private APIdocs when we run with `NODE_ENV=development`?
Having quick access to the docs might be helpful when developing our frontend.

src/main.ts
DerMolly: Maybe add an Todo then?

davidmehren:
To not clutter `main.ts` too much, I extracted the setup code into a separate file and then added the development check.

DerMolly (approved):

InnayTool (approved):

PR #771, opened 2021-01-26
Fix base64url transformation

DerMolly:
### Component/Part
base64url encoder

### Description
This PR fixes the base64url encoder and encodes the secret as base64url

### Steps

- [x] Added implementation

### DIFF
--- a/src/auth/auth.service.ts
+++ b/src/auth/auth.service.ts
@@ -70,9 +70,9 @@ export class AuthService {
     // base64url is quite easy buildable from base64
     return text
       .toString('base64')
-      .replace('+', '-')
-      .replace('/', '_')
-      .replace(/=+$/, '');
+      .replaceAll(/\+/g, '-')
+      .replaceAll(/\//g, '_')
+      .replaceAll(/=+$/g, '');
   }
 
   async createTokenForUser(
--- a/src/auth/auth.service.ts
+++ b/src/auth/auth.service.ts
@@ -70,9 +70,9 @@ export class AuthService {
     // base64url is quite easy buildable from base64
     return text
       .toString('base64')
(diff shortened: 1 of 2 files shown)


src/auth/auth.service.ts
```diff
@@ -88,7 +88,7 @@ export class AuthService {
         `User '${user.userName}' has already 200 tokens and can't have anymore`,
       );
     }
-    const secret = await this.randomString(64);
+    const secret = this.BufferToBase64Url(await this.randomString(64));
     const keyId = this.BufferToBase64Url(await this.randomString(8));
     const accessTokenString = await this.hashPassword(secret.toString());
```
InnayTool, line 93:
```suggestion
    const accessToken = await this.hashPassword(secret.toString());
```

src/auth/auth.service.ts
```diff
@@ -88,7 +88,7 @@ export class AuthService {
...
     const accessToken = this.BufferToBase64Url(Buffer.from(accessTokenString));
```
InnayTool, line 94: By now the validate never matches. This can be fixed by removing this line, but the output in database is not base64 encoded anymore. An alternative is to decode the string before the comparison with the token which was send be the client. 
```suggestion
```

tsconfig.json
```diff
@@ -10,6 +10,7 @@
     "sourceMap": true,
     "outDir": "./dist",
     "baseUrl": "./",
-    "incremental": true
+    "incremental": true,
+    "lib": ["esnext"]
```
InnayTool, line 14: Do we use esnext?

InnayTool (changes requested):

InnayTool (approved):

davidmehren (approved):

PR #778, opened 2021-01-28
auth: Fix secret length

DerMolly:
### Component/Part
auth tokens

### Description
This PR fixes the length of the secrets.

@InnayTool found out that if you remove the last characters from the secret the token isn't invalid as one would suspect, but still works until quite some characters where removed. We looked into this and found out:

The former length of 64 bytes (512-bit) is transformed into base64url (a 6-bit code) ~86 characters long. This is too long for bcrypt as it ignores any characters beyond the 72th.

This fix therefore reduces the amount of generated bytes to 54 (as 72*6/8 = 54) characters. This ensures that removing one character from the token the hash won't be the same anymore.

### Steps

- [x] Added implementation
- [x] Added / updated tests


SISheogorath:
While I actually double checked our used secret length, I ignored the encoding. Alternatively we could decode the secret before hashing, just as an idea. Either way, we should maybe add a check for a maximum secret length and reject them. (Of course this will result in a non-semi-constant time for secret validation, but I guess this can be considered neglectable.

InnayTool (dismissed):

DerMolly:
> Either way, we should maybe add a check for a maximum secret length and reject them. (Of course this will result in a non-semi-constant time for secret validation, but I guess this can be considered neglectable).

I added this functionality to the code.

src/auth/auth.service.ts
```diff
@@ -36,6 +36,12 @@ export class AuthService {
 
   async validateToken(token: string): Promise<User> {
     const [keyId, secret] = token.split('.');
+    if (secret.length > 72) {
+      // Only the first 72 characters of the tokens are considered by bcrypt
+      // This should prevent strange corner cases
+      // At the very least it won't hurt us
+      throw new TokenNotValidError(`AuthToken '${secret}' is too long the be a proper tokeb`)
```
InnayTool, line 43:
```suggestion
      throw new TokenNotValidError(`AuthToken '${secret}' is too long the be a proper token`)
```

InnayTool (changes requested):

src/auth/auth.service.ts
DerMolly: fixed

InnayTool (dismissed):

InnayTool (approved):

InnayTool (dismissed):

davidmehren (approved):

PR #781, opened 2021-01-29
fix: the markdown layout of the configuration.md

oupala:
### Component/Part
documentation

### Description
This PR fixes the markdown layout of the configuration.md file which has been messed up at some point.

### Steps

### Related Issue(s)
n/a


oupala:
As far as I understand, the `configuration.md` file was ok when I last modified it ([see here](https://github.com/hedgedoc/hedgedoc/blob/260e9b8958cd48f79ba40f67d86d91d2f4d137a6/docs/configuration.md)). Then the documentation was moved to a subdirectory by @mrdrogdrog and then the markdown layout was messed up ([see here](https://github.com/hedgedoc/hedgedoc/blob/4b7318ca33e7f5588fba11bb3d25def3c6c2d41b/docs/content/configuration.md)).

I don't understand why and how this markdown layout became messed out, but at least I think that the *content* has not been modified since my commit (the only think that was modified is the layout that was messed up).

As a consequence, I restored the previous file with a correct layout. Then I replaced every `CodiMd` word by `HedgeDoc`.

mrdrogdrog:
Thank you for your effort here. The reason why this file is still so messed up is because the branch `develop` is not ready for productional use. It is for the new 2.0 version. (Maybe we should rename it to `develop-2.0` ?). The docs are going to be completely reworked.

So if have cleaned this file up as preparation for https://github.com/hedgedoc/container/pull/153, then please keep in mind, that the current 1.x development is done on the `master` branch. Also docs.hedgedoc.org is generated from `master`.

oupala:
Thanks you @mrdrogdrog for these useful informations. I wasn't aware that the `master` branch was for 1.7.x development, and that `develop` branch was for 2.x.x development. This could eventually made clearer so that contributors do not get cheated.

I think this pull request has no sense now that I know that the `configuration.md` file is ok on the `master` branch. I think we could close this pull request.

mrdrogdrog:
If you say so, then I'll close it.

DerMolly:
>  This could eventually made clearer so that contributors do not get cheated.

Maybe we should add this to the PR template, too? It's already in the README…

oupala:
> > This could eventually made clearer so that contributors do not get cheated.
> 
> Maybe we should add this to the PR template, too? It's already in the README…

Yes, sure this should be added in the PR template.

But this is not sufficient as the PR template is only seen by a contributor when he opens a PR, and this might be at the end of his dev session, where it is too late because he started to dev from the wrong base-branch. So in my opinion, this should be indicated somewhere in the documentation, in the README and in the name of the branch (branch `develop-2.0` as proposed by @mrdrogdrog).

PR #786, opened 2021-01-29
Refactoring of controllers and service interfaces

InnayTool:
DTO should only be used for sending information to and from user.
Services now have methods which return normal internal objects and
methods which convert them to DTOs. This conversion is done in the
controlers

### Description
This PR fixes/adds/improves/...

### Steps

- [x] Added implementation
- [x] Added / updated tests

### DIFF
--- a/src/auth/mock-auth.guard.ts
+++ b/src/auth/mock-auth.guard.ts
@@ -5,14 +5,20 @@
  */
 
 import { ExecutionContext, Injectable } from '@nestjs/common';
+import { UsersService } from '../users/users.service';
+import { User } from '../users/user.entity';
 
 @Injectable()
 export class MockAuthGuard {
-  canActivate(context: ExecutionContext) {
+  private user: User;
+  constructor(private usersService: UsersService) {}
+
+  async canActivate(context: ExecutionContext) {
     const req = context.switchToHttp().getRequest();
-    req.user = {
-      userName: 'hardcoded',
-    };
+    if (!this.user) {
+      this.user = await this.usersService.createUser('hardcoded', 'Testy');
+    }
+    req.user = this.user;
     return true;
   }
 }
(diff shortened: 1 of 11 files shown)


src/auth/mock-auth.guard.ts
```diff
@@ -5,14 +5,15 @@
...
 export class MockAuthGuard {
-  canActivate(context: ExecutionContext) {
+  constructor(private usersService: UsersService) {}
+
+  async canActivate(context: ExecutionContext) {
     const req = context.switchToHttp().getRequest();
-    req.user = {
-      userName: 'hardcoded',
-    };
+    req.user = await this.usersService.createUser('hardcoded', 'Testy');
```
davidmehren, line 16: This might lead to the user getting created multiple times.
It should be safer to move the `createUser` call to the constructor.

davidmehren (changes requested):

src/auth/mock-auth.guard.ts
DerMolly: good point. I'll change that

DerMolly: Only problem is: This is a async code that can't be called in the consturctor. The normal solution would be to add a async builderPattern, but as we don't instantiate the class our self this won't do as any good. Only solution I could think of would be a singleton, not the most elegant solution, but I guess it should do.

mrdrogdrog: You could check if the user already exists?

mrdrogdrog: Would a boolean field enough for this check?

DerMolly: If I understand the tests correctly, the db gets recreated after each test so this way should suffice.

InnayTool: Is this resolved?

DerMolly: For me yes. But I'd leave the closing of this conversation to @davidmehren

davidmehren (approved):

PR #936, opened 2021-02-20
Stronger ESLint rules

### LINKED ISSUE #838: Directly returning async service methods leads to broken stack traces
**Description**
If we have a method `async foo()` and in it `return` a method `async bar()` that `throw`s an exception, `foo()` does not appear on the stack trace. Only `foo`s caller and `bar()` appear.

A more detailed explanation can be found here: https://www.even.li/posts/2020-06-04-async-stack-traces/

This problem bit me while debugging a `TypeError` in a service method as I could not find out from which controller it was called. I just got `@nestjs/core/router/router-execution-context.js` in the trace and then the service.

The only solution I could find was to always use `return await` instead of just `return` as laid out in the blog post.
I propose to turn on the [associated eslint check](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/return-await.md).

davidmehren:
### Component/Part
ESLint

### Description
This PR improves code quality by enabling more ESLint rules.
It also fixes #838 by enabling the rule mentioned there.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
#838

### DIFF
--- a/src/api/public/notes/notes.controller.ts
+++ b/src/api/public/notes/notes.controller.ts
@@ -221,7 +221,7 @@ export class NotesController {
         throw new UnauthorizedException('Updating note denied!');
       }
       return await this.noteService.toNotePermissionsDto(
-        await this.noteService.updateNotePermissions(note, updateDto),
+        this.noteService.updateNotePermissions(note, updateDto),
       );
     } catch (e) {
       if (e instanceof NotInDBError) {
--- a/src/api/public/notes/notes.controller.ts
+++ b/src/api/public/notes/notes.controller.ts
@@ -15,7 +15,7 @@ import {
   Param,
   Post,
   Put,
-  Request,
+  Req,
   UnauthorizedException,
   UseGuards,
 } from '@nestjs/common';
@@ -59,7 +59,7 @@ export class NotesController {
   @UseGuards(TokenAuthGuard)
   @Post()
   async createNote(
-    @Request() req,
+    @Req() req,
     @MarkdownBody() text: string,
   ): Promise<NoteDto> {
     // ToDo: provide user for createNoteDto
@@ -75,7 +75,7 @@ export class NotesController {
   @UseGuards(TokenAuthGuard)
   @Get(':noteIdOrAlias')
   async getNote(
-    @Request() req,
+    @Req() req,
     @Param('noteIdOrAlias') noteIdOrAlias: string,
   ): Promise<NoteDto> {
     let note: Note;
@@ -97,7 +97,7 @@ export class NotesController {
   @UseGuards(TokenAuthGuard)
   @Post(':noteAlias')
   async createNamedNote(
-    @Request() req,
+    @Req() req,
     @Param('noteAlias') noteAlias: string,
     @MarkdownBody() text: string,
   ): Promise<NoteDto> {
@@ -120,7 +120,7 @@ export class NotesController {
   @UseGuards(TokenAuthGuard)
   @Delete(':noteIdOrAlias')
   async deleteNote(
-    @Request() req,
+    @Req() req,
     @Param('noteIdOrAlias') noteIdOrAlias: string,
   ): Promise<void> {
     try {
@@ -143,7 +143,7 @@ export class NotesController {
   @UseGuards(TokenAuthGuard)
   @Put(':noteIdOrAlias')
   async updateNote(
-    @Request() req,
+    @Req() req,
     @Param('noteIdOrAlias') noteIdOrAlias: string,
     @MarkdownBody() text: string,
   ): Promise<NoteDto> {
@@ -168,7 +168,7 @@ export class NotesController {
   @Get(':noteIdOrAlias/content')
   @Header('content-type', 'text/markdown')
   async getNoteContent(
-    @Request() req,
+    @Req() req,
     @Param('noteIdOrAlias') noteIdOrAlias: string,
   ): Promise<string> {
     try {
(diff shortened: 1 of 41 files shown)


davidmehren:
This PR is best reviewed commit-by-commit.

src/auth/auth.service.ts
```diff
@@ -173,7 +175,10 @@ export class AuthService {
 
   toAuthTokenDto(authToken: AuthToken): AuthTokenDto | null {
     if (!authToken) {
-      this.logger.warn(`Recieved ${authToken} argument!`, 'toAuthTokenDto');
+      this.logger.warn(
+        `Recieved ${String(authToken)} argument!`,
```
DerMolly, line 179: Shouldn't this rather be `authToken.toString()`?

src/groups/groups.service.ts
```diff
@@ -45,7 +45,7 @@ export class GroupsService {
    */
   toGroupDto(group: Group | null | undefined): GroupInfoDto | null {
     if (!group) {
-      this.logger.warn(`Recieved ${group} argument!`, 'toGroupDto');
+      this.logger.warn(`Recieved ${String(group)} argument!`, 'toGroupDto');
```
DerMolly, line 48: same as above

src/logger/console-logger.service.ts
```diff
@@ -21,7 +21,7 @@ export class ConsoleLoggerService {
     this.classContext = context;
   }
 
-  error(message: any, trace = '', functionContext?: string) {
+  error(message: unknown, trace = '', functionContext?: string) {
     this.printMessage(
       message,
       clc.red,
```
DerMolly, line 27: `ESLint: Unsafe member access .magentaBright on an any value.(@typescript-eslint/no-unsafe-member-access)`

Maybe we should import the colors directly instead of importing clc?
`import { red } from 'cli-color';`

src/users/users.service.ts
```diff
@@ -56,7 +56,7 @@ export class UsersService {
 
   toUserDto(user: User | null | undefined): UserInfoDto | null {
     if (!user) {
-      this.logger.warn(`Recieved ${user} argument!`, 'toUserDto');
+      this.logger.warn(`Recieved ${String(user)} argument!`, 'toUserDto');
```
DerMolly, line 59: here again?

src/logger/console-logger.service.ts
davidmehren: That should be gone. Did you run `yarn install`?

src/auth/auth.service.ts
davidmehren: That does not work, as `null` and `undefined` do not have `.toString()`

src/logger/console-logger.service.ts
DerMolly: I did and this still occurs. Curiously it does not happen when I run `yarn lint`, but only as an error in the IDE.

ErikMichelson: Maybe delete your `.eslintcache`?

DerMolly: The backend doesn't have a `.eslintcache`or at least I don't…

tsconfig.json
```diff
@@ -10,6 +10,6 @@
     "sourceMap": true,
     "outDir": "./dist",
     "baseUrl": "./",
-    "incremental": true
+    "incremental": true,
```
InnayTool, line 13: I don't think this is standards-compliant ([RFC8259](https://tools.ietf.org/html/rfc8259) page 6).

src/api/public/notes/notes.controller.ts
```diff
@@ -188,16 +189,16 @@ export class NotesController {
...
   ): Promise<NoteMetadataDto> {
     try {
       const note = await this.noteService.getNoteByIdOrAlias(noteIdOrAlias);
       if (!this.permissionsService.mayRead(req.user, note)) {
         throw new UnauthorizedException('Reading note denied!');
       }
-      return this.noteService.toNoteMetadataDto(note);
-    } catch (e) {
+      return await this.noteService.toNoteMetadataDto(note);
+    } catch (e: unknown) {
```
InnayTool, line 201: Why is this added here? In other places there is no annotation.

src/main.ts
```diff
@@ -49,4 +49,4 @@ async function bootstrap() {
   logger.log(`Listening on port ${appConfig.port}`, 'AppBootstrap');
 }
 
-bootstrap();
+void bootstrap();
```
InnayTool, line 52: I think the better solution is to add the return type `Promise<void>` to the function.

InnayTool (changes requested):

src/main.ts
davidmehren: That won't help here. [no-floating-promises](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-floating-promises.md) wants us to handle all promises or explicitly ignore them with `void`.

tsconfig.json
davidmehren: fixed

src/api/public/notes/notes.controller.ts
davidmehren: removed

InnayTool (approved):

src/monitoring/monitoring.service.ts
```diff
@@ -16,12 +16,20 @@ let versionCache: null | {
   preRelease?: string;
   commit?: string;
 } = null;
-async function getServerVersionFromPackageJson() {
+async function getServerVersionFromPackageJson(): Promise<{
+  major: number;
+  minor: number;
+  patch: number;
+  preRelease?: string;
+  commit?: string;
```
DerMolly, line 24: Maybe this should be it's own type instead of typed out two times next to each other?

DerMolly (changes requested):

InnayTool (approved):

DerMolly (approved):

PR #937, opened 2021-02-20
Config: Add forbiddenNoteIds to AppConfig

### LINKED ISSUE #519: Which forbidden note ids should we keep?
Which of these via default forbidden note ids should we keep? Which are unnecessary anymore?
I think we want to keep this config option around, since admins will possibly used this to accomplish things.

https://github.com/codimd/server/blob/3461993ee05d09106e276e606b143d472908b166/lib/config/default.js#L36

DerMolly:
### Component/Part
config, notesService

### Description
This PR adds the forbiddenNoteIds to the AppConfig.

Also uses this config in the notes service and the appropriate tests.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
fixes #519

### DIFF
--- a/src/notes/notes.service.ts
+++ b/src/notes/notes.service.ts
@@ -30,6 +30,7 @@ import { NoteUserPermission } from '../permissions/note-user-permission.entity';
 import { NoteGroupPermission } from '../permissions/note-group-permission.entity';
 import { GroupsService } from '../groups/groups.service';
 import { checkArrayForDuplicates } from '../utils/arrayDuplicatCheck';
+import appConfiguration, { AppConfig } from '../config/app.config';
 
 @Injectable()
 export class NotesService {
@@ -41,6 +42,8 @@ export class NotesService {
     @Inject(GroupsService) private groupsService: GroupsService,
     @Inject(forwardRef(() => RevisionsService))
     private revisionsService: RevisionsService,
+    @Inject(appConfiguration.KEY)
+    private appConfig: AppConfig,
   ) {
     this.logger.setContext(NotesService.name);
   }
--- a/src/notes/notes.service.ts
+++ b/src/notes/notes.service.ts
@@ -9,6 +9,7 @@ import { InjectRepository } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
 import {
   AlreadyInDBError,
+  ForbiddenIdError,
   NotInDBError,
   PermissionsUpdateInconsistentError,
 } from '../errors/errors';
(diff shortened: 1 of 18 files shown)


InnayTool (approved):

davidmehren (commented):
Looks good!

Currently, we only check for a forbidden id when creating a new note. We probably also want to do this when accessing already existing notes?

DerMolly:
> Currently, we only check for a forbidden id when creating a new note. We probably also want to do this when accessing already existing notes?

Probably not wrong…

InnayTool (approved):

InnayTool (approved):

src/notes/notes.service.ts
```diff
@@ -148,6 +161,15 @@ export class NotesService {
       `Trying to find note '${noteIdOrAlias}'`,
       'getNoteByIdOrAlias',
     );
+    if (this.appConfig.forbiddenNoteIds.includes(noteIdOrAlias)) {
+      this.logger.debug(
+        `A note with the alias '${noteIdOrAlias}' is forbidden by the administrator.`,
+        'createNote',
```
davidmehren, line 167:
```suggestion
        'getNoteByIdOrAlias',
```

src/notes/notes.service.ts
```diff
@@ -88,6 +92,15 @@ export class NotesService {
     ]);
     if (alias) {
       newNote.alias = alias;
+      if (this.appConfig.forbiddenNoteIds.includes(alias)) {
+        this.logger.debug(
+          `A note with the alias '${alias}' is forbidden by the administrator.`,
+          'createNote',
+        );
+        throw new ForbiddenIdError(
+          `A note with the alias '${alias}' is forbidden by the administrator.`,
```
davidmehren, line 101:
```suggestion
          `Creating a note with the alias '${alias}' is forbidden by the administrator.`,
```

src/notes/notes.service.ts
```diff
@@ -148,6 +161,15 @@ export class NotesService {
       `Trying to find note '${noteIdOrAlias}'`,
       'getNoteByIdOrAlias',
     );
+    if (this.appConfig.forbiddenNoteIds.includes(noteIdOrAlias)) {
+      this.logger.debug(
+        `A note with the alias '${noteIdOrAlias}' is forbidden by the administrator.`,
+        'createNote',
+      );
+      throw new ForbiddenIdError(
+        `A note with the alias '${noteIdOrAlias}' is forbidden by the administrator.`,
```
davidmehren, line 170:
```suggestion
        `Accessing a note with the alias '${noteIdOrAlias}' is forbidden by the administrator.`,
```

davidmehren (changes requested):

src/notes/notes.service.ts
DerMolly: fixed

src/notes/notes.service.ts
DerMolly: fixed

src/notes/notes.service.ts
DerMolly: fixed

davidmehren (approved):

PR #947, opened 2021-02-23
S3Backend: Add S3 MediaBackend

### LINKED ISSUE #536: Add S3 MediaBackend

DerMolly:
### Component/Part
s3 media backend

### Description
This PR adds the s3 media backend.

To start testing it's recommended to use the minio docker container

`docker run -p 9000:9000 minio/minio server /data`

Then you need to create a bucket `test` in the minio web interface at <http://127.0.0.1:9000> and edit it's permissions to read & write.

image
image
image

After that this config in the `.env` file should be added:

```
HD_MEDIA_BACKEND="s3"
HD_MEDIA_BACKEND_S3_ACCESS_KEY="minioadmin"
HD_MEDIA_BACKEND_S3_SECRET_KEY="minioadmin"
HD_MEDIA_BACKEND_S3_BUCKET="test"
HD_MEDIA_BACKEND_S3_ENDPOINT="http://127.0.0.1:9000"
```

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
fixes #536


src/media/backends/s3-backend.ts
```diff
@@ -0,0 +1,78 @@
...
+  async saveFile(
+    buffer: Buffer,
+    fileName: string,
+  ): Promise<[string, BackendData]> {
+    try {
+      await this.client.putObject(this.config.bucket, fileName, buffer);
+      this.logger.log(`Uploaded file ${fileName}`, 'saveFile');
+      return [this.getUrl(fileName), null];
+    } catch (e) {
+      this.logger.error(e.message, e.stack, 'saveFile');
```
InnayTool, line 48: Reject or throw exception here.

src/media/backends/s3-backend.ts
```diff
@@ -0,0 +1,78 @@
...
+    }
+  }
+
+  async deleteFile(fileName: string, _: BackendData): Promise<void> {
+    try {
+      await this.client.removeObject(this.config.bucket, fileName);
+      this.logger.log(`Removed file ${fileName}`, 'saveFile');
+      return;
+    } catch (e) {
+      this.logger.error(e.message, e.stack, 'saveFile');
```
InnayTool, line 58: Reject or throw exception here.

InnayTool (changes requested):

src/config/media.config.ts
```diff
@@ -44,18 +45,29 @@ const mediaSchema = Joi.object({
...
-        secure: Joi.boolean().label('HD_MEDIA_BACKEND_S3_SECURE'),
-        port: Joi.number().label('HD_MEDIA_BACKEND_S3_PORT'),
+        bucket: Joi.string().label('HD_MEDIA_BACKEND_S3_BUCKET'),
+        endPoint: Joi.string()
+          .pattern(/^(?!http).*/)
+          .label('HD_MEDIA_BACKEND_S3_ENDPOINT')
+          .messages({
+            'string.pattern.base':
+              '{#label} can not start with "http" or "https"',
+          }),
```
ErikMichelson, line 60: Is this required to be an IP address?

DerMolly: No it can be a normal hostname, too. But the lib want's the assembly the protocol itself. And I think it's better to let the user break it down instead of trying to get the appropriate parts from a single url extracted…

ErikMichelson: And what about my hostname `http-balancer.example.com`? That would be blocked with your pattern.

ErikMichelson: I therefore suggest to check for the colon after the http/https as well.

```
.pattern(/^(?!https?:).*/)
```

DerMolly: that's a good point. I'll make an appropriate change

SISheogorath: I think we could just throw the URL into a URL object? That should resolve the URL parts for us, without the user need to configure multiple things. Does that sound like an idea?

src/media/backends/s3-backend.ts
DerMolly: fixed

src/media/backends/s3-backend.ts
DerMolly: fixed

src/config/media.config.ts
DerMolly: I changed it, tell me what you think about it this way.

InnayTool (approved):

InnayTool (approved):

InnayTool (dismissed):

InnayTool (approved):

davidmehren (approved):

InnayTool (approved):

PR #961, opened 2021-02-26
ImgurBackend: Add Imgur MediaBackend

### LINKED ISSUE #538: Add imgur MediaBackend

### LINKED ISSUE #957: Remove `getFileUrl` from MediaBackend Interface
### What?

I suggest removing the function `getFileURL(fileName: string, backendData: BackendData): Promise<string>;` as we already return the url in this function `saveFile(buffer: Buffer, fileName: string): Promise<[string, BackendData]>;`. We could save that into the DB and recall that from there.

### Why? 

While building the imgur upload, I found out that we only get the url for the image at the point of uploading the image and never after that.

Another thing to consider is that we can't really change the file url after the upload, because it's part of the note after that point in time (at least most times).

DerMolly:
### Component/Part
Imgur MediaBackend

### Description
This PR adds Imgur MediaBackend.

To test this the following lines should be added to the `.env` file

```
HD_MEDIA_BACKEND="imgur"
HD_MEDIA_BACKEND_IMGUR_CLIENT_ID="$imgurClientId"
```

If you don't want to create an own account and app on imgur you can ask me for my key.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
closes #957 
closes #538


DerMolly:
I alread implemented #957 in this, but am more than happy to change this PR if we find another solution…

src/media/backends/imgur-backend.ts
```diff
@@ -0,0 +1,83 @@
...
+      throw new MediaBackendError(`Could not save '${fileName}' on imgur`);
+    }
+  }
+
+  async deleteFile(fileName: string, backendData: BackendData): Promise<void> {
+    if (backendData === null) {
+      throw new Error();
+    }
+    try {
+      const result = await fetch(
```
InnayTool, line 59: Do we want to add this dependency? couldn't we use `http.request`?

DerMolly: We don't add this dependency. This is already part of the minio and azure…

InnayTool: ok

InnayTool:
Make a remove PR merge it and re-base the others on it?

DerMolly:
I did that. See #964

InnayTool (dismissed):

InnayTool:
`fileUrl` should be added to the plantUML file.

src/media/backends/imgur-backend.ts
```diff
@@ -0,0 +1,83 @@
...
+      return [result.data.link, result.data.deletehash];
+    } catch (e) {
+      this.logger.error(`error: ${e.message}`, e.stack, 'saveFile');
+      throw new MediaBackendError(`Could not save '${fileName}' on imgur`);
+    }
+  }
+
+  async deleteFile(fileName: string, backendData: BackendData): Promise<void> {
+    if (backendData === null) {
+      throw new Error();
```
davidmehren, line 56: Could you make this (hopefully rare) error a bit more descriptive?

davidmehren (changes requested):

src/media/backends/imgur-backend.ts
DerMolly: fixed

InnayTool (dismissed):

InnayTool (approved):

davidmehren (approved):

davidmehren (approved):

InnayTool (approved):

PR #964, opened 2021-02-26
MediaBackendInterface: Remove getFileURL

### LINKED ISSUE #957: Remove `getFileUrl` from MediaBackend Interface

DerMolly:
### Component/Part
media backend interface

### Description
This PR removed the gitFileUrl method from media backend interface.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
closes #957

### DIFF
--- a/src/media/backends/filesystem-backend.ts
+++ b/src/media/backends/filesystem-backend.ts
@@ -53,12 +53,6 @@ export class FilesystemBackend implements MediaBackend {
     }
   }
 
-  getFileURL(fileName: string, _: BackendData): Promise<string> {
-    const filePath = this.getFilePath(fileName);
-    // TODO: Add server address to url
-    return Promise.resolve('/' + filePath);
-  }
-
   private getFilePath(fileName: string): string {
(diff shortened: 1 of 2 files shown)


InnayTool (approved):

InnayTool:
We should remember to remove getFileUrl from filesystem backend.

DerMolly:
I could do this in this PR if you like

DerMolly:
done

InnayTool (approved):

davidmehren (approved):

PR #981, opened 2021-03-03
PrivateAPI: history

DerMolly:
### Component/Part
private api

### Description
This PR adds the history route to the private api.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### DIFF
--- a/src/history/history.service.ts
+++ b/src/history/history.service.ts
@@ -38,7 +38,7 @@ export class HistoryService {
   async getEntriesByUser(user: User): Promise<HistoryEntry[]> {
     return await this.historyEntryRepository.find({
       where: { user: user },
-      relations: ['note'],
+      relations: ['note', 'user'],
     });
   }
 
@@ -134,6 +134,18 @@ export class HistoryService {
     }
     await this.historyEntryRepository.remove(entry);
     return;
+  }
+
+  /**
+   * @async
+   * Delete all history entries of a specific user
+   * @param {User} user - the user that the entry belongs to
+   */
+  async deleteHistory(user: User): Promise<void> {
+    const entries: HistoryEntry[] = await this.getEntriesByUser(user);
+    for (const entry of entries) {
+      await this.historyEntryRepository.remove(entry);
+    }
   }
 
   /**
--- a/src/api/private/me/history/history.controller.spec.ts
+++ b/src/api/private/me/history/history.controller.spec.ts
@@ -1,0 +1,78 @@
+/*
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+import { Test, TestingModule } from '@nestjs/testing';
+import { HistoryController } from './history.controller';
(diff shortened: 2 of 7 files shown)


InnayTool (approved):

InnayTool (changes requested):
see comment above

src/history/history.service.ts
```diff
@@ -93,6 +92,16 @@ export class HistoryService {
     return;
   }
 
+  async deleteHistory(user: User): Promise<void> {
+    const entries: HistoryEntry[] = await this.getEntriesByUser(user);
+    if (!entries) {
```
davidmehren, line 97: When does `getEntriesByUser` return a nullish value?

src/history/history.service.ts
```diff
@@ -93,6 +92,16 @@ export class HistoryService {
     return;
   }
 
+  async deleteHistory(user: User): Promise<void> {
```
davidmehren, line 95: Please add JSDoc annotations for (at least new) service methods

davidmehren (changes requested):
Using a route under `/history` to get the history entries for the current user is inconsistent with the public API, which uses the `/me` route for things that relate to the current user.
Why do you want to change that for the private API?

I'd just move the whole thing to `/me/history` to mirror the public API.

(I realize we already merged `/tokens` which has the same issue and should also be moved)

src/history/history.service.ts
DerMolly: added

src/api/private/history/history.controller.ts
```diff
@@ -0,0 +1,129 @@
...
+    this.logger.setContext(HistoryController.name);
+  }
+
+  @Get()
+  async getHistory(): Promise<HistoryEntryDto[]> {
+    // ToDo: use actual user here
+    try {
+      const user = await this.userService.getUserByUsername('hardcoded');
+      const foundEntries = await this.historyService.getEntriesByUser(user);
+      return await Promise.all(
```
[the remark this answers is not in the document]
DerMolly: we could probably remove that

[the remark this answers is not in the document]
DerMolly: fixed

src/history/history.service.ts
DerMolly: Never, it's just a habit because `findOne` returns `undefined`.

InnayTool (approved):

DerMolly:
> I'd just move the whole thing to /me/history to mirror the public API.

I just did that

InnayTool:
Maybe we should move the `history.controller.ts` to `src/api/private/history/history.controller.ts`.
I think it makes the structure clearer.

DerMolly:
> Maybe we should move the `history.controller.ts` to `src/api/private/history/history.controller.ts`.
> I think it makes the structure clearer.

You probably mean `src/api/private/me/history/history.controller.ts`, don't you?

InnayTool:
I've updated the statement. I copied the wrong path. But maybe your suggestion is more consistent even if I don't like the very long path and the extra sub-directory.

InnayTool (approved):

davidmehren (approved):

Issue #987, opened 2021-03-04
PrivateAPI: Methods needed by the frontend


DerMolly:
I wrote down here all the different routes we would need in the private API to connect to the frontend.
- [ ] explore page (see #3833)
    - [ ] GET `/explore/all` - list of accessible notes, maybe paginated
    - [ ] GET `/explore/created` - list of notes the user has created
- [ ] auth (related to https://github.com/hedgedoc/hedgedoc/issues/1006)
	- [x] POST `/auth/local/*` implementation moved to #1515
	- [x] POST `/auth/ldap/:provider`
	- [ ] GET `/auth/saml/:provider`
	- [ ] GET `/auth/oauth2/:provider`
	- [ ] GET `/auth/gitlab/:provider`
	- [ ] GET `/auth/github`
	- [ ] GET `/auth/facebook`
	- [ ] GET `/auth/twitter`
	- [ ] GET `/auth/dropbox`
	- [ ] GET `/auth/google`
- [x] config (see #980)
	- [x] GET `/config`
- [x] me (see #992)
	- [x] GET `/me` 
	- [x] GET `/me/media`
	- [x] DELETE `/me`
	- [x] POST `/me/profile`
	- [x] history (see #981)
		- [x] POST `/history` (override current history with new history)
		- [x] DELETE `/history`
		- [x] PUT `/history/:noteId` (change pin status)
		- [x] DELETE `/history/:noteId`
- [x] media (see #1057)
	- [x] POST `/media/proxy`
	- [x] POST `/media/upload`
- [x] notes (see #1058)
	- [x] GET `/notes/:noteId`
	- [x] GET `/notes/:noteId/media`
	- [x] POST `/notes/`
	- [x] POST `/notes/:noteId`
	- [x] DELETE `/notes/:noteId`
	- [x] GET `/notes/:noteId/revisions/:timestamp`
	- [x] GET `/notes/:noteId/revisions`
- [x] notes (?) 
	- [x] GET `/notes/:noteId/content`
	- [x] GET `/notes/:noteId/metadata`
	- [x] PUT `/notes/:noteIdOrAlias/metadata/permissions`
	- [x] PUT `/notes/:noteId`
- [x] alias (see #1359) 
	- [x] POST `/alias`
	- [x] PUT `/alias/:alias`
	- [x] DELETE `/alias/:alias`
- [x] tokens
- [x] users
	- [x] GET `/users/:userid`

fvalle1:
[...]

Issue #988, opened 2021-03-04
Handling deleted users


DerMolly:
**Which part of the project should be enhanced?**
private API: user deletion

When the user wants to delete their account, there are a few thing that we definitely need to also delete:
- identity
- auth_token
- session
- note_user_permission
- group_members_user
- history_entry

Others may be kept:
- note (user is owner)
- media_upload

Should we make the user chose if they want to delete those in the frontend? And if so would we want to make each of those separately removable?

DerMolly:
I think it may be desirable to not delete one's notes when deleting the user account. For example when one created a lot of notes for the team. I'd also think that when one doesn't delete all notes, the same should be true for media_uploads.

It may be argued that we don't need this feature as long as it's possible to reassign the owner of a note / media_upload as most users won't want to keep all their notes / media_uploads around, but only a specific few. Which could be done by reassigning those to other users that still need / want them and then deleting everything else…

davidmehren:
> It may be argued that we don't need this feature as long as it's possible to reassign the owner of a note

That sounds like the way to go. #998 is the issue to follow.  
If transfer of ownership is possible, the frontend should show "these notes owned by you will be deleted along with all media files uploaded by you" along with a list of affected notes, when a user wants to delete themselves.

DerMolly:
> the frontend should show "these notes owned by you will be deleted along with all media files uploaded by you" along with a list of affected notes, when a user wants to delete themselves.

open an issue for that in the frontend

DerMolly:
If the solution of delete everything (because things that shouldn't be deleted, can be given to other people) is agreeable, I think we can close this issue…

PR #992, opened 2021-03-04
PrivateAPI: me

DerMolly:
### Component/Part
private api

### Description
This PR adds the me route to the private api.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
#987

### DIFF
--- a/src/api/private/me/me.controller.ts
+++ b/src/api/private/me/me.controller.ts
@@ -1,0 +1,60 @@
+/*
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+
+import { Body, Controller, Delete, Get, HttpCode, Post } from '@nestjs/common';
+import { UserInfoDto } from '../../../users/user-info.dto';
+import { ConsoleLoggerService } from '../../../logger/console-logger.service';
+import { UsersService } from '../../../users/users.service';
+import { MediaService } from '../../../media/media.service';
+import { MediaUploadDto } from '../../../media/media-upload.dto';
+
+@Controller('me')
+export class MeController {
+  constructor(
+    private readonly logger: ConsoleLoggerService,
+    private userService: UsersService,
+    private mediaService: MediaService,
+  ) {
+    this.logger.setContext(MeController.name);
+  }
+
+  @Get()
+  async getMe(): Promise<UserInfoDto> {
+    // ToDo: use actual user here
+    const user = await this.userService.getUserByUsername('hardcoded');
+    return this.userService.toUserDto(user);
+  }
+
+  @Get('media')
+  async getMyMedia(): Promise<MediaUploadDto[]> {
+    // ToDo: use actual user here
+    const user = await this.userService.getUserByUsername('hardcoded');
+    const media = await this.mediaService.listUploadsByUser(user);
+    return media.map((media) => this.mediaService.toMediaUploadDto(media));
+  }
+
+  @Delete()
+  @HttpCode(204)
+  async deleteUser(): Promise<void> {
+    // ToDo: use actual user here
+    const user = await this.userService.getUserByUsername('hardcoded');
+    const mediaUploads = await this.mediaService.listUploadsByUser(user);
+    for (const mediaUpload of mediaUploads) {
+      await this.mediaService.deleteFile(mediaUpload);
+    }
+    this.logger.debug(`Deleted all media uploads of ${user.userName}`);
+    await this.userService.deleteUser(user);
+    this.logger.debug(`Deleted ${user.userName}`);
+  }
+
+  @Post('profile')
+  @HttpCode(200)
+  async updateDisplayName(@Body('name') newDisplayName: string): Promise<void> {
+    // ToDo: use actual user here
+    const user = await this.userService.getUserByUsername('hardcoded');
+    await this.userService.changeDisplayName(user, newDisplayName);
+  }
+}
--- a/src/auth/auth-token.entity.ts
+++ b/src/auth/auth-token.entity.ts
@@ -21,7 +21,9 @@ export class AuthToken {
   @Column({ unique: true })
   keyId: string;
 
-  @ManyToOne((_) => User, (user) => user.authTokens)
+  @ManyToOne((_) => User, (user) => user.authTokens, {
+    onDelete: 'CASCADE', // This deletes the AuthToken, when the associated User is deleted
+  })
   user: User;
 
   @Column()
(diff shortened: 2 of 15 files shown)


DerMolly:
waiting on how do implement delete user in #988

src/users/users.service.ts
```diff
@@ -21,30 +21,81 @@ export class UsersService {
...
+  async changeDisplayName(user: User, displayName: string): Promise<void> {
+    user.displayName = displayName;
+    await this.userRepository.save(user);
+  }
+
+  /**
+   * @async
+   * Get the user specified by the username
+   * @param {string} userName the username by which the user is specified
+   * @param {boolean} [withTokens=false] if the user should contain authTokens
```
InnayTool, line 76:
```suggestion
   * @param {boolean} [withTokens=false] if the returned user object should contain authTokens
```

InnayTool (changes requested):

src/users/users.service.ts
DerMolly: done

src/users/users.service.ts
```diff
@@ -21,30 +21,81 @@ export class UsersService {
...
+    this.logger.debug(
+      `Successfully deleted user with username ${user.userName}`,
+      'deleteUser',
+    );
   }
 
+  /**
+   * @async
+   * Change the displayName of the specified user
+   * @param {User} user - the user to be changesd
```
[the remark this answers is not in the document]
DerMolly: done

src/users/users.service.ts
```diff
@@ -21,30 +21,81 @@ export class UsersService {
...
-    // TODO: Handle owned notes and edits
-    const user = await this.userRepository.findOne({
-      where: { userName: userName },
-    });
-    await this.userRepository.delete(user);
+  /**
+   * @async
+   * Delete the user with the specified userName
+   * @param {User} user - the username of the user to be delete
+   * @throws {NotInDBError} the userName has not user associated with it.
```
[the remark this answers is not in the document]
DerMolly: done

src/media/media.service.ts
```diff
@@ -115,12 +115,14 @@ export class MediaService {
    * Try to delete the file specified by the filename with the user specified by the username.
    * @param {string} filename - the name of the file to delete.
    * @param {string} username - the username of the user who uploaded this file
-   * @return {string} the url of the saved file
    * @throws {PermissionError} the user is not permitted to delete this file.
    * @throws {NotInDBError} - the file entry specified is not in the database
    * @throws {MediaBackendError} - there was an error deleting the file
    */
-  async deleteFile(filename: string, username: string): Promise<void> {
+  async deleteFileByFilename(
```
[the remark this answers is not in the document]
DerMolly: done

InnayTool (approved):

davidmehren (changes requested):

InnayTool (approved):

src/api/private/me/me.controller.ts
```diff
@@ -0,0 +1,60 @@
...
+    const user = await this.userService.getUserByUsername('hardcoded');
+    const media = await this.mediaService.listUploadsByUser(user);
+    return media.map((media) => this.mediaService.toMediaUploadDto(media));
+  }
+
+  @Delete()
+  @HttpCode(204)
+  async deleteUser(): Promise<void> {
+    // ToDo: use actual user here
+    const user = await this.userService.getUserByUsername('hardcoded');
```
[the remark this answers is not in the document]
DerMolly: I think this is not a good idea. For one the we would need to get the media service in the user service. And two we'll add extra parameters to delete the media when the user get's deleted like we did with the notes (https://github.com/hedgedoc/hedgedoc/pull/1094). So moving the whole logic around would be inconsistent.

InnayTool (approved):

InnayTool (dismissed):

src/users/users.service.spec.ts
```diff
@@ -31,15 +34,124 @@ describe('UsersService', () => {
...
+
+  describe('deleteUser', () => {
+    it('works', async () => {
+      const username = 'hardcoded';
+      const displayname = 'Testy';
+      const user = User.create(username, displayname) as User;
+      jest.spyOn(userRepo, 'remove').mockImplementationOnce(
+        // eslint-disable-next-line @typescript-eslint/require-await
+        async (user: User): Promise<User> => {
+          expect(user).toEqual(user);
```
InnayTool, line 81: I think this does not what it is supposed to do. The second `user` variable covers the other one.

InnayTool (changes requested):

src/users/users.service.spec.ts
DerMolly: good point. Changed…

InnayTool (approved):

src/users/users.service.ts
```diff
@@ -31,21 +31,42 @@ export class UsersService {
...
 
   /**
    * @async
    * Delete the user with the specified userName
-   * @param userName - the username of the user to be delete
+   * @param {User} user - the username of the user to be delete
    * @throws {NotInDBError} the userName has no user associated with it.
    */
-  async deleteUser(userName: string): Promise<void> {
-    // TODO: Handle owned notes and edits
```
davidmehren, line 30: Do we already handle this?

src/auth/auth-token.entity.ts
```diff
@@ -21,7 +21,9 @@ export class AuthToken {
   @Column({ unique: true })
   keyId: string;
 
-  @ManyToOne((_) => User, (user) => user.authTokens)
+  @ManyToOne((_) => User, (user) => user.authTokens, {
+    onDelete: 'CASCADE',
```
davidmehren, line 25: Please also document the usages of 'cascade'

davidmehren (changes requested):

src/users/users.service.ts
DerMolly: The CASCADE deletes the notes. The revisions (which I presume is meant by edits here) are kept, because they aren't directly associated with the user.

src/auth/auth-token.entity.ts
DerMolly: I did so

src/users/users.service.ts
DerMolly: Only thing is, that revisions have authorships, which in turn refer to User, but they don't get deleted, when the user is deleted (I think this is the correct way of handling this), but we than don't get any user at all.

DerMolly: To be clear revsions of your own notes get deleted, when the notes are deleted. But a user also has authorship of revisions in other notes…

davidmehren: So the authorships stay in the DB, but get `NULL` set in their `user` foreign key?

InnayTool (approved):

PR #1022, opened 2021-03-11
[1.x] Templates: Remove lang and add translation parameter

### LINKED ISSUE #881: Problems with Google Translator
**Description**

When using hedgedoc with Chrome and Google Translator (GT) set to always translate English to another language, things break horribly. Viewing documents and editing documents is not possible anymore as only garbage comes out and is shown, especially if the document is not actually written in English.

**Expected behavior**

Since the page and document was already in German, nothing should have been translated. There are two separate problems from what I can tell:
1. `<html lang="en">` seems to be hardcoded, even if a translated version is served. This forces GT to translate anything it finds, even it is not English. Not really your problem, but `lang="en"` is just not correct if serving translations.
2. [`class="notranslate"`](https://stackoverflow.com/questions/12238396/how-to-disable-google-translate-from-html-in-chrome) should be added to pad content to disable Google Translator at least when in an editing mode as it interferes with the input.

DerMolly:
### Component/Part
render templates

### Description

Since the interface is not always in english, we mostly removed the lang attribute from all html tags. Since the error messages in error.ejs are not translated, but always in english, there the global lang="en" should be kept.
Also in the slide and editor template the div, which contains the user generated text, has the attribute translate="no" now, to avoid unwanted translations.
Since on the publish view (pretty.ejs) only the user generated content is shown, we set the lang to the language defined in yaml (or 'en') as a default, but that was also moved to the corresponding markdown div instead of html.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
Fixes #881
See also #437

### DIFF
--- a/public/views/pretty.ejs
+++ b/public/views/pretty.ejs
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="<%= lang || "en" %>">
+<html>
 
 <head>
     <meta charset="utf-8">
@@ -60,7 +60,7 @@
             <% } %>
         </small>
     </div>
-    <div id="doc" class="container markdown-body"><%= body %></div>
+    <div id="doc" class="container markdown-body" <% if (lang) { %> lang="<%= lang %>"<% } %>><%= body %></div>
     <div class="ui-toc dropup unselectable hidden-print" style="display:none;">
         <div class="pull-right dropdown">
             <a id="tocLabel" class="ui-toc-label btn btn-default" data-toggle="dropdown" href="#" role="button" aria-haspopup="true" aria-expanded="false" title="Table of content">
--- a/public/views/hedgedoc.ejs
+++ b/public/views/hedgedoc.ejs
@@ -1,11 +1,11 @@
  <!DOCTYPE html>
-<html lang="en">
+<html>
 
 <head>
     <%- include('hedgedoc/head') %>
 </head>
 
-<body>
+<body translate="no">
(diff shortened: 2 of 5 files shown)


davidmehren:
We should disable translations for the whole editor. I just tested it and Chrome *will* translate the editor content while you type.

DerMolly:
> We should disable translations for the whole editor. I just tested it and Chrome _will_ translate the editor content while you type.

I changed the location of the `notranslate` around, what do you think?

davidmehren (dismissed):

InnayTool (approved):

public/views/pretty.ejs
```diff
@@ -60,7 +60,7 @@
             <% } %>
         </small>
     </div>
-    <div id="doc" class="container markdown-body"><%= body %></div>
+    <div id="doc" class="container markdown-body" lang="<%= lang || "en" %>"><%= body %></div>
```
ErikMichelson, line 63: Why do we have "en" as fallback here? Wouldn't it be better to just include this attribute when the `lang`-frontmatter field is set?

DerMolly: Well that was your change I just moved it, but we could do that of course.

DerMolly: fixed

ErikMichelson (approved):

InnayTool (approved):

davidmehren (approved):

PR #1033, opened 2021-03-15
Add dev docs about deleting notes & revisions

### LINKED ISSUE #560: Delete Note revisions
Should the owner of a note be able to
- delete specific revisions
- only delete all revisions
- delete no revisions at all?

In the last case it would still be possible to delete the note and recreate it with the last content, which is the same effect as "delete all revisions".

Note that in all cases, only the owner of a note would be able to delete something. Users with write permissions have the ability to "reset" a note to a previous revision, but that only appends a new revision with the content of that previous revision, so no content gets lost.

davidmehren:
### Component/Part
Dev documentation

### Description
This PR adds dev docs about deleting notes & revisions

### Steps

  made sure that:

### Related Issue(s)
Closes #560

### DIFF
--- a/docs/content/dev/2.0.md
+++ b/docs/content/dev/2.0.md
@@ -14,6 +14,15 @@ The software provides two special groups which have no explicit users:
 - `everyone` (Describing that everyone who wants to access a note can do if it is enabled in the config.)
 - `loggedIn` (Describing all users which are logged in)
 
+## Deleting notes and revisions
+- The owner of a note may delete it.
+    - By default, this also removes all revisions and all files that were uploaded to that note.
+    - The owner may choose to skip deleting associated uploads, leaving them without a note.
+    - The frontend should show a list of all uploads that will be affected
+      and provide a method of skipping deletion. 
+- The owner of a note may delete all revisions. This effectively purges the edit
+  history of a note.
+
 ## Dev Setup
 - Clone backend and frontend in two folders.
 - Run `yarn install` in both projects


DerMolly (dismissed):

docs/content/dev/2.0.md
```diff
@@ -14,6 +14,12 @@ The software provides two special groups which have no explicit users:
 - `everyone` (Describing that everyone who wants to access a note can do if it is enabled in the config.)
 - `loggedIn` (Describing all users which are logged in)
 
+## Deleting notes and revisions
+- The owner of a note may delete it. This also removes all revisions and all files
+  that were uploaded to that note.
```
SISheogorath, line 19: I'm not sure how ideal this will work for some use-cases. Images that are uploaded to a document might be re-used in other documents, since it's normal in Markdown to do that. This would only make sense if we inline pictures into markdown.

Otherwise I would consider to opt for attachments being managed independently from note deletions.

DerMolly: A thing that just occurred to me:
What about uploads of other people to my note, that I want to delete? Should they be removed, too? Or should they just lose their note attachment? I mean on the one hand they uploaded an image to somebody else's note and therefore one could argue if the note get's deleted so should the image, but I'm not sure how well that flies.

davidmehren: IIRC, the idea we contemplated to solve that was to let the frontend automatically re-upload images that are copy-pasted between notes, but granted, that might be a somewhat fragile solution.

I really like to have uploads attached to notes, as that provides a simple way to delete all data associated with a note. I'd argue that this behavior is more in line with what a user expects when pressing the delete button. The frontend could make this more obvious by showing a list of uploads that will be deleted when deleting the note.

davidmehren: We could also make deleting the associated uploads optional. I'm not super happy with that solution, as a user might forget they didn't delete an image which then stays online forever, but I guess that can be mitigated by the "files uploaded by me" list in the user profile.

ErikMichelson: From frontend perspective the first thing that comes to my mind would be an additional checkbox in the note deletion modal asking whether attached images should be removed too.
This way a user that wants to delete their note has to think about attached images.

SISheogorath: As a User I would definitely not expect to have images deleted when I delete a note. It's not the first time, that I would copy a note and continue to use that content independent from the original note. Means my image links would still be the same. Unless you want to write a parser that re-write the image URL inside my note, I don't think this would work.

And if you do, I would be quite unhappy that the application I use to write markdown, simply decides what image URLs I should use.

DerMolly: But the current situation of once uploaded you basically can't remove images ever is also not feasible. From a legal perspective at least. I think giving the option to delete media that's associated with a note is an okay way to go.

SISheogorath: I would rather associate media with users who uploaded it. This would leave the user in control, without breaking notes unexpectedly.

I mean, we need this anyway, when we decide not to require the media deletion for notes.

DerMolly: Currently an upload is associated with a user and a note.

davidmehren:
I have updated the PR to make deletion of the uploads optional.

It is currently planned that uploads are associated with the user that uploaded the file and the note the file was uploaded to. Even if the note is deleted, it will still be possible to delete an upload using a "files I uploaded" list in the user profile.

InnayTool (approved):

DerMolly (approved):

SISheogorath (approved):

PR #1117, opened 2021-04-08
Use upstream imgur library

mrdrogdrog:
### Component/Part
Imgur media upload

### Description
This PR replaces the dependency `node-imgur`.
We use a hackmd fork of "node-imgur" for no reason. The upstream lib is way more advanced and modern.

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/lib/web/imageRouter/imgur.js
+++ b/lib/web/imageRouter/imgur.js
@@ -1,8 +1,8 @@
 'use strict'
 const config = require('../../config')
 const logger = require('../../logger')
-
-const imgur = require('imgur')
+const fs = require('fs')
+const fetch = require('node-fetch')
 
 exports.uploadImage = function (imagePath, callback) {
   if (!callback || typeof callback !== 'function') {
@@ -15,12 +15,30 @@ exports.uploadImage = function (imagePath, callback) {
     return
   }
 
-  imgur.setClientId(config.imgur.clientID)
-  imgur.uploadFile(imagePath)
-    .then(function (json) {
+  // The following client ID is for use with HedgeDoc only
+  const clientId = config.imgur.clientID || '032aa2f687790cd'
+
+  const buffer = fs.readFileSync(imagePath)
+
+  const params = new URLSearchParams()
+  params.append('image', buffer.toString('base64'))
+  params.append('type', 'base64')
+  fetch('https://api.imgur.com/3/image', {
+    method: 'POST',
+    body: params,
+    headers: { Authorization: `Client-ID ${clientId}` }
+  })
+    .then((res) => {
+      if (!res.ok) {
+        callback(new Error(res.statusText), null)
+        return
+      }
+      return res.json()
+    })
+    .then((json) => {
       logger.debug(`SERVER uploadimage success: ${JSON.stringify(json)}`)
       callback(null, json.data.link.replace(/^http:\/\//i, 'https://'))
-    }).catch(function (err) {
+    }).catch((err) => {
       callback(new Error(err), null)
     })
 }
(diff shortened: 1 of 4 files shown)


DerMolly (approved):

InnayTool (approved):

davidmehren:
It might be a better idea to backport https://github.com/hedgedoc/hedgedoc/blob/develop/src/media/backends/imgur-backend.ts which just uses `node-fetch`, as that saves us from adding a pile of new dependencies.

DerMolly:
> It might be a better idea to backport [https://github.com/hedgedoc/hedgedoc/blob/develop/src/media/backends/imgur-backend.ts](https://github.com/hedgedoc/hedgedoc/blob/develop/src/media/backends/imgur-backend.ts?rgh-link-date=2021-04-15T20%3A21%3A33Z) which just uses `node-fetch`, as that saves us from adding a pile of new dependencies.

I just did that.

lib/web/imageRouter/imgur.js
```diff
@@ -15,12 +15,25 @@ exports.uploadImage = function (imagePath, callback) {
...
-    .then(function (json) {
+  const buffer = fs.readFileSync(imagePath)
+
+  const params = new URLSearchParams()
+  params.append('image', buffer.toString('base64'))
+  params.append('type', 'base64')
+  fetch('https://api.imgur.com/3/image', {
+    method: 'POST',
+    body: params,
+    headers: { Authorization: `Client-ID ${config.imgur.clientID}` }
```
mrdrogdrog, line 26: @DerMolly 

What if you don't provide an client id? The imgur lib defined a fallback id for this purpose. Maybe we should also fallback to it.

DerMolly: @SISheogorath Does the current imgur support uploads without configuring a client id?

DerMolly: If so we should probably continue to do so…

mrdrogdrog: If you inspect the code.. yes

setClientID accepts only strings. If you don't set the ID in your config, then the lib will continue to use the build-in ID.

mrdrogdrog: https://github.com/hackmdio/node-imgur/blob/de0a7a1f1eb2cb6628385fedb990ad396a190573/lib/imgur.js#L285

DerMolly: Do we want to continue to use that ID or do we want our own? And do we want to continue this behavior in 2.0 (currently it fails with no clientId)?

davidmehren: I think without a client ID, deletion of uploads is impossible?
If so, we should not allow an empty client id.

mrdrogdrog: To require a client id now would be a breaking change. Therefore we should keep a fallback id for now for 1.x

DerMolly: I restored the default behavior for this PR

lib/web/imageRouter/imgur.js
```diff
@@ -15,12 +15,27 @@ exports.uploadImage = function (imagePath, callback) {
     return
   }
 
-  imgur.setClientId(config.imgur.clientID)
-  imgur.uploadFile(imagePath)
-    .then(function (json) {
+  const clientId = config.imgur.clientID || 'f0ea04148a54268'
```
mrdrogdrog, line 18: @DerMolly 
Use 032aa2f687790cd. But please add a comment.

DerMolly: Done

InnayTool (dismissed):

davidmehren (approved):

ErikMichelson (approved):

InnayTool (approved):

PR #1127, opened 2021-04-10
PrivateAPI: Rename HistoryEntryCreationDto to HistoryEntryImportDto

DerMolly:
### Component/Part
private api

### Description
This PR improves the private apis `HistoryEntryCreationDto` by
1. renaming it to `HistoryEntryImportDto` as it is used to import a whole list of history entries rather than creating a single history entry
2. adding the `pinStatus` field to allow the frontend to maintain the pinStatus of historyEntries through the POST
3. adding the `lastVisited` for the same reason

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### DIFF
--- a/src/history/history-entry-import.dto.ts
+++ b/src/history/history-entry-import.dto.ts
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
-import { IsString } from 'class-validator';
+import { IsBoolean, IsString } from 'class-validator';
 
 export class HistoryEntryImportDto {
   /**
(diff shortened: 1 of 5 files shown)


src/history/history-entry-import.dto.ts
```diff
@@ -4,12 +4,18 @@
...
    * ID or Alias of the note
    */
   @IsString()
   note: string;
+  /**
+   * True if the note should be pinned
+   * @example true
+   */
+  @IsBoolean()
+  pinStatus: boolean;
```
ErikMichelson, line 20: We also need the `lastVisited` time here as that can be imported from the frontend too. It should not be set to the time of the import by the backend.

DerMolly: added

DerMolly:
While working on PR #1130, which is based of this PR I removed the changes to:
- history service
- history service test

Therefore they most likely can be ignored.

InnayTool (approved):

davidmehren (approved):

PR #1130, opened 2021-04-12
HistoryService: Add setHistory method 

### LINKED ISSUE #1126: PrivateAPI: Histroy should be completly uploaded or not changed at all
**Description**
When uploading a new history if the frontend sends us garbage, the history is first deleted and then the process fails. This should not happen. Either the history upload worked or nothing should be changed.

Also in case of a not found note (NotInDbError) the API should not return 404 but 400 `Bad Request`, because `Not Found` is a strange response for a `POST` request

**Expected behavior**
When an error occurs during posting a new history, the old history should prevail

**Additional context**
This can probably be fixed by using transactions:
https://github.com/typeorm/typeorm/blob/master/docs/transactions.md

DerMolly:
### Component/Part
history service

### Description
This PR reimplements the business logic of the history controllers setHistory method (of the private api) in the history serivce in  a transactional way. This should prevent the problem that the history gets deleted, but a later error in the handling of the list of HistoryEntryImportDto let's the call fail.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
fixes #1126

### DIFF
--- a/src/history/history.service.ts
+++ b/src/history/history.service.ts
@@ -80,25 +80,15 @@ export class HistoryService {
    * Create or update a history entry by the user and note. If the entry is merely updated the updatedAt date is set to the current date.
    * @param {Note} note - the note that the history entry belongs to
    * @param {User} user - the user that the history entry belongs to
-   * @param {boolean} pinStatus - if the pinStatus of the history entry should be set
-   * @param {Date} lastVisited - the last time the associated note was accessed
    * @return {HistoryEntry} the requested history entry
    */
   async createOrUpdateHistoryEntry(
     note: Note,
     user: User,
-    pinStatus?: boolean,
-    lastVisited?: Date,
   ): Promise<HistoryEntry> {
     let entry = await this.getEntryByNote(note, user);
     if (!entry) {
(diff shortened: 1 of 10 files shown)


DerMolly:
This PR is ready for review, but will be marked as such, after the merging of #1127

InnayTool (approved):

src/history/history.service.ts
```diff
@@ -148,6 +151,54 @@ export class HistoryService {
...
+              alias: historyEntry.note,
+            },
+          ],
+        });
+        if (note === undefined) {
+          this.logger.debug(
+            `Could not find note '${historyEntry.note}'`,
+            'getNoteByIdOrAlias',
+          );
+          throw new NotInDBError(
```
davidmehren, line 190: I suppose throwing automagically makes the manager rollback the transaction?

davidmehren (approved):

src/history/history.service.ts
DerMolly: That's how I'm understanding https://docs.nestjs.com/techniques/database#transactions

davidmehren (approved):

PR #1131, opened 2021-04-12
Logging: Add logging configuration

### LINKED ISSUE #1097: How to configure loglevels
Looking at NestJS they support the debug levels
```
'log' | 'error' | 'warn' | 'debug' | 'verbose'
```

which can be individually included. Like so:
```
const app = await NestFactory.create(AppModule, {
  logger: ['error', 'warn'],
});
```
or
```
const app = await NestFactory.create(AppModule);
app.useLogger(['error', 'warn']);
```

Our backend currently defines these log levels:
```
export enum Loglevel {
  TRACE = 'trace',
  DEBUG = 'debug',
  INFO = 'info',
  WARN = 'warn',
  ERROR = 'error',
}
```

The intent with these loglevels was that they are additive. So I'd suggest the following configuration
```
our loglevel -> nest log level config
ERROR -> ['error']
WARN -> ['error', 'warn']
INFO -> ['error', 'warn', 'log']
DEBUG -> ['error', 'warn', 'log', 'debug']
TRACE -> ['error', 'warn', 'log', 'debug', 'verbose']
```

DerMolly:
### Component/Part
logging

### Description
This PR add the functionality to configure which log messages should be shown on the basis of the appConfigs loglevel.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
fixes #1097


InnayTool (approved):

InnayTool (dismissed):

InnayTool (dismissed):

src/logger/console-logger.service.ts
```diff
@@ -29,43 +35,63 @@ export class ConsoleLoggerService {
...
 
   log(message: unknown, functionContext?: string): void {
-    this.printMessage(
-      message,
-      clc.green,
-      this.makeContextString(functionContext),
-      false,
-    );
+    if (
+      this.appConfig.loglevel === Loglevel.INFO ||
```
davidmehren, line 43: Looking at these statements, I'm wondering why we don't define an order of the levels and just use something like `loglevel <= Loglevel.INFO`

davidmehren (changes requested):

src/logger/console-logger.service.ts
DerMolly: I build something similar with a helper function…

src/config/utils.ts
```diff
@@ -87,3 +89,27 @@ export function replaceAuthErrorsWithEnvironmentVariables(
   message = message.replace('.accessRole', '_ACCESS_ROLE');
   return message;
 }
+
+export function needToLog(
+  currentLoglevel: Loglevel,
+  requestedLoglevel,
```
InnayTool, line 95: Is there a reason this is not 
```suggestion
  requestedLoglevel: LogLevel,
```
?

InnayTool (changes requested):

src/config/utils.ts
DerMolly: Fixed

InnayTool (approved):

davidmehren (approved):

PR #1132, opened 2021-04-12
Guest notes auto-lock

SISheogorath:
### Component/Part
Guest Notes

### Description
This patch adds a basic implementation for an auto-lock feature for guest
notes. It uses the config options `CMD_LOCKGUESTNOTES_TIMEOUT` to set
a timeout in seconds when a note should be locked. As well as
`CMD_LOCKGUESTNOTES_PERMISSION` to set the desired permission.

It is possible to extend this feature later on to allow a deletion of
notes or to lock the note based on the absence of activity instead of
statically based on its creation date. However this patch doesn't
facilitate those possibilities.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
https://matrix.to/#/!ybaSlrrMNLeuoQuUuR:matrix.org/$1618218839141841VlNVd:matrix.org?via=matrix.org&via=kif.rocks&via=tchncs.de

### DIFF
--- a/app.js
+++ b/app.js
@@ -276,6 +276,13 @@ models.sequelize.authenticate().then(function () {
     } else {
       throw new Error('server still not ready after db synced')
     }
+    // Allow to timeout-lock guest notes so they can no longer be changed after the timeout period
+    if (config.lockGuestNotes.timeout > 0) {
+      setInterval(models.Note.lockOldGuestNotes, (config.lockGuestNotes.timeout / 2) * 1000, config.lockGuestNotes.timeout, config.lockGuestNotes.permission)
+      // Additional call of the lockOldGuestNotes function to make sure notes that weren't locked during the last app run, can get locked.
+      // We run this 30 seconds after start, to give the app a bit of warmup, but not waiting too long.
+      setTimeout(models.Note.lockOldGuestNotes, 30 * 1000, config.lockGuestNotes.timeout, config.lockGuestNotes.permission)
+    }
   })
 })
 
--- a/docs/content/configuration.md
+++ b/docs/content/configuration.md
@@ -102,6 +102,14 @@ these are rarely used for various reasons.
 | `sessionLife`         | `CMD_SESSION_LIFE`          | **`14 * 24 * 60 * 60 * 1000`**, `1209600000` (14 days)                  | Cookie session life time in milliseconds.                                                                                                                      |
 | `sessionSecret`       | `CMD_SESSION_SECRET`        | **`secret`**                                                            | Cookie session secret used to sign the session cookie. If none is set, one will randomly generated on each startup, meaning all your users will be logged out. |
 
+## Guest Notes autolock
+
+| config file           | environment                 | **default** and example value                                           | description                                                                                                                                                    |
+| --------------------- | --------------------------- | ----------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `lockGuestNotes` | | `{ timeout: 0,  permission: 'locked' }` | If the `timeout` setting is `>` 0, the autolock feature is activated. It'll automatically set the permission of guest notes, which have a creation date older than the timeout setting, to the permission defined in the `permission` setting. |
+| | `CMD_LOCKGUESTNOTES_TIMEOUT` | `0` | This is the maximum age in seconds a note is allowed to have before before the `CMD_LOCKGUESTNOTES_PERMISSION` is applied to it. If the timeout is set to 0, the autolock feature is disabled |
+| | `CMD_LOCKGUESTNOTES_PERMISSION` | `locked` | This variable defines the permission used when locking guest notes. |
+
 ## Login methods
 
 ### Email (local account)
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -80,6 +80,15 @@ if (!(config.defaultPermission in config.permission)) {
   config.defaultPermission = config.permission.editable
 }
 
+if (!(config.lockGuestNotes.permission in config.permission)) {
+  logger.error('Invalid lockGuestNotes permission: %s\nAvailable permissions are: %s', config.lockGuestNotes, Object.keys(config.permission).toString())
+}
+
+if (!(config.lockGuestNotes.timeout >= 0)) {
+  logger.warn('Invalid lockGuestNotes timeout. Timeout has to be positive number. Disabled feature.')
+  config.lockGuestNotes.timeout = 0
+}
+
 // cache result, cannot change config in runtime!!!
 config.isStandardHTTPsPort = (function isStandardHTTPsPort () {
   return config.useSSL && config.port === 443
--- a/lib/models/note.js
+++ b/lib/models/note.js
@@ -22,6 +22,8 @@ const logger = require('../logger')
 // ot
 const ot = require('../ot')
 
+const Op = Sequelize.Op
+
 // permission types
 const permissionTypes = ['freely', 'editable', 'limited', 'locked', 'protected', 'private']
 
@@ -582,6 +584,35 @@ module.exports = function (sequelize, DataTypes) {
       }
     }
     return operations
+  }
+
+  Note.lockOldGuestNotes = async function (secondsUntilLock, permission) {
+    if (typeof secondsUntilLock !== 'number') {
+      throw new Error("Parameter 'secondsUntilLock' has to be of type 'number'")
+    }
+    if (!permissionTypes.includes(permission)) {
+      throw new Error('Unexpected permission for locking old guest notes')
+    }
+    const notes = await Note.findAll({
+      attributes: ['id'],
+      where: {
+        // Only guest notes
+        ownerId: null,
+        // Select only notes whose permissions require adjustment
+        permission: {
+          [Op.ne]: permission
+        },
+        // Select only notes who are beyond the timeout
+        createdAt: {
+          [Op.lt]: new Date(new Date() - secondsUntilLock * 1000)
+        }
+      }
+    })
+    notes.forEach(async function (note) {
+      note.permission = permission
+      await note.save()
+      logger.info(`Locked guest note ${note.id}`)
+    })
   }
 
   return Note
--- a/lib/realtime.js
+++ b/lib/realtime.js
@@ -625,31 +625,42 @@ function updateUserData (socket, user) {
 function ifMayEdit (socket, callback) {
   const noteId = socket.noteId
   if (!noteId || !notes[noteId]) return
-  const note = notes[noteId]
-  let mayEdit = true
-  switch (note.permission) {
-    case 'freely':
-      // not blocking anyone
-      break
-    case 'editable': case 'limited':
-      // only login user can change
-      if (!socket.request.user || !socket.request.user.logged_in) { mayEdit = false }
-      break
-    case 'locked': case 'private': case 'protected':
-      // only owner can change
-      if (!note.owner || note.owner !== socket.request.user.id) { mayEdit = false }
-      break
-  }
-  // if user may edit and this is a text operation
-  if (socket.origin === 'operation' && mayEdit) {
-    // save for the last change user id
-    if (socket.request.user && socket.request.user.logged_in) {
-      note.lastchangeuser = socket.request.user.id
-    } else {
-      note.lastchangeuser = null
+  models.Note.findOne({
+    where: {
+      id: noteId
+    },
+    attributes: ['id', 'permission']
+  }).then(function (note) {
+    let mayEdit = true
+    switch (note.permission) {
+      case 'freely':
+        // not blocking anyone
+        break
+      case 'editable': case 'limited':
+        // only login user can change
+        if (!socket.request.user || !socket.request.user.logged_in) { mayEdit = false }
+        break
+      case 'locked': case 'private': case 'protected':
+        // only owner can change
+        if (!note.owner || note.owner !== socket.request.user.id) { mayEdit = false }
+        break
     }
-  }
-  return callback(mayEdit)
+    // if user may edit and this is a text operation
+    if (socket.origin === 'operation' && mayEdit) {
+      // save for the last change user id
+      if (socket.request.user && socket.request.user.logged_in) {
+        note.lastchangeuser = socket.request.user.id
+      } else {
+        note.lastchangeuser = null
+      }
+    }
+    return callback(mayEdit)
+  }).catch(function (error) {
+    logger.error(error)
+    // false is required here in as the callback function only expects a boolean.
+    // Not accepting changes the best way to handle such an unknown situation is to not accept changes.
+    return callback(false) // eslint-disable-line node/no-callback-literal
+  })
 }
 
 function operationCallback (socket, operation) {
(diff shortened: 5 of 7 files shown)


lib/models/note.js
```diff
@@ -584,5 +586,34 @@ module.exports = function (sequelize, DataTypes) {
     return operations
   }
 
+  Note.lockOldGuestNotes = async function (seconds, permission) {
```
DerMolly, line 589: I think the parameter `seconds` should be called `secondsUntilLock` or something along those lines.

docs/content/configuration.md
```diff
@@ -102,6 +102,13 @@ these are rarely used for various reasons.
 | `sessionLife`         | `CMD_SESSION_LIFE`          | **`14 * 24 * 60 * 60 * 1000`**, `1209600000` (14 days)                  | Cookie session life time in milliseconds.                                                                                                                      |
 | `sessionSecret`       | `CMD_SESSION_SECRET`        | **`secret`**                                                            | Cookie session secret used to sign the session cookie. If none is set, one will randomly generated on each startup, meaning all your users will be logged out. |
 
+## Guest Notes autolock
+
+| config file           | environment                 | **default** and example value                                           | description                                                                                                                                                    |
```
DerMolly, line 107:
```suggestion
| config file           | environment                 | **default** and example value                                           | description                                                                                                                                                    |
| --------------------- | --------------------------- | ----------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
```

You missed the header / table divider

lib/models/note.js
```diff
@@ -584,5 +586,34 @@ module.exports = function (sequelize, DataTypes) {
...
+    if (typeof seconds !== 'number') {
+      throw new Error("Parameter 'seconds' has to be of type 'number'")
+    }
+    if (!permissionTypes.includes(permission)) {
+      throw new Error('Unexpected permission for locking old guest notes')
+    }
+    const notes = await Note.findAll({
+      attributes: ['id', 'ownerId', 'permission', 'createdAt'],
```
DerMolly, line 597: Do we need the extra fields in the results, because we want to filter by them or could they be omitted?

docs/content/configuration.md
DerMolly: image

DerMolly (changes requested):
I found some small problem

docs/content/configuration.md
SISheogorath: Hope that's fixed now, thanks for pointing it out!

lib/models/note.js
SISheogorath: Originally I added those since they are all used in the request, but dropped it down. Main reason to specify attributes in first place, so the database doesn't pull all content each time we try to expire the notes.

docs/content/configuration.md
DerMolly: looks good to me

docs/content/configuration.md
```diff
@@ -102,6 +102,14 @@ these are rarely used for various reasons.
...
+| --------------------- | --------------------------- | ----------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `lockGuestNotes` | | `{ timeout: 0,  permission: 'locked' }` | If the `timeout` setting is `>` 0, the autolock feature is activated. It'll automatically set the permission of guest notes, that exist longer than defined in the timeout setting, to the permission defined in the `permission` setting. |
```
ccoenen, line 109: these are missing what the base of the timeout is. It should be noted that the timeout starts from creation, not modification date.

docs/content/configuration.md
```diff
@@ -102,6 +102,14 @@ these are rarely used for various reasons.
...
+| `lockGuestNotes` | `CMD_LOCKGUESTNOTES_TIMEOUT` | `0` | This is the maximum age in seconds a note is allowed to have before before the `CMD_LOCKGUESTNOTES_PERMISSION` is applied to it. If the timeout is set to 0, the autolock feature is disabled |
```
ccoenen, line 110: for some reason, `lockGuestNotes` is in here, twice?

lib/config/index.js
```diff
@@ -80,6 +80,15 @@ if (!(config.defaultPermission in config.permission)) {
   config.defaultPermission = config.permission.editable
 }
 
+if (!(config.lockGuestNotes.permission in config.permission)) {
+  logger.error('Invalid lockGuestNotes permission: %s\nAvailable permissions are: %s', config.lockGuestNotes, Object.keys(config.permission).toString())
+}
+
+if (config.lockGuestNotes.timeout < 0) {
```
ccoenen, line 87: Is there some place where this is already coerced into a number? What would happen if I set `timeout` to `"tuesday"`?
If not, this needs a `parseInt(..., 10);`.

DerMolly (dismissed):

app.js
```diff
@@ -276,6 +276,10 @@ models.sequelize.authenticate().then(function () {
     } else {
       throw new Error('server still not ready after db synced')
     }
+    // Allow to timeout-lock guest notes so they can no longer be changed after the timeout period
+    if (config.lockGuestNotes.timeout > 0) {
+      setInterval(models.Note.lockOldGuestNotes, (config.lockGuestNotes.timeout / 2) * 1000, config.lockGuestNotes.timeout, config.lockGuestNotes.permission)
```
ccoenen, line 281: worst case, this is off by half of `timeout / 2`. So I can't actually be certain that 1 day is actually one day, it might be just under 36 hours. Depending on how important this is, we could query the newest guest note and set a `setTimeout` specifically for its timout time. We would also only have to perform this check once at server start and once per created note.
Again: this might be overkill.

lib/realtime.js
```diff
@@ -625,31 +625,42 @@ function updateUserData (socket, user) {
...
-      note.lastchangeuser = socket.request.user.id
-    } else {
-      note.lastchangeuser = null
+  models.Note.findOne({
+    where: {
+      id: noteId
+    },
+    attributes: ['id', 'permission']
+  }).then(function (note) {
+    let mayEdit = true
```
ccoenen, line 634: if this is edited anyway: I'd start with `mayEdit = false` and then allow it when criteria are met (currently it's the other way round). This would remove a ton of negations down in the case statements, making it clearer overall.

docs/content/configuration.md
DerMolly: Once explaining the Config via config.json, once via environment variables…

docs/content/configuration.md
SISheogorath: okay, maybe the "that exist longer than defined in the timeout setting" didn't make clear, that creation date is used. I'll find a better explanation ^^

lib/config/index.js
SISheogorath: Funnily enough inverting the expression is enough: `(!(config.lockGuestNotes.timeout >= 0))`

app.js
SISheogorath: I would rather go for a time constant than for an individual timeout, since it should be less stressful for the node application.

Unless someone explicitly complaints about the timeout inaccuracy, it should be fine. The original idea of the implementation was based on using a cron-job, so I would say we are close enough 😉

lib/realtime.js
SISheogorath: I hope the new commit fixes this to your liking :)

docs/content/configuration.md
SISheogorath: Fixed, this one was left in by mistake, i leave the `config file` part empty for the environment variables

app.js
ccoenen: that would be fixed by running it once on startup.

lib/realtime.js
ccoenen: In my opinion the new commit is much more readable. Thanks!

ccoenen (dismissed):

SISheogorath:
Honestly speaking, I'm a bit concerned about the cache-busting implementation:

```
2021-04-13T00:08:38.860Z info: 	Executing (default): SELECT `id` FROM `Notes` AS `Note` WHERE `Note`.`ownerId` IS NULL AND `Note`.`permission` != 'locked' AND `Note`.`createdAt` < '2021-04-13 00:08:08.857 +00:00';
2021-04-13T00:08:49.158Z info: 	Executing (default): SELECT `id`, `permission` FROM `Notes` AS `Note` WHERE `Note`.`id` = '43b124a5-d6ca-47f1-b63f-0b9c28cd678e';
2021-04-13T00:08:49.161Z debug: 	new operation: {"wrapped":[102," "],"meta":{"ranges":[{"anchor":103,"head":103}]}}
2021-04-13T00:08:49.817Z info: 	Executing (default): SELECT `id`, `permission` FROM `Notes` AS `Note` WHERE `Note`.`id` = '43b124a5-d6ca-47f1-b63f-0b9c28cd678e';
2021-04-13T00:08:49.819Z debug: 	new operation: {"wrapped":[103,"o"],"meta":{"ranges":[{"anchor":104,"head":104}]}}
```

We are talking about multiple queries per second for just one typing user 🤔 maybe adding a little bit of caching is a sane idea.

DerMolly:
> We are talking about multiple queries per second for just one typing user thinking maybe adding a little bit of caching is a sane idea.

Maybe something to consider for 2.0 then? I mean the whole feature, not only the caching…

ccoenen:
> Honestly speaking, I'm a bit concerned about the cache-busting implementation:

Is this related to this pull request?

SISheogorath:
It is related, since this PR adjust the permissions of a note and the realtime code is unaware that this can happen. As the realtime code caches the permissions of the note, it'll not notice when the permission was changed by the autolock feature and continue to apply changes for existing sessions.

lib/models/note.js
```diff
@@ -584,5 +586,34 @@ module.exports = function (sequelize, DataTypes) {
...
   }
 
+  Note.lockOldGuestNotes = async function (secondsUntilLock, permission) {
+    if (typeof secondsUntilLock !== 'number') {
+      throw new Error("Parameter 'secondsUntilLock' has to be of type 'number'")
+    }
+    if (!permissionTypes.includes(permission)) {
+      throw new Error('Unexpected permission for locking old guest notes')
+    }
+    const notes = await Note.findAll({
```
davidmehren, line 596: This and `note.save()` should probably get some kind of error handling

davidmehren (changes requested):

davidmehren:
I'm not really sure about the user experience of this feature.
If it's enabled, anonymous notes will magically become locked, without any indication why that happened. I fear great user confusion, especially as the instance admin has no way of warning the users that the feature is on.

docs/content/configuration.md
```diff
@@ -102,6 +102,14 @@ these are rarely used for various reasons.
...
 
+## Guest Notes autolock
+
+| config file           | environment                 | **default** and example value                                           | description                                                                                                                                                    |
+| --------------------- | --------------------------- | ----------------------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `lockGuestNotes` | | `{ timeout: 0,  permission: 'locked' }` | If the `timeout` setting is `>` 0, the autolock feature is activated. It'll automatically set the permission of guest notes, which have a creation date older than the timeout setting, to the permission defined in the `permission` setting. |
+| | `CMD_LOCKGUESTNOTES_TIMEOUT` | `0` | This is the maximum age in seconds a note is allowed to have before before the `CMD_LOCKGUESTNOTES_PERMISSION` is applied to it. If the timeout is set to 0, the autolock feature is disabled |
```
InnayTool, line 110:
```suggestion
| | `CMD_LOCKGUESTNOTES_TIMEOUT` | `0` | This is the maximum age in seconds a note is allowed to have before the `CMD_LOCKGUESTNOTES_PERMISSION` is applied to it. If the timeout is set to 0, the autolock feature is disabled |
```

lib/realtime.js
```diff
@@ -625,31 +625,46 @@ function updateUserData (socket, user) {
...
+      if (socket.request.user && socket.request.user.logged_in) {
+        note.lastchangeuser = socket.request.user.id
+      } else {
+        note.lastchangeuser = null
+      }
+    }
+    return callback(mayEdit)
+  }).catch(function (error) {
+    logger.error(error)
+    // false is required here in as the callback function only expects a boolean.
```
InnayTool, line 664:
```suggestion
    // false is required here as the callback function only expects a boolean.
```

lib/realtime.js
```diff
@@ -625,31 +625,46 @@ function updateUserData (socket, user) {
...
+    // Not accepting changes the best way to handle such an unknown situation is to not accept changes.
```
InnayTool, line 665:
```suggestion
    // Not accepting changes. The best way to handle such an unknown situation is to not accept changes.
```

InnayTool (changes requested):

mrdrogdrog:
> I'm not really sure about the user experience of this feature.
> If it's enabled, anonymous notes will magically become locked, without any indication why that happened. I fear great user confusion, especially as the instance admin has no way of warning the users that the feature is on.

We could add an indicator to the frontend. An icon with a tooltip that says "note will be locked at $time"

app.js
mrdrogdrog: But if I don't restart the app that often this also doesn't help. Why don't you check the database every e.g. 60 minutes?

app.js
```diff
@@ -276,6 +276,13 @@ models.sequelize.authenticate().then(function () {
...
```
mrdrogdrog, line 281: Why don't you check the database every e.g. 60 minutes? Otherwise the lock might happen waaaay to late.

davidmehren:
Okay, we need to make a decision on this feature.

From my point of view, we can merge this if the change in cache-busting behaviour is gated behind the feature flag and we clearly communicate the possible implications to performance and user experience.
Adding an indicator to the frontend would be nice to have, but adding that to the 1.x code is probably a pain.

ErikMichelson:
I honestly don't see the point of this feature. Deleting inactive notes after some time might be reasonable for some instances, but why exactly should the permissions change after some time?

mrdrogdrog:
I don't think we should merge this version.

1. The cache busting can be considered as a breaking change because some instances, that are either huge and heavily used, with lots of people who are online at the same time, or run on very low powered hardware (raspi) might have decreased performance.
2. If we merge this now, then people will expect it this feature be present in 2.0. With the current behaviour. So we can either keep a not well planned feature in 2.0 or we can drop it for 2.0, so it will be present in 1.8, not 2.0, but then again in 2.x.
3. There are some loop holes in the logic. What if I create a note as a guest that needs to stay open longer than the lockdown timer is? I can't prevent the lockdown because I can't claim the ownership with my account and I can't increase or reset the timeout.

IMHO Instead of merging this we should take our time and plan this feature through.

SISheogorath:
Well, as I said before opening this PR, it's open to us whether we actually want to implement it. It solves the issue the person in chat was asking for. For the shortcoming that come with it, it might not be worth doing. Therefore I think we should close this PR and if at any point in the future we feel we want this feature, some basic cases have already been explored.

PR #1141, opened 2021-04-16
Add WebDAV media backend

### LINKED ISSUE #1041: Add WebDAV media backend
As @ErikMichelson suggested in https://github.com/hedgedoc/hedgedoc/issues/529#issuecomment-783736443 we could include WebDAV as a media backend. In this I tried to think this through and will show my findings.

I'd suggest using https://www.npmjs.com/package/webdav. Looking at the auth methods possible with this package, it seems we could support many auth methods for WebDAV. But I'm not too sure, if that is needed or if we could just use username and password as this is the way NextCloud and ownCloud (two of the most used WebDAV servers) do auth.

We would then need the following additional configs:

```
HD_MEDIA_BACKEND_WEBDAV_SERVER
HD_MEDIA_BACKEND_WEBDAV_UPLOAD_DIR
HD_MEDIA_BACKEND_WEBDAV_PUBLIC_URL
```

`SERVER` should include the username and password in the known `schema://user:pass@url` notation.

As is shown beneath, the `PUBLIC_URL` would need to support a placeholder, that the media backend then replaces with the filename, to be usable with NextCloud. 

### NextCloud
<del>
I played around with NextCloud and found that if the folder, that the images are in, is shared, it's possible to build the link to the image the following way:

`https://<SERVER>/apps/files_sharing/publicpreview/<SHARE_ID_FOLER>?file=/<FILE_NAME>&x=3840&y=2160&a=true`

It's necessary to include a screen resolution via the `x` and `y` parameters. The `a` parameter seems to indicate if the image should be kept at its ratio or if it should be trimmed / to the ratio indicated by `x` and `y`.

This should be the relevant docs from NextCloud: <https://docs.nextcloud.com/server/21/admin_manual/configuration_files/previews_configuration.html>

</del>

As @mrdrogdrog and @ErikMichelson suggested it's possible to use the download url of NextCloud instead:

`https://<SERVER>/s/<SHARE_ID_FOLER>/download?path=%2F&files=<FILE_NAME>`

### webdav package

The necessary code from the webdav package should be somthing like this:

#### client creation
```
const client = createClient(
    "https://webdav.example.com/marie123",
    {
        username: "marie",
        password: "myS3curePa$$w0rd"
    }
);
```

#### file upload
```
await client.putFileContents("/my/file.jpg", imageBuffer, { overwrite: false });
```

#### file deletion
```
await client.deleteFile("/tmp.dat");
```

DerMolly:
### Component/Part
webdav media backend

### Description
This PR adds the webdav media backend

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
fixes #1041

### DIFF
--- a/docs/content/media/webdav.md
+++ b/docs/content/media/webdav.md
@@ -1,0 +1,46 @@
+# WebDAV
+
+You can use any [WebDAV](https://en.wikipedia.org/wiki/WebDAV) server to handle your image uploads in HedgeDoc.
+
+The WebDAV server must host the files in a way that allows HedgeDoc to request and receive them.
+
+You just add the following lines to your configuration:  
+<small>(with the appropriate substitution for `<CONNECTION_STRING>`, `<UPLOAD_DIR>`, and `<PUBLIC_URL>` of course)</small>
+```
+HD_MEDIA_BACKEND="webdav"
+HD_MEDIA_BACKEND_WEBDAV_CONNECTION_STRING="<CONNECTION_STRING>"
+HD_MEDIA_BACKEND_WEBDAV_UPLOAD_DIR="<UPLOAD_DIR>"
+HD_MEDIA_BACKEND_WEBDAV_PUBLIC_URL="<PUBLIC_URL>"
+```
+
+The `<CONNECTION_STRING>` should include the username and password (if needed) in the familiar way of `schema://user:password@url`.
+
+With `<UPLOAD_DIR>` you can specify a folder you want to upload to, but you can also omit this (just don't spcify this value at all), if you prefer to upload directly to the root of the WebDAV server.
+
+Finally, `<PUBLIC_URL>` specifies with which url HedgeDoc can access the upload. For this purpose the filename will be appended to `<PUBLIC_URL>`. So the file `test.png` with `<PUBLIC_URL>` `https://dav.example.com` should be accessible via `https://dav.example.com/test.png`.
+
+## Using Nextcloud
+
+If you want to use Nextcloud as a WebDAV server, follow the following instructions:
+
+This guide was written using Nextcloud 21 in April 2021.
+
+Because the username and app password will be included in the config, we suggest using a dedicated Nextcloud user for the uploads.  
+
+In this example the username will be `TestUser`.
+
+1. Create an app password by going to `Settings` > `Security`. Nextcloud will generate a password for you. Let's assume it's `passw0rd`.
+2. In the Files app [create a new folder](https://docs.nextcloud.com/server/latest/user_manual/en/files/access_webgui.html#creating-or-uploading-files-and-directories) that will hold your uploads (e.g `HedgeDoc`).
+3. [Share](https://docs.nextcloud.com/server/latest/user_manual/en/files/sharing.html#public-link-shares) the newly created folder. The folder should (per default) be configured with the option `Read Only` (which we will assume in this guide), but `Allow upload and editing` should be fine, too.
+4. Get the public link of the share. It should be in your clipboard after creation. If not you can copy it by clicking the clipboard icon at the end of the line of `Share link`. We'll assume it is `https://cloud.example.com/s/some-id` in the following.
+5. Append `/download?path=%2F&files=` to this URL. To continue with our example the url should now be `https://cloud.example.com/s/some-id/download?path=%2F&files=`.
+6. Get the [WebDAV url of you Nextcloud server](https://docs.nextcloud.com/server/latest/user_manual/en/files/access_webdav.html). It should be located in the Files app in the bottom left corner under `Settings` > `WebDAV`. We'll assume it is `https://cloud.example.com/remote.php/dav/files/TestUser/` in the following.
+7. Add your login information to the link. This is done by adding `username:password@` in between the url schema (typically `https://`) and the rest of the url (`cloud.example.com/remote.php/dav/files/TestUser/` in our example). The WebDAV url in our example should now look like this `https://TestUser:passw0rd@cloud.example.com/remote.php/dav/files/TestUser/`.
+8. Configure HedgeDoc: 
+```
+HD_MEDIA_BACKEND="webdav"
+HD_MEDIA_BACKEND_WEBDAV_CONNECTION_STRING="https://TestUser:passw0rd@cloud.example.com/remote.php/dav/files/TestUser/"
+HD_MEDIA_BACKEND_WEBDAV_UPLOAD_DIR="HedgeDoc"
+HD_MEDIA_BACKEND_WEBDAV_PUBLIC_URL="https://cloud.example.com/s/some-id/download?path=%2F&files="
+```
+Start using image uploads backed by Nextclouds WebDAV server.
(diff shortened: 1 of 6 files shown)


DerMolly:
The necessary config for testing can be build with the guide included in this PR.

When the server is correctly configures the upload can be tested via

```
curl --location --request POST 'http://localhost:3000/media' \                                                                                                                 
         --header 'HedgeDoc-Note: test' \
         --form 'file=@"/your/hedgedoc_repo/test/private-api/fixtures/test.png"'
```

if you first insert a user 'hardcoded' and also a note 'test' they own into the db.

docs/content/media/webdav.md
```diff
@@ -0,0 +1,45 @@
...
+
+The `<CONNECTION_STRING>` should include the username and password (if needed) in the familiar way of `schema://user:password@url`.
+
+With `<UPLOAD_DIR>` you can specify a folder you want to upload to, but you can also omit this, if you prefer to upload directly to the root of the WebDAV server.
+
+Finally, `<PUBLIC_URL>` specifies with which url HedgeDoc can access the upload. For this purpose the filename will be appended to `<PUBLIC_URL>`. So the file `test.png` with `<PUBLIC_URL>` `https://dav.example.com` should be accessible via `https://dav.example.com/test.png`.
+
+## Using NextCloud
+
+If you want to use the WebDAV server of a NextCloud server, follow the following instructions:
```
InnayTool, line 24:
```suggestion
If you want to use the NextCloud as WebDAV server, follow the following instructions:
```

docs/content/media/webdav.md
```diff
@@ -0,0 +1,45 @@
...
+2. [Share](https://docs.nextcloud.com/server/latest/user_manual/en/files/sharing.html#public-link-shares) the newly created folder. The folder should (per default) be configured with the option `Read Only` (which we will assume in this guide), but `Allow upload and editing` should be fine, too.
+3. Get the public link of the share. It should be in your clipboard after creation. If not you can copy it by clicking the clipboard icon at the end of the line of `Share link`. We'll assume it is `https://cloud.example.com/s/some-id` in the following.
+4. Append `/download?path=%2F&files=` to this URL. To continue with our example the url should now be `https://cloud.example.com/s/some-id/download?path=%2F&files=`.
+5. Get the [WebDAV url of you NextCloud server](https://docs.nextcloud.com/server/latest/user_manual/en/files/access_webdav.html). It should be located in the Files app in the bottom left corner under `Settings` > `WebDAV`. Well assume it is `https://cloud.example.com/remote.php/dav/files/TestUser/` in the following.
```
InnayTool, line 36:
```suggestion
5. Get the [WebDAV url of you NextCloud server](https://docs.nextcloud.com/server/latest/user_manual/en/files/access_webdav.html). It should be located in the Files app in the bottom left corner under `Settings` > `WebDAV`. We'll assume it is `https://cloud.example.com/remote.php/dav/files/TestUser/` in the following.
```

InnayTool (changes requested):

docs/content/media/webdav.md
```diff
@@ -0,0 +1,45 @@
...
+
+## Using NextCloud
+
+If you want to use the WebDAV server of a NextCloud server, follow the following instructions:
+
+This guide was written using NextCloud 21 in April 2021.
+
+Because the username and password will be included in the config, we suggest using a dedicated NextCloud user for the uploads.  
+
+In this example the username will be `TestUser` with the password `passw0rd`.
```
SuperSandro2000, line 30: Instead of a password you should probably create an app password. You Can find that in your account settings > security.

docs/content/media/webdav.md
DerMolly: done

docs/content/media/webdav.md
DerMolly: done

docs/content/media/webdav.md
DerMolly: done

docs/content/media/webdav.md
```diff
@@ -0,0 +1,46 @@
...
+```
+HD_MEDIA_BACKEND="webdav"
+HD_MEDIA_BACKEND_WEBDAV_CONNECTION_STRING="<CONNECTION_STRING>"
+HD_MEDIA_BACKEND_WEBDAV_UPLOAD_DIR="<UPLOAD_DIR>"
+HD_MEDIA_BACKEND_WEBDAV_PUBLIC_URL="<PUBLIC_URL>"
+```
+
+The `<CONNECTION_STRING>` should include the username and password (if needed) in the familiar way of `schema://user:password@url`.
+
+With `<UPLOAD_DIR>` you can specify a folder you want to upload to, but you can also omit this, if you prefer to upload directly to the root of the WebDAV server.
```
davidmehren, line 18: Omitting means setting `HD_MEDIA_BACKEND_WEBDAV_UPLOAD_DIR` so an empty string of not setting it?

docs/content/media/webdav.md
```diff
@@ -0,0 +1,46 @@
...
+
+Finally, `<PUBLIC_URL>` specifies with which url HedgeDoc can access the upload. For this purpose the filename will be appended to `<PUBLIC_URL>`. So the file `test.png` with `<PUBLIC_URL>` `https://dav.example.com` should be accessible via `https://dav.example.com/test.png`.
+
+## Using NextCloud
```
davidmehren, line 22: The 'c' in 'Nextcloud' is *not* capitalized.

src/config/media.config.ts
```diff
@@ -70,6 +75,19 @@ const mediaSchema = Joi.object({
       }),
       otherwise: Joi.optional(),
     }),
+    webdav: Joi.when('use', {
+      is: Joi.valid(BackendType.WEBDAV),
+      then: Joi.object({
+        connectionString: Joi.string().label(
```
davidmehren, line 81: This and `publicUrl` should both be HTTP(S) URLs, right?
We should probably validate them as such.

package.json
```diff
@@ -57,7 +57,8 @@
     "shortid": "2.2.16",
     "sqlite3": "5.0.2",
     "swagger-ui-express": "4.1.6",
-    "typeorm": "0.2.32"
+    "typeorm": "0.2.32",
+    "webdav": "4.3.0"
```
davidmehren, line 61: We should be able to do the PUT and DELETE requests ourselves instead of using the external library.

davidmehren (changes requested):

docs/content/media/webdav.md
DerMolly: fixed

docs/content/media/webdav.md
DerMolly: clarified

package.json
DerMolly: done

src/config/media.config.ts
DerMolly: added

InnayTool (approved):

davidmehren (approved):

PR #1150, opened 2021-04-18
Add Prometheus endpoint

davidmehren:
### Component/Part
Metrics

### Description
This PR adds a `/metrics` Prometheus endpoint using https://github.com/PayU/prometheus-api-metrics.
The custom stats currently available at `/status` are also available as gauges.

I quickly build this for my internal monitoring, but thought it might be nice to have upstream as well.
I never created any Prometheus metrics before, so feedback on the naming/structure is welcome!

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/app.js
+++ b/app.js
@@ -18,6 +18,7 @@ const passportSocketIo = require('passport.socketio')
 const helmet = require('helmet')
 const i18n = require('i18n')
 const flash = require('connect-flash')
+const apiMetrics = require('prometheus-api-metrics')
 
 // core
 const config = require('./lib/config')
@@ -62,6 +63,9 @@ if (!config.useSSL && config.protocolUseSSL) {
 app.use(morgan('combined', {
   stream: logger.stream
 }))
+
+// Register prometheus metrics endpoint
+app.use(apiMetrics())
 
 // socket io
 const io = require('socket.io')(server, { cookie: false })
--- a/app.js
+++ b/app.js
@@ -26,6 +26,7 @@ const logger = require('./lib/logger')
 const errors = require('./lib/errors')
 const models = require('./lib/models')
 const csp = require('./lib/csp')
+const metrics = require('./lib/prometheus')
 
 // server setup
 const app = express()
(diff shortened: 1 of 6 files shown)


SISheogorath:
I think it would be useful to move the Prometheus client definition out if the utils Funktion as I don't think it's useful to always have realtime in that context.

lib/utils.js
```diff
@@ -25,3 +27,47 @@ exports.getImageMimeType = function getImageMimeType (imagePath) {
       return undefined
   }
 }
+
+exports.setupCustomPrometheusMetrics = function () {
```
DerMolly, line 31: I'd suggest moving this in its own prometheus.ts file

InnayTool (dismissed):

davidmehren:
The function `setupCustomPrometheusMetrics` has been moved to its own `metrics.js` file.

davidmehren:
As suggested by @ErikMichelson, the new endpoint is now documented.

DerMolly (approved):

InnayTool (dismissed):

lib/metrics.js
```diff
@@ -0,0 +1,46 @@
...
+    name: 'hedgedoc_online_sessions',
+    help: 'Sessions currently editing notes',
+    labelNames: ['type']
+  })
+  const onlineUsers = new promClient.Gauge({
+    name: 'hedgedoc_online_users',
+    help: 'Distinct users currently editing notes',
+    labelNames: ['type']
+  })
+  const notesCount = new promClient.Gauge({ name: 'hedgedoc_notes', help: 'Notes in the instance' })
```
ErikMichelson, line 19: Any reason why this Gauge's config is inlined and not like the other ones split into multiple lines?

davidmehren: I think the line is short enough to not make eslint unhappy

ErikMichelson (dismissed):

InnayTool (dismissed):

DerMolly (dismissed):

DerMolly (approved):

InnayTool (dismissed):

SISheogorath:
The library file should either be called `prometheus.js` or also contain the code from the `/status` endpoint. I don't think it makes a lot of sense to spread this code around like this when using more universal names.

davidmehren:
I opted for renaming to `prometheus.js`, moving the `/status` code didn't seem to make much sense, as it's mostly an Express router.

DerMolly (approved):

SISheogorath (approved):

PR #1173, opened 2021-04-23
NoteEntity: Add version field

DerMolly:
### Component/Part
note entity

### Description
This PR adds a version field to note entity

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:


src/notes/note-metadata.dto.ts
```diff
@@ -33,6 +34,12 @@ export class NoteMetadataDto {
   @ApiPropertyOptional()
   alias: string;
 
+  /**
+   * If the note is note from before version 2 was a thing
+   */
+  @IsBoolean()
+  preVersionTwoNote: boolean;
```
ErikMichelson, line 41: Why do we have a boolean for this if we've got the actual version?
Wouldn't it make more sense to adjust the frontend implementation on this?

DerMolly: That could also be done.

davidmehren:
What problem does this solve?

mrdrogdrog:
This field can be used to track if a note (and their content) was already migrated.

mrdrogdrog:
If you have lots of notes then the content migration could take a while. This field can make sure that if (because of whatever reason) the server got interrupted it won't try to migrate an already migrated note.

DerMolly:
It's needed for the frontend to decide if a url should redirect to the correct note or if an error page should be displayed.

DerMolly:
@mrdrogdrog that is a neat idea, but this field stays on 1 for all old notes and as nothing to do with migrations what so ever

mrdrogdrog:
Then we don't talk about the same feature.

DerMolly:
> Then we don't talk about the same feature.

Probably

mrdrogdrog:
If the only reason for this flag is to determine if a direct linking should be allowed then you it should be a Boolean called "directLinkAllowed".
Variables should always reveal their intend.

DerMolly:
> If the only reason for this flag is to determine if a direct link should be allowed then you it should be a Boolean called "directLinkAllowed" or "accessWithoutNoteNamespaceAllowed".
> Variables should always reveal their intend. Especially if they have only one purpose.

I guess that's a reasonable idea. I think `directLinkAllowed` is the way to go, what does the rest think?

ErikMichelson:
I disagree with that.

The direct link feature is just one thing that changed between notes created in 1.x and 2.0. But when for whatever reason the structure of notes needs to change again (e.g. with a version 3), we'll need another attribute. Having a single version fields is way more useful especially when we're at the point that specific features or warning messages should only be available/displayed to notes of "generation 1" or maybe some when in the future "generation 2".

ErikMichelson:
But when re-reading the changes, maybe the better name would be something like `createdWithVersion`?

DerMolly:
My point of view is this:
All HedgeDoc 1 notes will be migrated by an extern script, because the DB schemas are very different and the ORM changed too. So I think a HedgeDoc 1 note should be no different than a HegdeDoc 2 note, with the exception that some people have weird links to it, which we need to handle correctly.

davidmehren:
A note version might actually be a sensible thing to have. I'd prefer a monotonically increasing integer `version` to something like `createdWithVersion` as that could lead to us having to parse a semver string.

In any case, we should document
- changes in the note content
- changes in backend/API behavior, if any
- changes in frontend behavior
that each version entails.

We already talk about the `version` field in https://github.com/hedgedoc/hedgedoc/blob/develop/docs/content/dev/notes.md, but skip most of the things mentioned above.

DerMolly:
@davidmehren I'm not sure I understand what you suggest here. If you wanted changes between a version 1 and version 2 note to be documented, that's already fully done.
Or do you want to use that field for something else, then please tell us.

mrdrogdrog:
"createdWithVersion" would be okay.
I have understood it to be a increasing integer. So it's rather "created as note version n" than "created with HedgeDoc x.y.z"

InnayTool (approved):

davidmehren:
I think we need to need work out what this number is going to be used for.
Is it for "content migration"?
Is it for different behavior of the frontend (like the note-subpath mentioned)?
Are these things that can only change with major releases?

This all influences if the field should reflect a HedgeDoc version or if it should be a independent version.

If we are in agreement on that, this PR should include an update to https://github.com/hedgedoc/hedgedoc/blob/develop/docs/content/dev/notes.md#L51 which documents the intended use of the field and the specifics of what each version means. I want to avoid having a sparely documented magic attribute that changes the behaviour of the application at a dozen places.

DerMolly:
We (@davidmehren, @ErikMichelson, @mrdrogdrog and myself) just discussed this issue/PR and decided it's probably best to not add a version field, but instead add a simple boolean flag, as we don't actually want to track the version, but just if this one feature should be active. It's more intuitive and easier to comprehend this way. Therefore I'll close this issue…

PR #1195, opened 2021-04-25
[1.x] Add unified supported locales list

ErikMichelson:
### Component/Part
Language picker, i18n

### Description
When adding new languages to HedgeDoc, you had to change three different places in the app. This PR simplifies that by having a single JSON file that defines the available languages.

### Steps
- [x] Added implementation
  made sure that:


DerMolly (approved):
I tried that and it works. The only thing that changes in contrast to the previous version is that, while switching the language the language picker is empty for a short period of time.

In my opinion nothing too problematic.

ErikMichelson:
>  the language the language picker is empty for a short period of time.

Yes, that's due to the picker being now dynamically generated by the JS and not being hardcoded anymore.

mrdrogdrog:
Can't you include the json during buildtime?

mrdrogdrog:
Aka: Import the json? Like we do in the frontend?

ErikMichelson:
That's exactly what I'm doing in this PR.

mrdrogdrog:
Got me. I haven't read the code yet 😋

mrdrogdrog:
Then why don't you generate it in the template?

PR #1207, opened 2021-04-29
Enable TypeScript strict mode

davidmehren:
### Component/Part
everything™

### Description
This PR enables TypeScripts strict mode and performs all adjustments necessary for the project to still compile.

### Steps

- [x] Added implementation
- [ ] Added / updated tests
  made sure that:

### DIFF
--- a/src/auth/auth-token.dto.ts
+++ b/src/auth/auth-token.dto.ts
@@ -15,8 +15,8 @@ export class AuthTokenDto {
   createdAt: Date;
   @IsDate()
   @IsOptional()
-  validUntil: Date;
+  validUntil: Date | null;
   @IsDate()
   @IsOptional()
-  lastUsed: Date;
+  lastUsed: Date | null;
 }
--- a/src/config/utils.ts
+++ b/src/config/utils.ts
@@ -6,7 +6,7 @@
 
 import { Loglevel } from './loglevel.enum';
 
-export function toArrayConfig(configValue: string, separator = ','): string[] {
+export function toArrayConfig(configValue?: string, separator = ','): string[] {
   if (!configValue) {
     return [];
   }
--- a/src/config/utils.ts
+++ b/src/config/utils.ts
@@ -113,3 +113,10 @@ function transformLoglevelToInt(loglevel: Loglevel): number {
       return 1;
   }
 }
+
+export function parseOptionalInt(value?: string): number | undefined {
+  if (value === undefined) {
+    return undefined;
+  }
+  return parseInt(value);
+}
--- a/src/notes/note.entity.ts
+++ b/src/notes/note.entity.ts
@@ -36,7 +36,7 @@ export class Note {
     unique: true,
     nullable: true,
   })
-  alias?: string;
+  alias: string | null;
   @OneToMany(
     (_) => NoteGroupPermission,
     (groupPermission) => groupPermission.note,
@@ -70,11 +70,11 @@ export class Note {
   @Column({
     nullable: true,
   })
-  description?: string;
+  description: string | null;
   @Column({
     nullable: true,
   })
-  title?: string;
+  title: string | null;
 
   @ManyToMany((_) => Tag, (tag) => tag.notes, { eager: true, cascade: true })
   @JoinTable()
--- a/src/notes/note.entity.ts
+++ b/src/notes/note.entity.ts
@@ -56,8 +56,9 @@ export class Note {
   viewCount: number;
   @ManyToOne((_) => User, (user) => user.ownedNotes, {
     onDelete: 'CASCADE', // This deletes the Note, when the associated User is deleted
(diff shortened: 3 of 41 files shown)


davidmehren:
Note: Tests are still *excluded* from strict checks for now, as the type errors in Jest mocks needed deeper investigation.

InnayTool (approved):

davidmehren:
There still are some inconsistencies, where properties are `|null` in an entity, but `|undefined` in the corresponding DTO. I'll fix them tomorrow.

src/auth/auth-token.dto.ts
```diff
@@ -15,8 +15,8 @@ export class AuthTokenDto {
   createdAt: Date;
   @IsDate()
   @IsOptional()
-  validUntil: Date;
+  validUntil: Date | null;
```
DerMolly, line 18: Looking at the logic of the AuthToken creation. This shouldn't be optional, because either the user requested a validUntil Date or it's good for two years…

src/config/utils.ts
```diff
@@ -113,3 +113,12 @@ function transformLoglevelToInt(loglevel: Loglevel): number {
       return 1;
   }
 }
+
+export function parseOptionalInt(
+  value: string | undefined,
```
DerMolly, line 118:
```suggestion
  value?: string,
```
Maybe use this to be consistent?

src/config/utils.ts
```diff
@@ -113,3 +113,12 @@ function transformLoglevelToInt(loglevel: Loglevel): number {
       return 1;
   }
 }
+
+export function parseOptionalInt(
```
DerMolly, line 117: This function should be covered by a unit test the same way `toArrayConfig` is…

src/notes/note.entity.ts
```diff
@@ -89,9 +93,9 @@ export class Note {
     }
     const newNote = new Note();
     newNote.shortid = shortid;
```
DerMolly, line 95: Shouldn't this need a default, too?

src/seed.ts
```diff
@@ -56,7 +56,16 @@ createConnection({
...
     const foundUser = await connection.manager.findOne(User);
+    if (!foundUser) {
+      throw new Error('Could not find freshly seeded user. Aborting.');
+    }
     const foundNote = await connection.manager.findOne(Note);
+    if (!foundNote) {
+      throw new Error('Could not find freshly seeded note. Aborting.');
+    }
+    if (!foundNote.alias) {
+      throw new Error('Could not find alias of freshly seeded note. Aborting.');
```
DerMolly, line 67: These should be combined

DerMolly (changes requested):
Some small problems I found.

DerMolly:
We probably should include extra tests to cover the new `throw Error` lines.

Also I noticed we don't log anything if we encounter an NotInDBError or similiar, in other places we do that.

src/auth/auth-token.dto.ts
davidmehren: I'd like to keep changes in this PR focused on adjusting the types to the current database schema. Can we push this change to another PR?

src/notes/note.entity.ts
davidmehren: I don't follow, it has one, see https://github.com/hedgedoc/hedgedoc/blob/maintenance/ts_strict_mode/src/notes/note.entity.ts#L91-L93

src/seed.ts
davidmehren: They intentionally are not combined, as they are different errors.
I could merge all seed-errors into one, but then the error message will be less helpful...

src/config/utils.ts
davidmehren: fixed

src/config/utils.ts
davidmehren: fixed

davidmehren:
The new `NotInDBError`s have tests now, the coverage still decreases because of all the `req.user` tests that cannot be reached.

src/revisions/revisions.service.spec.ts
```diff
@@ -5,7 +5,9 @@
  */
 
 import { Test, TestingModule } from '@nestjs/testing';
-import { getRepositoryToken } from '@nestjs/typeorm';
+import { getConnectionToken, getRepositoryToken } from '@nestjs/typeorm';
```
DerMolly, line 8: `yarn lint` informs me that `getConnectionToken` is not used here…

DerMolly (changes requested):

InnayTool (approved):

src/notes/note.entity.ts
DerMolly: my bad

DerMolly (approved):

PR #1223, opened 2021-05-04
Automatically enable protocolUseSSL when useSSL is set

davidmehren:
### Component/Part
config

### Description
 This makes the behavior consistent with the docs and
 saves the user from having to both set
 `useSSL` and `protocolUseSSL`.

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -1,4 +1,3 @@
-
 'use strict'
 
 const crypto = require('crypto')
@@ -31,7 +30,7 @@ const packageConfig = {
 }
 
 const configFilePath = path.resolve(appRootPath, process.env.CMD_CONFIG_FILE ||
-'config.json')
+  'config.json')
 const fileConfig = fs.existsSync(configFilePath) ? require(configFilePath)[env] : undefined
 
 let config = require('./default')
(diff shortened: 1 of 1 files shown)


lib/config/index.js
```diff
@@ -88,6 +88,11 @@ config.isStandardHTTPPort = (function isStandardHTTPPort () {
   return !config.useSSL && config.port === 80
 })()
 
+// Use HTTPS protocol if the internal TLS server is enabled
+if (config.useSSL === true) {
+  config.protocolUseSSL = true
```
ErikMichelson, line 93: As I wrote in the matrix chat, do we want to raise a warning if `protocolUseSSL` is explicitly set to false?

Would be something like this:

```suggestion
  if (config.protocolUseSSL === false) {
    logger.warn('You configured protocolUseSSL to false, but this will be overriden as you set useSSL to true.')
  }
  config.protocolUseSSL = true
```

davidmehren: I added a warning

DerMolly (dismissed):

InnayTool (approved):

DerMolly (approved):

PR #1227, opened 2021-05-05
Use esbuild to minify frontend JS

davidmehren:
### Component/Part
Webpack config

### Description
This PR switches the minifier in Webpack from https://github.com/terser/terser to https://esbuild.github.io. esbuild is written in Go and reduces the build time from `54s wall/198s user` to `12s wall/21s user` on my machine. 

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -2,8 +2,7 @@
   "presets": [
     ["env", {
       "targets": {
-        "node": "8",
-        "uglify": true
+        "browsers": "defaults"
       }
     }]
   ],
--- a/package.json
+++ b/package.json
@@ -151,6 +151,7 @@
     "copy-webpack-plugin": "6.4.1",
     "css-loader": "5.2.4",
     "emojify.js": "^1.1.0",
+    "esbuild-loader": "2.13.0",
     "escape-html": "^1.0.3",
     "eslint": "7.25.0",
     "eslint-config-standard": "16.0.2",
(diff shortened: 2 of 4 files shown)


.babelrc
```diff
@@ -2,8 +2,7 @@
   "presets": [
     ["env", {
       "targets": {
-        "node": "8",
-        "uglify": true
+        "node": "8"
```
ErikMichelson, line 5: Why do we target node 8, if we raised the minimum node version to 12?

davidmehren: I have no idea. It doesn't even make sense to target Node, as Babel only runs on code for the frontend.

mrdrogdrog (approved):

.babelrc
davidmehren: I changed the Babel settings to `browsers: defaults`, which equals `> 0.5%, last 2 versions, Firefox ESR, not dead`, which seems more reasonable.
Not sure if that had any effect on bundle size...

InnayTool (approved):

package.json
```diff
@@ -179,6 +179,7 @@
     "babel-runtime": "6.26.0",
     "copy-webpack-plugin": "6.4.1",
     "css-loader": "5.2.4",
+    "esbuild-loader": "^2.12.0",
```
DerMolly, line 182: This should be pinned…

DerMolly (changes requested):

mrdrogdrog (dismissed):

ErikMichelson (approved):

package.json
davidmehren: Pinned!

mrdrogdrog (approved):

DerMolly (approved):

InnayTool (approved):

PR #1228, opened 2021-05-05
[PoC] Add a Dockerfile

davidmehren:
### Component/Part
Dockerfile

### Description
This PR contains a proof-of-concept Dockerfile, written from scratch.
It is not necessarily intended to be merged here, we could also integrate some ideas into https://github.com/hedgedoc/container.

I intentionally don't support legacy environment variables or 'magic' variable rewrites as the entrypoint in the current container does.

Instead, this PR moves the logic to wait for the database on startup to the application and therefore does not add a dependency to dockerize.

It also adds a simple healthcheck, built natively with Node.

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+FROM node:lts as builder
+WORKDIR /home/node/app
+
+# Install dependencies first to not invalidate the cache on every source change
+COPY package.json yarn.lock ./
+RUN yarn install --production=false --pure-lockfile
+
+# Copy stuff needed to build the frontend
+COPY locales locales
+COPY public public
+COPY webpack.*.js .babelrc ./
+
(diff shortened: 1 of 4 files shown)


Dockerfile
```diff
@@ -0,0 +1,38 @@
...
+
+# Production container
+FROM base
+
+# Setup various metadata
+WORKDIR /home/node/app
+USER node
+ENV NODE_ENV=production
+EXPOSE 3000
+HEALTHCHECK CMD node healthcheck.mjs
```
SuperSandro2000, line 30: Please don't enable a healthcheck by default. It usually just delays the startup of the container by 10s of seconds and the availability in proxies like Traefik.

PR #1269, opened 2021-05-11
Refactor(note): Remove shortid from note

### LINKED ISSUE #1268: Remove shortid column from note entity
The `shortid` is a relic from 1.x.
This alternative ID will be moved to an alias in 2.0.

DerMolly:
### Component/Part
note

### Description
This PR removes the shortid form the note.

This is a 1.x relict and isn't used anymore in 2.x. All 1.x shortids will be converted to aliases on migration of 1.x to 2.x.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
fixes #1268

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,7 +57,6 @@
     "reflect-metadata": "0.1.13",
     "rimraf": "3.0.2",
     "rxjs": "7.0.0",
-    "shortid": "2.2.16",
     "sqlite3": "5.0.2",
     "swagger-ui-express": "4.1.6",
     "typeorm": "0.2.32"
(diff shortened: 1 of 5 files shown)


InnayTool (approved):

davidmehren (approved):

package.json
```diff
@@ -57,7 +57,6 @@
     "reflect-metadata": "0.1.13",
     "rimraf": "3.0.2",
     "rxjs": "7.0.0",
-    "shortid": "2.2.16",
```
ErikMichelson, line 60: Good thing that we dropped this.

> shortid is deprecated, because the architecture is unsafe. we instead recommend Nano ID, which has the advantage of also being significantly faster than shortid

PR #1284, opened 2021-05-16
feat(note): Add publicId field 

DerMolly:
Signed-off-by: Philip Molares <philip.molares@udo.edu>### Component/Part
notes

### Description
This PR adds the publicId field to the note entity, uses it in the service and adjusts the tests accordingly.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### DIFF
--- a/docs/content/dev/db-schema.plantuml
+++ b/docs/content/dev/db-schema.plantuml
@@ -5,6 +5,7 @@ skinparam nodesep 60
 entity "note" {
   *id : uuid <<generated>>
   --
+  publicId: text
   alias : text
   *viewCount : number
   *ownerId : uuid <<FK user>>
--- a/package.json
+++ b/package.json
@@ -39,6 +39,7 @@
     "@types/minio": "7.0.7",
     "@types/node-fetch": "2.5.10",
     "@types/passport-http-bearer": "1.0.36",
+    "base32-encode": "1.2.0",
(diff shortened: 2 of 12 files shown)


src/utils/randomString.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Generate a random string of defined length in bytes.
+ * @param {number} length - the length in bytes.
+ */
+export function randomString(length: number): Buffer {
+  if (length <= 0) {
+    throw new Error('randomString cannot have a length < 1');
+  }
+  return randomBytes(length);
```
SISheogorath, line 16: Uhm, is this function really needed? All it does is checking that the expected string length to be bigger than 0. Is there any place where we let users control this length parameter? On my review I was only able to spot hardcoded settings. So the if clause is never reached and the function boils down to a fassade.

DerMolly: > Is there any place where we let users control this length parameter?

I'm leaning towards no, so I think this can be removed…

InnayTool (approved):

src/utils/randomString.ts
DerMolly: done

InnayTool (approved):

davidmehren (approved):

Issue #101, opened 2019-06-04
Free-URL-Mode: alias can mask the ID of a Note.


ccoenen:
I ran into a weird situation just now. I was not able to figure out the source of the problem, but here is the symptom:

Something or someone created an article with the Note ID but as an alias. This made accessing the actual document impossible. An example:

The real document: https://example.com/KVlJ9-dBQWKrMYWnq3Z1BA
Somehow another document with this url (alias/free url!): https://example.com/kvlj9-dbqwkrmywnq3z1ba

Whenever you would follow the first link, this code here would find an alias (because this match apparently is case-insensitive?) ...

https://github.com/codimd/server/blob/6462968e84e8d92292dd23764a9e558d7800147d/lib/models/note.js#L143-L151

... and then would redirect you to the "correct" spelling of the faulty alias.

https://github.com/codimd/server/blob/6462968e84e8d92292dd23764a9e558d7800147d/lib/response.js#L184

ErikMichelson:
Maybe this could be fixed in 1.x with the next fix release?

davidmehren:
Hmm, as far as I understand, the `where` in `findOne` is case-sensitive, so we should not find the alias `kvlj9-dbqwkrmywnq3z1ba` for the note id `KVlJ9-dBQWKrMYWnq3Z1BA`.

While testing, I noticed that browsers like to autocomplete URLs. In my case, I created a standard note (with a random id using upper and lower-case letters) first and then another note with the same id but all lower-case. Firefox then liked to complete the lowercase alias with the mixed-case id.

PR #1359, opened 2021-06-06
add list of aliases to note entity

DerMolly:
### Component/Part
note

### Description
This PR adds a list of aliases to the note.

One of the aliases can be primary for each note, but all can be used to get information from the apis.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:


src/notes/alias.entity.ts
```diff
@@ -0,0 +1,56 @@
...
+    unique: true,
+  })
+  name: string;
+
+  /**
+   * Is this alias the primary alias, by which people access the note?
+   */
+  @Column({
+    comment:
+      'This field telly you if this is the primary alias of the note. If this field is null, that means false.',
```
InnayTool, line 36: Why should this field be nullable?

InnayTool (changes requested):

src/api/private/notes/notes.controller.ts
```diff
@@ -174,6 +178,84 @@ export class NotesController {
     }
   }
 
+  @Post(':noteIdOrAlias/metadata/alias/:alias')
+  async addAlias(
+    @Req() req: Request,
+    @Param('noteIdOrAlias') noteIdOrAlias: string,
+    @Param('alias') alias: string,
```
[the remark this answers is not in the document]
DerMolly: I have no strong feelings either way, so I'll change that.

src/api/private/notes/notes.controller.ts
```diff
@@ -174,6 +178,84 @@ export class NotesController {
...
+      }
+      throw e;
+    }
+  }
+
+  @Put(':noteIdOrAlias/metadata/alias/:alias')
+  async makeAliasPrimary(
+    @Req() req: Request,
+    @Param('noteIdOrAlias') noteIdOrAlias: string,
+    @Param('alias') alias: string,
```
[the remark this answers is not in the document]
DerMolly: Fair enough.

src/notes/alias.entity.ts
DerMolly: Because of this line:

https://github.com/hedgedoc/hedgedoc/blob/4be02939b3525c1c43c80f2f999558e613841f7a/src/notes/alias.entity.ts#L17

If a non-primary alias would be saved with `primary: false,` you could only every have one non-primary and one primary alias. But a nullable field does not have such problems. It's really a bit ugly, but this way the DB keeps track one note really only has one primary alias and I think this worth the slightly uglier code. I'll try to add a better explanation in the code. Maybe not in the DB comment though…

src/api/private/notes/notes.controller.ts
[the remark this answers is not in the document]
DerMolly: Or not. This request while possible in theory is a bit problematic:

The path parameter noteIdOrAlias will be there anyway, so removing is only a cosmetic change. But you would also need to put the handling of incorrect aliases into the controller instead of the service, something I'm uncomfortable
with. I think it's best to keep it like it is…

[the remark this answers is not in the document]
InnayTool: What would it change to put `noteIdOrAlias` as parameter of `getNoteByIdOrAlias` and `makeAliasPrimary`? I think incorrect aliases will be found by `getNoteByIdOrAlias`.

src/notes/alias.entity.ts
InnayTool: Yes, I understand it's a bit hacky but should be ok.

src/api/private/notes/notes.controller.ts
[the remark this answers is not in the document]
DerMolly: Yes but this way I would allow semantically problematic requests. Any note identifier including nonexistent and ones for other notes could be include in the url to make an alias primary or delete it altogether. I don't think we want that.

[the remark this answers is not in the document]
DerMolly: I misunderstood what you wanted and I implemented your request changing the primary alias is now done via `PUT :newPrimaryAlias/metadata/alias`

src/api/private/notes/notes.controller.ts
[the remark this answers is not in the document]
InnayTool: I didn't explain enough. I meant to change the parameter name from `alias` to `newAlias`.
I think `addAlias` is absolutely fine as methode name.

src/api/private/notes/notes.controller.ts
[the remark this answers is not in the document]
InnayTool: Yeah it's maybe also better to rename `:noteIdOrAlias` to `:alias` in the route because it could be confusing because noteIds can't be the primary alias.

src/api/private/notes/notes.controller.ts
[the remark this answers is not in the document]
DerMolly: ah, well that can be done, too.

src/api/private/notes/notes.controller.ts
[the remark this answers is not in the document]
DerMolly: changed

src/api/private/notes/notes.controller.ts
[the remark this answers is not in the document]
DerMolly: changed

InnayTool (approved):

src/history/history.service.ts
```diff
@@ -206,7 +202,10 @@ export class HistoryService {
    */
   toHistoryEntryDto(entry: HistoryEntry): HistoryEntryDto {
     return {
-      identifier: entry.note.alias ? entry.note.alias : entry.note.id,
+      identifier:
+        !entry.note.aliases || entry.note.aliases.length === 0
+          ? entry.note.publicId
+          : entry.note.aliases.filter((alias) => alias.primary)[0].name,
```
davidmehren, line 208: I wonder if we should extract this bit into a `getPrimaryAlias` method, to make clearer what happens.

src/notes/notes.service.ts
```diff
@@ -158,24 +166,29 @@ export class NotesService {
...
-        'groupPermissions',
-        'groupPermissions.group',
-        'userPermissions',
-        'userPermissions.user',
-        'tags',
-      ],
-    });
+
+    const note = await this.noteRepository
+      .createQueryBuilder('note')
```
davidmehren, line 171: This query wants comments ;)

src/notes/notes.service.ts
```diff
@@ -333,6 +346,104 @@ export class NotesService {
...
+      (anAlias) =>
+        anAlias.name !== alias || (anAlias.name === alias && anAlias.primary),
+    );
+    if (note.aliases.length === filteredAliases.length) {
+      this.logger.debug(
+        `The alias '${alias}' is not used by this note or is the primary alias, that can't be removed.`,
+        'removeAlias',
+      );
+      throw new NotInDBError(
+        `The alias '${alias}' is not used by this note or is the primary alias, that can't be removed.`,
```
davidmehren, line 440: `alias` being primary is a different error case from it not existing at all and should not be included in a `NotInDBError`

src/api/private/notes/notes.controller.ts
```diff
@@ -174,6 +178,82 @@ export class NotesController {
     }
   }
 
+  @Post(':noteIdOrAlias/metadata/alias/:newAlias')
```
davidmehren, line 181: I don't like the inconsistency in paths. At first, the relevant alias is at the end of the path, but for PUT and DELETE it is in the middle. For consistency, the `alias` that is operated on should always be at the end of the path.

src/api/private/notes/notes.controller.ts
```diff
@@ -174,6 +178,82 @@ export class NotesController {
...
+  async addAlias(
+    @Req() req: Request,
+    @Param('noteIdOrAlias') noteIdOrAlias: string,
+    @Param('newAlias') newAlias: string,
+  ): Promise<NoteMetadataDto> {
```
davidmehren, line 186: It's also very weird, that one operates (according to the path) on alias objects, but the response contains a whole metadata object...

davidmehren (changes requested):
Please also fix the typos in the commit messages ;)

src/api/private/notes/notes.controller.ts
davidmehren: Unfortunately, this whole API is quite ugly.
- Normally, POST should operate on the collection directly, parameters (like the alias name) should be in the request body, not in the path.
- We are in the `/notes` namespace. At all other places, the `noteIdOrAlias` parameter in the path is used to find a *note*. If a note has 3 aliases, you can use them interchangeably and always operate in the same object. The PUT and DELETE operations here break this expectation, as they now use the parameter to identify an *alias*.

I think we should move the alias operations to a new `/aliases` route, to not create confusion here and have a cleaner API.

src/notes/notes.service.ts
```diff
@@ -333,6 +346,104 @@ export class NotesService {
...
   }
 
+  /**
+   * @async
+   * Add the specified alias to the note.
+   * @param {Note} note - the note to add the alias to
+   * @param {string} alias - the alias to add to the note
+   * @throws {AlreadyInDBError} the alias is already in use.
+   */
+  async addAlias(note: Note, alias: string): Promise<Note> {
```
davidmehren, line 356: This should check if the new alias is already used as a noteId somewhere, to prevent https://github.com/hedgedoc/hedgedoc/issues/101

src/notes/note-metadata.dto.ts
```diff
@@ -25,13 +25,17 @@ export class NoteMetadataDto {
   id: string;
 
   /**
-   * Alias of the note
-   * Can be null
+   * Aliases of the note
+   */
+  @IsArray()
+  @IsString({ each: true })
+  aliases: string[];
```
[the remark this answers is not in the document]
DerMolly: Yes. I clarified that in the comment

src/history/history.service.ts
DerMolly: extracted.

src/notes/notes.service.ts
DerMolly: commented

src/notes/notes.service.ts
DerMolly: fixed

src/api/private/notes/notes.controller.ts
DerMolly: There can't be done much, at least not if we don't want't to split the NoteMetadataDto further down. Which I'm reluctant to do…

src/notes/notes.service.ts
```diff
@@ -79,7 +88,7 @@ export class NotesService {
    */
   async createNote(
     noteContent: string,
-    alias?: NoteMetadataDto['alias'],
+    alias?: string,
     owner?: User,
   ): Promise<Note> {
     const newNote = Note.create();
```
InnayTool, line 94: The `create` function of Note also has the alias and and the owner parameter. This should be used here I think.

InnayTool (changes requested):

InnayTool (approved):

src/notes/utils.ts
```diff
@@ -14,3 +17,11 @@ export function generatePublicId(): string {
   const randomId = randomBytes(16);
   return base32Encode(randomId, 'Crockford').toLowerCase();
 }
+
+/**
+ * Extract the primary alias from a aliases of a note
+ * @param {Note} note - the note from which the primary alias should be extracted
+ */
+export function getPrimaryAlias(note: Note): string {
+  return note.aliases.filter((alias: Alias) => alias.primary)[0].name;
```
davidmehren, line 26: What happens if there is no primary alias?

src/history/history.service.ts
```diff
@@ -150,23 +151,19 @@ export class HistoryService {
...
-          where: [
-            {
-              id: historyEntry.note,
-            },
-            {
-              alias: historyEntry.note,
-            },
-          ],
-        });
+        const note = await manager
```
davidmehren, line 161: We should probably call `getNoteByIdOrAlias` instead of replicating the query here

src/history/history.service.ts
```diff
@@ -191,7 +188,10 @@ export class HistoryService {
    */
   toHistoryEntryDto(entry: HistoryEntry): HistoryEntryDto {
     return {
-      identifier: entry.note.alias ? entry.note.alias : entry.note.id,
+      identifier:
```
davidmehren, line 191: Phew, a ternary operator over three lines, I feel a separate function is warranted

test/private-api/alias.e2e-spec.ts
```diff
@@ -0,0 +1,205 @@
...
+    const newAlias = 'normalAlias3';
+    beforeAll(async () => {
+      const note = await notesService.createNote(content, testAlias, user);
+      await aliasService.addAlias(note, newAlias);
+    });
+
+    it('works with normal alias', async () => {
+      await request(app.getHttpServer())
+        .delete(`/alias/${newAlias}`)
+        .expect(204);
```
davidmehren, line 179: We should also check that the alias was actually deleted

test/private-api/alias.e2e-spec.ts
```diff
@@ -0,0 +1,205 @@
...
+
+    it('works with normal alias', async () => {
+      const metadata = await request(app.getHttpServer())
+        .put(`/alias/${newAlias}`)
+        .set('Content-Type', 'application/json')
+        .send(changeAliasDto)
+        .expect(200);
+      expect(metadata.body.name).toEqual(newAlias);
+      expect(metadata.body.primaryAlias).toBeTruthy();
+      expect(metadata.body.noteId).toEqual(publicId);
```
davidmehren, line 146: We should also check that the database was updated

test/private-api/alias.e2e-spec.ts
```diff
@@ -0,0 +1,205 @@
...
+      const newAlias = 'normalAlias';
+      newAliasDto.newAlias = newAlias;
+      const metadata = await request(app.getHttpServer())
+        .post(`/alias`)
+        .set('Content-Type', 'application/json')
+        .send(newAliasDto)
+        .expect(201);
+      expect(metadata.body.name).toEqual(newAlias);
+      expect(metadata.body.primaryAlias).toBeFalsy();
+      expect(metadata.body.noteId).toEqual(publicId);
```
davidmehren, line 102: We should also check that the database was updated

test/public-api/alias.e2e-spec.ts
```diff
@@ -0,0 +1,201 @@
...
+      const newAlias = 'normalAlias';
+      newAliasDto.newAlias = newAlias;
+      const metadata = await request(app.getHttpServer())
+        .post(`/alias`)
+        .set('Content-Type', 'application/json')
+        .send(newAliasDto)
+        .expect(201);
+      expect(metadata.body.name).toEqual(newAlias);
+      expect(metadata.body.primaryAlias).toBeFalsy();
+      expect(metadata.body.noteId).toEqual(publicId);
```
davidmehren, line 98: Should check if the database was updated

test/public-api/alias.e2e-spec.ts
```diff
@@ -0,0 +1,201 @@
...
+
+    it('works with normal alias', async () => {
+      const metadata = await request(app.getHttpServer())
+        .put(`/alias/${newAlias}`)
+        .set('Content-Type', 'application/json')
+        .send(changeAliasDto)
+        .expect(200);
+      expect(metadata.body.name).toEqual(newAlias);
+      expect(metadata.body.primaryAlias).toBeTruthy();
+      expect(metadata.body.noteId).toEqual(publicId);
```
davidmehren, line 142: Should check if the database was updated

test/public-api/alias.e2e-spec.ts
```diff
@@ -0,0 +1,201 @@
...
+    const newAlias = 'normalAlias3';
+    beforeAll(async () => {
+      const note = await notesService.createNote(content, testAlias, user);
+      await aliasService.addAlias(note, newAlias);
+    });
+
+    it('works with normal alias', async () => {
+      await request(app.getHttpServer())
+        .delete(`/alias/${newAlias}`)
+        .expect(204);
```
davidmehren, line 175: Should check if the database was updated

davidmehren (changes requested):
I stopped commenting on the E2E tests after realizing that all tests should ensure that the database was updated on success and that the object is still unmodified on failure.

src/api/private/alias/alias.controller.ts
```diff
@@ -0,0 +1,140 @@
...
+    private noteService: NotesService,
+    private userService: UsersService,
+    private permissionsService: PermissionsService,
+  ) {
+    this.logger.setContext(AliasController.name);
+  }
+
+  @Post()
+  async addAlias(
+    @Req() req: Request,
```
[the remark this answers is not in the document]
DerMolly: This will be fixed inanother PR. This made no sense to include here as we'll need to touch all theses methods again…

src/api/private/alias/alias.controller.ts
```diff
@@ -0,0 +1,140 @@
...
+      if (e instanceof ForbiddenIdError) {
+        throw new BadRequestException(e.message);
+      }
+      throw e;
+    }
+  }
+
+  @Put(':alias')
+  async makeAliasPrimary(
+    @Req() req: Request,
```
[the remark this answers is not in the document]
DerMolly: See above

src/api/private/alias/alias.controller.ts
```diff
@@ -0,0 +1,140 @@
...
+        throw new BadRequestException(e.message);
+      }
+      throw e;
+    }
+  }
+
+  @Delete(':alias')
+  @HttpCode(204)
+  async removeAlias(
+    @Req() req: Request,
```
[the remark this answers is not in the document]
DerMolly: See above

src/history/history.service.ts
DerMolly: We could do that, but I'm not sure if that is too wise as `getNoteByIdOrAlias` does 8 joins and a subquery which basically is what we do here. I'm uncertain what the best way forward is here…

src/notes/utils.ts
DerMolly: This should not happen, as if there are aliases one of them should be the primaryAlias. If there are actually no aliases `filter` should return `undefined`, which probably should be part of the return type.

src/notes/alias.service.ts
```diff
@@ -0,0 +1,189 @@
...
+    if (note.aliases.length === filteredAliases.length) {
+      this.logger.debug(
+        `The alias '${alias}' is not used by this note or is the primary alias, that can't be removed.`,
+        'removeAlias',
+      );
+      throw new NotInDBError(
+        `The alias '${alias}' is not used by this note or is the primary alias, that can't be removed.`,
+      );
+    }
+    const aliasToDelete = note.aliases.find(
```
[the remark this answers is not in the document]
DerMolly: Please open an issue for this, as I don't think it's wise to add this in this PR.

src/notes/alias.service.ts
```diff
@@ -0,0 +1,189 @@
...
+   * @param {Note} note - the note to change the primary alias
+   * @param {string} alias - the alias to be the new primary alias of the note
+   * @throws {NotInDBError} the alias is not part of this note.
+   */
+  async makeAliasPrimary(note: Note, alias: string): Promise<Note> {
+    let newPrimaryFound = false;
+    let oldPrimaryId = '';
+    let newPrimaryId = '';
+
+    for (const anAlias of note.aliases) {
```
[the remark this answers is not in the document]
DerMolly: Please open an issue for this, as I don't think it's wise to add this in this PR.

src/history/history.service.ts
davidmehren: Let's leave it for now and wait for https://github.com/hedgedoc/hedgedoc/issues/1642, which should solve the issue

davidmehren (approved):

PR #1368, opened 2021-06-07
Remove script-loader

davidmehren:
### Component/Part
Webpack config

### Description
[script-loader](https://github.com/webpack-contrib/script-loader) is deprecated since 2 years and a very ugly way of loading libraries. This PR removes script-loader and replaces them with supported webpack loaders.
It also replaces our vendored dependencies with upstream versions on NPM where appropriate (meaning when our version does not look patched).

### Steps

- [x] Added implementation
- [x] I read the [contribution documentation](https://github.com/hedgedoc/hedgedoc/blob/develop/CONTRIBUTING.md) and
  made sure that:
  - My commits are signed-off to accept the DCO.
  - This PR targets the correct branch: `master` for 1.x & docs, `develop` for 2.x

PR #1370, opened 2021-06-07
Lazy-load various frontend-libraries

davidmehren:
### Component/Part
frontend JS

### Description
This PR uses `require.ensure` to lazy-load various frontend libraries, reducing the size of the initial JS bundle by 60%.
See the individual commit messages for more details.

### Steps

- [x] Added implementation
  made sure that:


davidmehren:
This is a draft, as #1368 must be merged first.

ErikMichelson (approved):

public/js/index.js
```diff
@@ -106,36 +105,198 @@ const cursorActivityDebounce = 50
 const cursorAnimatePeriod = 100
 const supportContainers = ['success', 'info', 'warning', 'danger']
 const supportCodeModes = [
```
SISheogorath, line 107: We could drop the hardcoding here: https://github.com/hedgedoc/hedgedoc/pull/21/files#diff-93842c4bc014d741ce85adbcd0da539017de00a76b65956e403c4732d925788fR95

SISheogorath (changes requested):
One remark, which could be considered.

InnayTool (approved):

mrdrogdrog (approved):

mrdrogdrog (approved):

public/js/index.js
davidmehren: I'm not sure if I understand.
I explicitly hardcoded the list of languages to get rid of the import of `highlight.js`, as it can't be lazy-loaded and otherwise would get included in all bundles that contain `index.js`,

SISheogorath (approved):

public/js/extra.js
```diff
@@ -388,25 +383,26 @@ export function finishView (view) {
...
-      mermaid.init(undefined, $ele)
-    } catch (err) {
-      let errormessage = err
-      if (err.str) {
-        errormessage = err.str
+    const $value = $(value)
+    const $ele = $(value).closest('pre')
+    require.ensure([], function (require) {
+      try {
+        const mermaid = require('mermaid')
```
DerMolly, line 390: Doesn't this load mermaid for each instance of a mermaid diagram newly? Or is that handled by `require.ensure`?

public/js/extra.js
```diff
@@ -415,14 +411,15 @@ export function finishView (view) {
...
       const $ele = $(value).parent().parent()
-
-      abcjs.renderAbc(value, $value.text())
-
-      $ele.addClass('abc')
-      $value.children().unwrap().unwrap()
-      const svg = $ele.find('> svg')
-      svg[0].setAttribute('viewBox', `0 0 ${svg.attr('width')} ${svg.attr('height')}`)
-      svg[0].setAttribute('preserveAspectRatio', 'xMidYMid meet')
+      require.ensure([], function (require) {
```
DerMolly, line 414: same here?

public/js/extra.js
```diff
@@ -372,13 +365,15 @@ export function finishView (view) {
...
       $value = $(value)
       const $ele = $(value).parent().parent()
-
-      const graphviz = Viz($value.text())
-      if (!graphviz) throw Error('viz.js output empty graph')
-      $value.html(graphviz)
-
-      $ele.addClass('graphviz')
-      $value.children().unwrap().unwrap()
+      require.ensure([], function (require) {
```
DerMolly, line 368: same here

DerMolly (changes requested):

public/js/extra.js
davidmehren: `require` caches resolved modules.
> Modules are cached after the first time they are loaded. This means (among other things) that every call to require('foo') will get exactly the same object returned, if it would resolve to the same file.

See https://nodejs.org/api/modules.html#modules_caching

mrdrogdrog: The require command isn't included into the resulting JavaScript. It's replaced with the imported call. So there is no "reloading again".
Require.ensure is the command that does the code splitting. It tells webpack to put the inner code into a separate chunk.

https://webpack.js.org/api/module-methods/#requireensure

DerMolly (approved):

PR #1394, opened 2021-06-17
Remove cdn support

mrdrogdrog:
### Component/Part
CDN Support

### Description
This PR removes CDN support

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:


mrdrogdrog:
The CDN links are still present in the export templates

davidmehren:
The only remaining place where we use a CDN is https://github.com/hedgedoc/hedgedoc/blob/master/public/js/extra.js#L632, which replaces the URLs to emoji-images with URLs to Cloudflare.

I see two ways to remove that:
- Just use the UTF-8 codepoint and hope that the user has a compatible font. That also requires to have a way to get the codepoint from the emoji name.
- Inline each emoji-image using a data URI in the  [`generateCleanHTML`](https://github.com/hedgedoc/hedgedoc/blob/master/public/js/extra.js#L615) function.

mrdrogdrog:
I guess the data URI method would make more sense. The detection for the first is a bit to unsafe.

ccoenen:
I would want the export to be as self-sufficient as possible.

a) I do not want to ping some CDN each time I look at a document
b) Yes it's a big company, but what if they at some point go down. Or decide to update their server names ... and so on.

mrdrogdrog:
@ccoenen i don't get what you want to achieve with that comment. The whole reason of this pull request is to remove CDN support. Nobody argues for keeping CDN in the export 😅

ccoenen:
yeah, me neither. Sorry for the noise. 😆

InnayTool:
Two things in the raw export are not optimal by now:
- The exported icons are too large.
- The exported document has references to the instance from which the html was exported.

davidmehren:
> Two things in the raw export are not optimal by now:
> 
>   - The exported icons are too large.

This is probably intended, as the raw export does not contain any CSS that would influence the emojis to be the correct size

 
>   - The exported document has references to the instance from which the html was exported.

This depends on the definition of "raw export". Currently, the export contains the exact HTML that was shown in the view-pane of the editor, but without any CSS or JavaScript. We could of course also run that through the "inline emoji images" logic, but then it would be less "raw". Also, a change here might break someone's script.

In any case, this problem has nothing to do with CDNs, so I'd like to keep it out of this PR.

InnayTool (approved):

ErikMichelson (approved):

DerMolly (approved):

PR #1463, opened 2021-07-21
feat: Remove target _blank when link start by .

kosssi:
### Description
This PR remove target _blank on internal link.

### Steps

- [X] Added implementation
  made sure that:
  - This PR targets the correct branch: `master` for 1.x

### DIFF
--- a/public/js/extra.js
+++ b/public/js/extra.js
@@ -574,7 +574,7 @@ export function postProcess (code) {
   // link should open in new window or tab
   // also add noopener to prevent clickjacking
   // See details: https://mathiasbynens.github.io/rel-noopener/
-  result.find('a:not([href^="#"]):not([target])').attr('target', '_blank').attr('rel', 'noopener')
+  result.find('a:not([href^="#"]):not([href^="."]):not([target])').attr('target', '_blank').attr('rel', 'noopener')
   // update continue line numbers
   const linenumberdivs = result.find('.gutter.linenumber').toArray()
   for (let i = 0; i < linenumberdivs.length; i++) {


KillianKemps (approved):

ccoenen:
if this is meant for all relative links, then they could also be `href="abcdef"` - where abcdef would be the other document's ID. Currently this would not be covered.

Why do links get the `target="_blank"` in the first place? Do we even want that?

kosssi:
Maybe we can just add `_blank` and `noopener` on external link?

`$("a[href^='http']")` from [stackoverflow](https://stackoverflow.com/a/1227670/6429616)

ErikMichelson:
If we really want to drop the `target=_blank` on external links, we could simplify the whole query by just adding the `noopener` attribute to every link. HedgeDoc doesn't use the `opener` property internally, so there should be nothing that breaks.

However, I'd consider it a bad decision to change the behaviour of link clicks with a minor version. Users are probably used to just click a link in a note and let it open in a new tab. Changing this behaviour is kind of breaking. It's true that you can't open a link in the same tab as the note with `target=_blank`, but you still have the option to close the HedgeDoc tab directly afterwards, so I see no greater problem with that.

The problem to solve is how to distinguish internal from external links. The query mentioned above is not sufficient as it matches all internal links starting with "http" too. For example is a link from one note to another note with the alias "http-documentation" suddenly an external link.
We could however check for either `http://` or `https://` at the beginning of the link.

Such a query looks like this:
```javascript
result.find('a[href^="http://"],a[href^="https://"]').attr('target', '_blank').attr('rel', 'noopener')
```

Btw: The original code seems to set the `rel=noopener` attribute only when the link does not contain a `target` attribute. This seems wrong and could be solved by using two finds.
```javascript
result.find('a[href^="http://"],a[href^="https://"]').attr('rel', 'noopener')
result.find('a[href^="http://"]:not([target]),a[href^="https://"]:not([target])').attr('target', '_blank')
```

kosssi:
For my part I will remove all `target='_blank'` but I know that this is a big change and that it cannot be part of a minor version, that's why I only proposed the explicit links starting with a point ;)

This article can help us https://css-tricks.com/use-target_blank/

kosssi:
"The 1.x release is maintenance-only. We do not accept feature requests or PRs for this release anymore and may choose to close non-critical bug reports, if the bug will be non-existent in 2.0."

PR #1503, opened 2021-08-04
Update manual-setup.md

El-Virus:
Added "--prod=false" to yarn install so the NODE_ENV variable is overridden, and the development dependencies are installed.

### Component/Part
Documentation

### Description
This PR Adds "--prod=false" to "yarn install" in the manual installation documentation so the NODE_ENV variable is overridden, and the development dependencies are installed.

### Steps

- [X] I read the [contribution documentation](https://github.com/hedgedoc/hedgedoc/blob/develop/CONTRIBUTING.md) and
  made sure that:

### DIFF
--- a/docs/content/setup/manual-setup.md
+++ b/docs/content/setup/manual-setup.md
@@ -34,7 +34,7 @@
    It's also possible to use environment variables.
    For details, have a look at [the configuration documentation](../configuration.md).
 5. *:octicons-light-bulb-16: If you use the release tarball for 1.7.0 or newer, this step can be skipped.*  
-   Build the frontend bundle by running `yarn install` and then `yarn build`. The extra `yarn install` is necessary as `bin/setup` does not install the build dependencies.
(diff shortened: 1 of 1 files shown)


davidmehren:
This argument is only necessary if the `NODE_ENV` environment variable has been manually set to `production`. By default, everything should work with our current instructions.

As we would like to keep our installation instructions reasonably concise and manually setting the environment variable implies some knowledge of Node.js, I'm inclined to leaving the docs as is.

InnayTool:
@El-Virus: Is there a specific reason for having set `NODE_ENV`, even if you want to install development dependencies?

El-Virus:
@InnayTool, not really. That's the whole point, as some people have `NODE_ENV` set to production by default in their enviroment.

ErikMichelson:
The `NODE_ENV` should be set as environment variable per process or application but not global.
That's an edge case and I don't think that we need to include this in the docs.

El-Virus:
@ErikMichelson I see your reasoning, Maybe the documentation should have a note on that though.

PR #1514, opened 2021-08-08
docs: add documentation on private api authentication

DerMolly:
### Component/Part
docs

### Description
This PR adds documentation on private api authentication

### Steps

  made sure that:

### DIFF
--- a/docs/content/dev/private-api-auth.md
+++ b/docs/content/dev/private-api-auth.md
@@ -1,0 +1,18 @@
+# Private API Auth
+
+## Supported kinds of authentication
+
+- Username & Password (`local`)
+- LDAP 
+- SAML
+- OAuth2
+- GitLab
+- GitHub
+- Facebook
+- Twitter
+- Dropbox
+- Google
+
+## How the authentication works
+
+The backend is called directly from the frontend. The different routes that handle different kinds of authentication perform any kind of verification needed and then create a session cookie. This session cookie is than provided with each subsequent call to the private api by the frontend (until it expires or the user logs out). The SessionGuard, which is added to each other (appropriate) controller method of the private api, checks if the provided session is still valid and provides the controller method with the correct user.


docs/content/dev/private-api-auth.md
```diff
@@ -0,0 +1,18 @@
...
+- GitLab
+- GitHub
+- Facebook
+- Twitter
+- Dropbox
+- Google
+
+## How the authentication works
+
+The backend is called either directly from the frontend or via an OAuth provider. The different routes that handle different kinds of authentication perform any kind of verification needed and then create a JWT. This JWT is than provided with each subsequent call to the private api by the frontend (until it expires or the user logs out). The JwtAuthGuard, which is added to each other controller method of the private api, checks if the provided JWT is still valid and provides the controller method with the correct user.
```
davidmehren, line 18: > The backend is called either directly from the frontend or via an OAuth provider.

Could you elaborate a bit on how this looks? What route do I use? What do I send there?
You could start with just user & password, so we can add the other options later.

> and then create a JWT

Why not a cookie?

> This JWT is than provided with each subsequent call

How?

DerMolly: > Could you elaborate a bit on how this looks? What route do I use? What do I send there?
You could start with just user & password, so we can add the other options later. 

The routes were defined back in March in this Issue https://github.com/hedgedoc/hedgedoc/issues/987. I'd go with them. Each route would work slightly different, but this will not be part of this docs of course, as there is the auto generated private api docs.

> Why not a cookie?

You wanted docs first and then discuss how to do it later. So interpet this a mere suggestion.

> How?

I added an explantation and a link to the coressponding RFC.

docs/content/dev/private-api-auth.md
```diff
@@ -0,0 +1,18 @@
...
+- GitLab
+- GitHub
+- Facebook
+- Twitter
+- Dropbox
+- Google
+
+## How the authentication works
+
+The backend is called either directly from the frontend or via an OAuth provider. The different routes that handle different kinds of authentication perform any kind of verification needed and then create a session cookie. This session cookie is than provided with each subsequent call to the private api by the frontend (until it expires or the user logs out). The SessionGuard, which is added to each other (appropriate) controller method of the private api, checks if the provided session is still valid and provides the controller method with the correct user.
```
ErikMichelson, line 18: Even when using an (o)auth provider, the backend is called directly from the frontend because only the backend knows the app-key and secret of the provider.

DerMolly: I changed that

mrdrogdrog (approved):

ErikMichelson (approved):

davidmehren (approved):

PR #1517, opened 2021-08-08
Add local login with username and password

### LINKED ISSUE #1515: Login with Username & Password
https://github.com/hedgedoc/hedgedoc/issues/987 mentions these routes for login with username and password:

- [ ] POST `/auth/internal/login` for logins
- [ ] POST `/auth/internal` for registering
- [ ] PUT `/auth/internal` for updating the password

I'm fine with calling the login method "internal", even though it sounds a bit awkward.

DerMolly:
### Component/Part
internal login, registration and password changes

### Description
This PR adds the basic functionality of an internal login for the private api

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
closes #1515


src/api/private/auth/auth.controller.ts
```diff
@@ -0,0 +1,83 @@
...
+  ) {
+    this.logger.setContext(AuthController.name);
+  }
+
+  @Post('internal')
+  async registerUser(@Body() registerDto: RegisterDto): Promise<void> {
+    try {
+      const user = await this.usersService.createUser(
+        registerDto.username,
+        registerDto.username,
```
ErikMichelson, line 40: I guess we wanted to separate username from display name in 2.x.
At least we built that in the frontend a year ago.

davidmehren:
Hmm, the auth stuff for the public API is in `src/auth`, but for the private API it's in `src/api/private/auth`?

DerMolly:
> Hmm, the auth stuff for the public API is in `src/auth`, but for the private API it's in `src/api/private/auth`?

Actuall the auth for the public API is in `src/auth` and the auth for the private API is in `src/identity`. Not to keen on that either, but I think it's best to seperate those two. Maybe instead of `src/auth` ist should be `src/public-auth` or something. `src/identity` makes sense for the private api imho, because identity holds all the information for the different auth types and so on. 

Regarding `src/api/private/auth` the equivalent for the public api would be `src/api/private/tokens` since the tokens for the public api will be created via the frontend…

src/api/private/auth/auth.controller.ts
DerMolly: In principle it's seperated, but the current frontend implementation only asks for a username and not for a seperate display name so I kept with that. In principle on could add a display name to the RegisterDto and change those lines. 
grafik

DerMolly: I added the displayname to the RegisterDto just now

DerMolly:
CodeQL is remarking that the key `my-secret` is hardcoded. As you can see here https://docs.hedgedoc.org/configuration/#users-and-privileges in HD1 we made the following things, about the session changable:

- name (default: connect-sid)
- expire time (default 1209600000 (14 days))
- session secret

I guess for compatibility reasons allone we should keep all those and put in session object of the authConfig (since it affacts all configured auth methods).

src/utils/session.ts
```diff
@@ -0,0 +1,26 @@
...
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { TypeormStore } from 'connect-typeorm';
+import session from 'express-session';
+import { Repository } from 'typeorm';
+
+import { Session } from '../users/session.entity';
+
+export function createSession(app: INestApplication) {
+  app.use(
+    session({
```
ErikMichelson, line 16: I propose to set the `name` attribute per default to something like `hedgedoc-session`.
This has particularly two reasons:
1. People get an idea what the cookie is about. It results in more trust than some cookie with an arbitrary name like `connect.sid`.
2. Running multiple nodejs apps relying on the express-session module on your local machine often results in problems because they all share the same session cookie. Setting another name per default makes development runs easier as you don't need to configure something for that.

src/identity/identity.service.ts
```diff
@@ -0,0 +1,136 @@
...
+      user,
+      ProviderType.LOCAL,
+    );
+    if (internalIdentity === undefined) {
+      this.logger.debug(
+        `The user with the username ${user.userName} does not have a internal identity.`,
+        'loginWithLocalIdentity',
+      );
+      throw new NotInDBError(
+        'This username and password combination did not work.',
```
davidmehren, line 123: Hmm, shouldn't the controller-layer create this user-facing error message?

src/identity/identity.service.ts
```diff
@@ -0,0 +1,136 @@
...
+  /**
+   * @async
+   * Create a new identity for internal auth
+   * @param {User} user - the user the identity should be added to
+   * @param {string} password - the password the identity should have
+   * @throws {FunctionalityDisabledError} user registration is disabled
+   * @return {Identity} the new local identity
+   */
+  async createLocalIdentity(user: User, password: string): Promise<Identity> {
+    this.checkIfRegistrationEnabled();
```
davidmehren, line 65: Using exceptions as control flow is [bad](https://softwareengineering.stackexchange.com/questions/189222/are-exceptions-as-control-flow-considered-a-serious-antipattern-if-so-why).

We could wrap `createLocalIdentity` and `changeLocalPassword` with a decorator instead, but just throwing explicitly in two functions is probably also fine.

src/identity/identity.service.ts
```diff
@@ -0,0 +1,136 @@
...
+      );
+      throw new FunctionalityDisabledError('User registration is disabled.');
+    }
+  }
+
+  /**
+   * Check if the local login is enabled
+   * @throws {FunctionalityDisabledError} user login is disabled
+   */
+  checkIfLoginIsEnabled(): void {
```
davidmehren, line 49: Should probably be called `localLogin` or `localAuth`, as it does not check if *any* login is enabled.

src/identity/identity.service.ts
```diff
@@ -0,0 +1,136 @@
...
+    if (!this.authConfig.email.enableLogin) {
+      this.logger.debug('User login is disabled.', 'checkIfLoginIsEnabled');
```
davidmehren, line 51:
```suggestion
      this.logger.debug('Local auth is disabled.', 'checkIfLoginIsEnabled');
```

src/utils/session.ts
```diff
@@ -0,0 +1,31 @@
...
+import { INestApplication } from '@nestjs/common';
+import { getRepositoryToken } from '@nestjs/typeorm';
+import { TypeormStore } from 'connect-typeorm';
+import session from 'express-session';
+import { Repository } from 'typeorm';
+
+import { AuthConfig } from '../config/auth.config';
+import { Session } from '../users/session.entity';
+
+export function createSession(app: INestApplication, authConfig: AuthConfig) {
```
davidmehren, line 15:
```suggestion
export function setupSessionMiddleware(app: INestApplication, authConfig: AuthConfig) {
```
as this function does not create a single session ;)

src/identity/local/local.strategy.ts
```diff
@@ -0,0 +1,49 @@
...
+        UserRelationEnum.IDENTITIES,
+      ]);
+      await this.identityService.loginWithLocalIdentity(user, password);
+      return user;
+    } catch (e) {
+      if (e instanceof FunctionalityDisabledError) {
+        throw new BadRequestException(e.message);
+      }
+      if (e instanceof NotInDBError) {
+        throw new UnauthorizedException(e.message);
```
davidmehren, line 44: Doesn't this leak if a username does not exist in the database, as we call `getUserByUsername` first?

src/identity/identity.service.ts
```diff
@@ -0,0 +1,136 @@
...
+        `The user with the username ${user.userName} does not have a internal identity.`,
+        'loginWithLocalIdentity',
+      );
+      throw new NotInDBError(
+        'This username and password combination did not work.',
+      );
+    }
+    if (!(await checkPassword(password, internalIdentity.passwordHash ?? ''))) {
+      this.logger.debug(
+        `The user with the username ${user.userName} has another password.`,
```
davidmehren, line 128:
```suggestion
        `Password check for ${user.userName} did not succeed.`,
```

src/api/private/auth/auth.controller.ts
```diff
@@ -0,0 +1,106 @@
...
+      if (e instanceof FunctionalityDisabledError) {
+        throw new BadRequestException(e.message);
+      }
+      throw e;
+    }
+  }
+
+  @UseGuards(SessionGuard)
+  @Put('local')
+  async changePassword(
```
davidmehren, line 69:
```suggestion
  async updatePassword(
```
The other `@Put` handlers are all named `update...`

src/api/private/auth/auth.controller.ts
```diff
@@ -0,0 +1,106 @@
...
+    @Req() request: Request & { session: { user: string } },
+    @Body() changePasswordDto: ChangePasswordDto,
+  ): Promise<void> {
+    try {
+      if (!request.session.user) {
```
davidmehren, line 74: Shouldn't this be not necessary, as `SessionGuard` would throw if `request.session.user` doesn't exist?

davidmehren (changes requested):

src/identity/identity.service.ts
DerMolly: I think I fixed this…

src/identity/local/local.strategy.ts
DerMolly: I think this if fixed now

src/identity/identity.entity.ts
```diff
@@ -24,6 +25,9 @@ export class Identity {
   })
   user: User;
 
+  @Column()
+  providerType: string;
```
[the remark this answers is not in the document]
DerMolly: We don't do this anywhere else?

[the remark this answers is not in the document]
DerMolly: I added what I could…

src/identity/identity.entity.ts
```diff
@@ -0,0 +1,98 @@
...
+   * When the identity was last updated.
+   */
+  @UpdateDateColumn()
+  updatedAt: Date;
+
+  @Column({
+    nullable: true,
+    type: 'text',
+  })
+  providerUserId: string | null;
```
DerMolly, line 73: We still need a doc string here

src/identity/identity.entity.ts
```diff
@@ -0,0 +1,98 @@
...
+
+  @Column({
+    nullable: true,
+    type: 'text',
+  })
+  oAuthAccessToken: string | null;
```
DerMolly, line 79: We still need a doc string here

davidmehren (changes requested):

src/identity/identity.service.spec.ts
```diff
@@ -0,0 +1,124 @@
...
+        .spyOn(identityRepo, 'save')
+        .mockImplementationOnce(
+          async (identity: Identity): Promise<Identity> => identity,
+        );
+      const identity = await service.createLocalIdentity(user, password);
+      user.identities = Promise.resolve([identity]);
+    });
+    it('works', async () => {
+      jest
+        .spyOn(identityRepo, 'save')
```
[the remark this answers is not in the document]
DerMolly: Yeah I think you're right

src/api/private/auth/auth.controller.ts
```diff
@@ -0,0 +1,104 @@
...
+    // There is no further testing needed as we only get to this point if LocalAuthGuard was successful
+    request.session.user = loginDto.username;
+  }
+
+  @UseGuards(SessionGuard)
+  @Delete('logout')
+  logout(@Req() request: Request & { session: Session }): void {
+    request.session.destroy((err) => {
+      if (err) {
+        throw new BadRequestException('Unable to log out');
```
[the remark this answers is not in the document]
DerMolly: Of course. I changed that.

src/auth/auth.service.ts
```diff
@@ -155,7 +141,9 @@ export class AuthService {
   }
 
   async getTokensByUsername(userName: string): Promise<AuthToken[]> {
```
InnayTool, line 143: I think we should change this also to `user: User`.

InnayTool (changes requested):

src/auth/auth.service.ts
```diff
@@ -52,43 +57,24 @@ export class AuthService {
...
-    // base64url is quite easy buildable from base64
-    return text
-      .toString('base64')
-      .replace(/\+/g, '-')
-      .replace(/\//g, '_')
-      .replace(/=+$/, '');
-  }
-
   async createTokenForUser(
     userName: string,
```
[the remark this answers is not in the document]
DerMolly: As I didn't change this code in this PR, I'd suggest to open an extra PR for that…

src/auth/auth.service.ts
DerMolly: As I didn't change this code in this PR, I'd suggest to open an extra PR for that…

src/identity/utils.ts
```diff
@@ -0,0 +1,21 @@
+/*
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+import { User } from '../users/user.entity';
+import { Identity } from './identity.entity';
+import { ProviderType } from './provider-type.enum';
+
+export async function getIdentityFromUser(
```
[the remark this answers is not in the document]
DerMolly: Yeah this is correct and intended, but the name probably is not to good for that.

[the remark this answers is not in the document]
DerMolly: Done

src/config/auth.config.ts
```diff
@@ -332,9 +344,13 @@ export default registerAs('authConfig', () => {
 
   const authConfig = authSchema.validate(
     {
-      email: {
-        enableLogin: process.env.HD_AUTH_EMAIL_ENABLE_LOGIN,
-        enableRegister: process.env.HD_AUTH_EMAIL_ENABLE_REGISTER,
+      session: {
+        secret: process.env.HD_SESSION_SECRET,
+        lifeTime: parseOptionalInt(process.env.HD_SESSION_LIFE_TIME),
```
SISheogorath, line 349: I think the english word lifetime would be fine without camelcasing it.

src/utils/password.spec.ts
```diff
@@ -0,0 +1,36 @@
+/*
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+import { randomBytes } from 'crypto';
+
+import { bufferToBase64Url, checkPassword, hashPassword } from './password';
```
SISheogorath, line 8: There should be an additional chrck to make sure we use bcrypt. Currently, if bcrypt broke and return the string itself, we wouldn't notice, would we?

Or to put it more into perspective of test-driven development: If one would like to update the password schema, would they need to touch the tests?

SISheogorath (changes requested):
Some minor remarks

src/utils/password.spec.ts
DerMolly: I could add
```
expect(hash).not.toEqual(testPassword);
```

src/config/auth.config.ts
DerMolly: Then it should be 
```
lifetime: parseOptionalInt(process.env.HD_SESSION_LIFETIME),
```
correct?

SISheogorath: Yes

src/utils/password.spec.ts
SISheogorath: Then you just verify that thr password is not just returned, but you actually want to verify that it's using a bcrypt hash.

We need an actual hashed result of the password and compare it to the output of the password hash function.

DerMolly: > Then you just verify that the password is not just returned, but you actually want to verify that it's using a bcrypt hash.

We added a test that the bcrypt hash looks right (via a regex)

> We need an actual hashed result of the password and compare it to the output of the password hash function.

As bcrypt doesn't compute the same result twice. I'm not sure if we really want that. I guess we would need to mock the PRNG used in bcrypt to achive that.…

src/config/auth.config.ts
DerMolly: done

src/utils/password.spec.ts
mrdrogdrog: IMHO it's not our responsibility to test node functions. Or other packages. It should be enough to test if the bcrypt function was called.

mrdrogdrog: So.. mock the bcrypt function and check if it was called.

DerMolly: iirc jest can check if a function as called without it being necessary to be mocked.

SISheogorath (approved):

InnayTool (approved):

davidmehren (approved):

ErikMichelson (approved):

PR #1545, opened 2021-08-14
Fix endless loop on shutdown when DB can't be reached

davidmehren:
### Component/Part
Shutdown handler

### Description
The shutdown handler calls `checkAllNotesRevision` on a 100 ms
interval. If the database connection is broken, this will return
an error. Previously, this error was effectively ignored and resulted
in an endless loop printing out the error message every 100 ms.

This improves the error handling by terminating the process with a
nonzero exit code when an error was encountered.

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/app.js
+++ b/app.js
@@ -309,9 +309,15 @@ process.on('uncaughtException', function (err) {
   process.exit(1)
 })
 
+let alreadyHandlingTermSignals = false
 // install exit handler
 function handleTermSignals () {
+  if (alreadyHandlingTermSignals) {
+    logger.info('Forcefully exiting.')
+    process.exit(1)
+  }
   logger.info('HedgeDoc has been killed by signal, try to exit gracefully...')
+  alreadyHandlingTermSignals = true
   realtime.maintenance = true
   // disconnect all socket.io clients
   Object.keys(io.sockets.sockets).forEach(function (key) {
--- a/app.js
+++ b/app.js
@@ -331,17 +331,28 @@ function handleTermSignals () {
       }
     })
   }
+  const maxCleanTries = 30
+  let currentCleanTry = 1
   const checkCleanTimer = setInterval(function () {
     if (realtime.isReady()) {
       models.Revision.checkAllNotesRevision(function (err, notes) {
-        if (err) return logger.error(err)
+        if (err) {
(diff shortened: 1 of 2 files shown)


SISheogorath:
Mhm, I'm kind of undecided on this one, since it's supposed to do a graceful shutdown, which isn't actually graceful when it can't write all revisions. However I can see your problem with the current behaviour. Mhm… Some programs have some kind of double <kbd>ctrl</kbd> + <kbd>c</kbd> mechanic? I mean, for actual production this should be non-issue, unless you run some ancient initrc system, that still has a shell script that tries to kill hedgedoc, you should be fine. Docker, systemd, … all have kill timeout by default that should take care. Only interactive use might still be an issue. Here the mentioned double <kbd>ctrl</kbd> + <kbd>c</kbd> might be a (better?) solution.

app.js
```diff
@@ -321,7 +321,10 @@ function handleTermSignals () {
   const checkCleanTimer = setInterval(function () {
     if (realtime.isReady()) {
       models.Revision.checkAllNotesRevision(function (err, notes) {
-        if (err) return logger.error(err)
+        if (err) {
+          logger.error('Error while saving note revisions: ' + err)
+          return process.exit(1)
```
InnayTool, line 326:
```suggestion
          process.exit(1)
```
This is awful.

InnayTool (changes requested):

davidmehren:
@SISheogorath  
I had a look at `saveAllNotesRevision` and could not see a case where we get an error in the callback that is *not* caused by sequelize. Are there DB errors that solve themselves after 100 ms? Maybe just retry a finite amount of times and then exit?

ErikMichelson:
Retrying for a finite amount of times and then exit sounds good in my opinion.

SISheogorath:
@davidmehren yes, a finite amount of times, also sounds fine to me 👍🏻

davidmehren:
I now retry up to 30 times and wait 200 ms after each try. This gives the database 6 seconds to recover, which should hopefully be enough for intermittent issues.

SISheogorath (approved):

InnayTool (approved):

ErikMichelson (approved):
image

PR #1559, opened 2021-08-18
Ignore local installed fonts

ErikMichelson:
### Component/Part
font loading

### Description
There were several reports([1](https://matrix.to/#/!oaRMJImPLfUnnXzwnN:shivering-isles.com/$aHMBKJA2QiGIkiQwad_9TQR53alK2DvFmvYkg-oNC0w?via=matrix.org&via=kif.rocks&via=matrix.michelson.eu))([2](https://community.hedgedoc.org/t/firefox-no-text-colors-in-editor/437)) of HedgeDoc not looking correctly when having some variants of fonts locally installed which HedgeDoc uses. The only way to fix this for the users was to remove the locally installed font or update them to another variant.

This PR removes the directive to load fonts from the local user's machine when they have the same name. Instead  always the font files from the server will be used.

As we use woff font files which aren't very heavy in terms of file-size, it seems acceptable to fetch them always from the server (or the local browser cache).

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/public/css/font.css
+++ b/public/css/font.css
@@ -3,7 +3,7 @@
   font-family: 'Source Code Pro';
   font-style: normal;
   font-weight: 300;
-  src: local('Source Code Pro Light'), local('SourceCodePro-Light'), url('../fonts/SourceCodePro-Light.woff') format('woff');
+  src: url('../fonts/SourceCodePro-Light.woff') format('woff');
   unicode-range: U+0100-024F, U+1E00-1EFF, U+20A0-20AB, U+20AD-20CF, U+2C60-2C7F, U+A720-A7FF;
 }
 /* latin */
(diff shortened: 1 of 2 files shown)


public/css/font.css
```diff
@@ -3,206 +3,206 @@
   font-family: 'Source Code Pro';
   font-style: normal;
   font-weight: 300;
-  src: local('Source Code Pro Light'), local('SourceCodePro-Light'), url('../fonts/SourceCodePro-Light.woff') format('woff');
+  src: url('../fonts/SourceCodePro-Light.woff') format('woff');
```
SISheogorath, line 6: Maybe we also want to switch to woff2 fonts?

davidmehren: PRs welcome, but let's focus this PR on the bug fix ;)

davidmehren (approved):

Issue #1064, opened 2021-03-25
Private API: Purge note history


davidmehren:
**Describe the solution you'd like**
The private API should provide a way to purge the note history.

The documentation merged in documentation merged in https://github.com/hedgedoc/hedgedoc/pull/1033 says:
> The owner of a note may delete all revisions. This effectively purges the edit history of a note.

Note: We expect that a Note *always* has at least one revision. Purging the history must therefore preserve the latest revision.

Abhilasha06:
Hey @davidmehren can I work on this?

davidmehren:
Hi @Abhilasha06, great that you want to help. Please join the [dev chat](https://chat.hedgedoc.org/dev) to coordinate your contribution :smiley:

davidmehren:
Fixed by #1595

PR #1595, opened 2021-08-30
Add new API to purge note history

Abhilasha06:
Signed-off-by: Abhilasha Sinha <abhisinha662000@gmail.com>

### Component/Part

Backend

### Description
This PR adds a new API to purge the history of a note. 
The API deletes all the revisions of a note except the latest one.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)

#1064

### DIFF
--- a/src/revisions/revisions.service.spec.ts
+++ b/src/revisions/revisions.service.spec.ts
@@ -97,4 +97,63 @@ describe('RevisionsService', () => {
       );
     });
   });
+
+  describe('purgeRevisions', () => {
+    it('purges the revision history', async () => {
+      const note = {} as Note;
+      note.id = 'test';
+      let revisions: Revision[] = [];
+      const revision1 = Revision.create('a', 'a');
+      revision1.id = 1;
+      const revision2 = Revision.create('b', 'b');
+      revision2.id = 2;
+      const revision3 = Revision.create('c', 'c');
+      revision3.id = 3;
+      revisions.push(revision1, revision2, revision3);
+      note.revisions = Promise.resolve(revisions);
+      jest.spyOn(revisionRepo, 'find').mockResolvedValueOnce(revisions);
+      jest.spyOn(service, 'getLatestRevision').mockResolvedValueOnce(revision3);
+      revisionRepo.remove = jest
+        .fn()
+        .mockImplementation((deleteList: Revision[]) => {
+          revisions = revisions.filter(
+            (item: Revision) => !deleteList.includes(item),
+          );
+          return Promise.resolve(deleteList);
+        });
+
+      // expected to return all the purged revisions
+      expect(await service.purgeRevisions(note)).toHaveLength(2);
+
+      // expected to have only the latest revision
+      const updatedRevisions: Revision[] = [revision3];
+      expect(revisions).toEqual(updatedRevisions);
+    });
+    it('has no effect on revision history when a single revision is present', async () => {
+      const note = {} as Note;
+      note.id = 'test';
+      let revisions: Revision[] = [];
+      const revision1 = Revision.create('a', 'a');
+      revision1.id = 1;
+      revisions.push(revision1);
+      note.revisions = Promise.resolve(revisions);
+      jest.spyOn(revisionRepo, 'find').mockResolvedValueOnce(revisions);
+      jest.spyOn(service, 'getLatestRevision').mockResolvedValueOnce(revision1);
+      revisionRepo.remove = jest
+        .fn()
+        .mockImplementation((deleteList: Revision[]) => {
+          revisions = revisions.filter(
+            (item: Revision) => !deleteList.includes(item),
+          );
+          return Promise.resolve(deleteList);
+        });
+
+      // expected to return all the purged revisions
+      expect(await service.purgeRevisions(note)).toHaveLength(0);
+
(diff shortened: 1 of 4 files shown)


ErikMichelson:
Thanks for your contribution!

src/revisions/revisions.service.ts
```diff
@@ -34,6 +34,23 @@ export class RevisionsService {
     });
   }
 
+  async purgeRevision(note: Note): Promise<Revision[]> {
```
DerMolly, line 37: This function should get it's own unit test in the `revsions.service.spec.ts` file.

DerMolly (changes requested):
This was very good work. Thank you very much. I'd be happy to merge this once the requested changes were made (and the code was rebased).

src/revisions/revisions.service.ts
```diff
@@ -34,6 +34,23 @@ export class RevisionsService {
     });
   }
```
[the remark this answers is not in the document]
DerMolly: I'm not sure we mention this already in the CONTRIBUTING docs, so we should probably add this there @davidmehren @InnayTool

src/revisions/revisions.service.ts
Abhilasha06: Done

DerMolly (changes requested):
You forgot to add the requested esdoc

Abhilasha06:
> You forgot to add the requested esdoc
Done

DerMolly (approved):

src/revisions/revisions.service.ts
InnayTool: I think this function should be called 
```suggestion
  async purgeRevisions(note: Note): Promise<Revision[]> {
```
or
```suggestion
  async purgeHistory(note: Note): Promise<Revision[]> {
```

src/revisions/revisions.service.ts
```diff
@@ -34,6 +34,26 @@ export class RevisionsService {
...
+    const revisions = await this.revisionRepository.find({
+      where: {
+        note: note,
+      },
+    });
+    const latestRevison = await this.getLatestRevision(note.id);
+    const latestRevisonIndex = revisions.findIndex(
+      (item) => item.id === latestRevison.id,
+    );
+    revisions.splice(latestRevisonIndex, 1);
```
davidmehren, line 53: Unfortunately, this code is buggy. If there is only one revision, it gets deleted.
You could fix this by using `filter` instead of `splice`:
```js
// get the latest revision, we want to keep it
const latestRevison = await this.getLatestRevision(note.id);
// get all revisions except the latest
const oldRevisions = revisions.filter(
 (item) => item.id !== latestRevison.id,
);
// delete the old revisions
return await this.revisionRepository.remove(oldRevisions);
```

Please also add a unit test for the "only one revision" case.

src/revisions/revisions.service.spec.ts
```diff
@@ -97,4 +97,27 @@ describe('RevisionsService', () => {
...
+      revision3.id = 3;
+      revisions.push(revision1, revision2, revision3);
+      note.revisions = Promise.resolve(revisions);
+      jest.spyOn(revisionRepo, 'find').mockResolvedValueOnce(revisions);
+      jest.spyOn(service, 'getLatestRevision').mockResolvedValueOnce(revision3);
+      revisionRepo.remove = jest.fn().mockImplementation(() => {
+        return Promise.resolve(note.revisions);
+      });
+      // expected to return all the purged revisions
+      expect(await service.purgeRevisions(note)).toHaveLength(2);
```
davidmehren, line 120: The test should also check if the expected revision remains in the 'database'.
You will need to change the mocked `remove` function to actually change the `revisions` array like this:
```js
revisionRepo.remove = jest
        .fn()
        .mockImplementation((deleteList: Revision[]) => {
          revisions = revisions.filter((item) => !deleteList.includes(item));
          return Promise.resolve(deleteList);
        });
```

davidmehren (changes requested):
Thanks for your contribution!

There are two things that still need work before we can merge this.  
Please also don't forget to [clean up the commit history](https://github.com/hedgedoc/hedgedoc/blob/develop/CONTRIBUTING.md#commit-guidelines)

src/revisions/revisions.service.spec.ts
```diff
@@ -97,4 +97,68 @@ describe('RevisionsService', () => {
...
+
+      // expected to return all the purged revisions
+      expect(await service.purgeRevisions(note)).toHaveLength(2);
+
+      // expected to have only the latest revision
+      const updatedRevisions: Revision[] = [revision3];
+      expect(revisions).toEqual(updatedRevisions);
+    });
+
+    describe('purgeRevisions', () => {
```
davidmehren, line 133: You can merge this `describe` block with the previous one.

davidmehren (changes requested):
Nearly there!

You seem to have accidentally included some other changes in your commit. You should be able to solve this by updating the `develop` branch, rebasing your branch on the updated `develop`, fixing the commit message and finally force-pushing. If you have any problems or questions, you can always ask in the dev chat :smiley:

davidmehren (approved):
Thanks!

Issue #1641, opened 2021-09-16
Ensure consistent behaviour of entity `create` methods


davidmehren:
**Describe the solution you'd like**
The `create` methods of our entities currently have very inconsistent parameters.
They should always only require the non-optional properties of the corresponding entity and no optional parameters.

They should also either return a complete and fully useable instance or return a `Pick` type.

**Describe alternatives you've considered**
Keep the current inconsistencies.

**Additional context**

DerMolly:
As I'm currently working on this, I noticed that both `session.entity.ts` and `tag.entity.ts` currently don't have any `create` methods  at all. Should they get some?

Currently session are created here https://github.com/hedgedoc/hedgedoc/blob/develop/src/utils/session.ts#L33-L37 and no `create` method is needed. 

Tags are currently not created, because we don't parse the yaml frontmatter of notes currently.

So I guess I'll leave it like it is know and we'll see how it goes from there.

PR #1680, opened 2021-09-25
fix: create methods

DerMolly:
### Component/Part
entities

### Description
This PR consolidates entities create methods

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
#1641

### DIFF
--- a/docs/content/dev/2.0.md
+++ b/docs/content/dev/2.0.md
@@ -28,3 +28,14 @@ The software provides two special groups which have no explicit users:
 - Run `yarn install` in both projects
 - Start the backend server in watch mode using `yarn start:dev`. The backend is then accessible on port 3000.
 - Start the frontend dev server using `yarn start`. The frontend is now accessible on port 3001. 
+
+## Entity `create` methods
+
+Because we need to have empty constructors in our entity classes for TypeORM to work, the actual constructor is a separate `create` method. These methods should adhere to these guidelines:
+
+- Only require the non-optional properties of the corresponding entity
+- Have no optional parameters
+- Have no lists which can be empty (so probably most of them)
+- Should either return a complete and fully useable instance or return a Pick/Omit type.
+- Exceptions to these rules are allowed, if they are mentioned in the method documentation
+
--- a/src/groups/group.entity.ts
+++ b/src/groups/group.entity.ts
@@ -43,11 +43,12 @@ export class Group {
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   private constructor() {}
 
-  public static create(name: string, displayName: string): Group {
+  public static create(name: string, displayName: string): Omit<Group, 'id'> {
     const newGroup = new Group();
-    newGroup.special = false; // this attribute should only be true for the two special groups
     newGroup.name = name;
     newGroup.displayName = displayName;
+    newGroup.special = false; // this attribute should only be true for the two special groups
+    newGroup.members = [];
     return newGroup;
   }
 }
--- a/src/groups/group.entity.ts
+++ b/src/groups/group.entity.ts
@@ -43,11 +43,15 @@ export class Group {
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   private constructor() {}
 
-  public static create(name: string, displayName: string): Omit<Group, 'id'> {
+  public static create(
+    name: string,
+    displayName: string,
+    special: boolean,
+  ): Omit<Group, 'id'> {
     const newGroup = new Group();
     newGroup.name = name;
     newGroup.displayName = displayName;
-    newGroup.special = false; // this attribute should only be true for the two special groups
+    newGroup.special = special; // this attribute should only be true for the two special groups
     newGroup.members = [];
     return newGroup;
   }
--- a/src/history/history-entry.entity.ts
+++ b/src/history/history-entry.entity.ts
@@ -28,15 +28,21 @@ export class HistoryEntry {
   @UpdateDateColumn()
   updatedAt: Date;
 
-  // The optional note parameter is necessary for the createNote method in the NotesService,
-  // as we create the note then and don't need to add it to the HistoryEntry.
-  public static create(user: User, note?: Note): HistoryEntry {
+  /**
+   * Create a history entry
+   * @param user the user the history entry is associated with
+   * @param note the note the history entry is associated with
+   * @param [pinStatus=false] if the history entry should be pinned
+   */
+  public static create(
+    user: User,
+    note: Note,
+    pinStatus = false,
+  ): Omit<HistoryEntry, 'updatedAt'> {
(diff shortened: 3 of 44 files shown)


src/groups/group.entity.ts
```diff
@@ -43,11 +43,12 @@ export class Group {
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   private constructor() {}
 
-  public static create(name: string, displayName: string): Group {
+  public static create(name: string, displayName: string): Omit<Group, 'id'> {
```
InnayTool, line 46: I think `special` should be added here because it is a mandatory attribute. Maybe it should have the default value false.
```suggestion
  public static create(name: string, displayName: string, special: boolean): Omit<Group, 'id'> {
```

src/history/history-entry.entity.ts
```diff
@@ -28,14 +28,13 @@ export class HistoryEntry {
   @UpdateDateColumn()
   updatedAt: Date;
 
-  // The optional note parameter is necessary for the createNote method in the NotesService,
-  // as we create the note then and don't need to add it to the HistoryEntry.
-  public static create(user: User, note?: Note): HistoryEntry {
+  public static create(
+    user: User,
+    note: Note,
```
InnayTool, line 33: I think `pinStatus` should be added here and have default value false.

src/notes/note.entity.ts
```diff
@@ -82,15 +92,19 @@ export class Note {
   // eslint-disable-next-line @typescript-eslint/no-empty-function
   private constructor() {}
 
-  public static create(owner?: User, alias?: string): Note {
+  public static create(owner?: User, alias?: string): Omit<Note, 'id'> {
```
InnayTool, line 95: The undefined user is never used in the software, so it may be better to remove the `?` even from the service method `createNote`.
```suggestion
  public static create(owner: User, alias?: string): Omit<Note, 'id'> {
```

src/media/media-upload.entity.ts
```diff
@@ -59,7 +59,7 @@ export class MediaUpload {
     extension: string,
     backendType: BackendType,
     backendData?: string,
-  ): MediaUpload {
+  ): Omit<MediaUpload, 'createdAt'> {
```
InnayTool, line 62: I think this entity has to be modified. Can the Note be null or not? What is with `fileUrl`? It is never set in the create method.
It would be nicer if the `backendData` has the default value `null`.

InnayTool (changes requested):

src/groups/group.entity.ts
DerMolly: I think this change does not make any sense what so ever. There are only two group that are special and they will be created on startup, so I think we don't need this.

src/history/history-entry.entity.ts
DerMolly: Not sure if we ever want to create a new HistoryEntry with pinStatus true directly, but I guess this could be usefull, so I'll add this.

src/groups/group.entity.ts
davidmehren: It *technically* is a required attribute, but it will only be used to create our two special groups. So I think it is reasonable to not include it here, as the parameter would be set to `false` most of the time.

src/history/history-entry.entity.ts
davidmehren: When new entries are created, aren't they always unpinned?

src/media/media-upload.entity.ts
davidmehren: `note` is nullable in the schema, "as it is possible to delete a note without also deleting the associated media uploads".
But when *creating* a new upload, a note must be provided. This fact should probably be mentioned in the docs of this method, though.

Having no `fileUrl` does not seem to make sense at first glance.

To be consistent with the other methods, `backendData` should not be part of this `create`.

DerMolly: The note can be `null`, if the note is deleted afterwards and the mediaUploads still persists. `backendData` can be either `null` or a `string` depending on the `backendType`. `fileUrl` should always be a `string`, but we're in the awkward position that we need the `id` of the upload to get the `fileUrl` from the appropiate upload handler… So I'm not sure what to do here…

davidmehren (changes requested):
Could you add a snippet to `docs/dev/2.0.md`, explaining the standard for `create` methods?
Something like this:

`create` methods in entities should follow these rules:
- Only require the non-optional properties of the corresponding entity
- Have no optional parameters
- Have no lists which can be empty (so probably most of them)
- Should either return a complete and fully useable instance or return a `Pick`/`Omit` type.
- Exceptions to these rules are allowed, if they are mentioned in the method documentation

src/notes/note.entity.ts
davidmehren: To create anonymous notes, `owner` must be nullable

docs/content/dev/2.0.md
```diff
@@ -28,3 +28,14 @@ The software provides two special groups which have no explicit users:
 - Run `yarn install` in both projects
 - Start the backend server in watch mode using `yarn start:dev`. The backend is then accessible on port 3000.
 - Start the frontend dev server using `yarn start`. The frontend is now accessible on port 3001. 
+
+## Entity `create` methods
+
+Because we need to have empty constructs for TypeORM to work. The actual constructors of all our entity classes is a `create` method. These methods should adhere to these few guidelines:
```
davidmehren, line 34:
```suggestion
Because we need to have empty constructors in our entity classes for TypeORM to work, the actual constructor is a separate `create` method. These methods should adhere to these guidelines:
```

davidmehren (approved):

src/notes/note.entity.ts
InnayTool: Yes, I know but this method is never called without this argument. It is always explicitly null.

src/groups/group.entity.ts
InnayTool: I think we should be consistent here and as you say it is called, why omit it? with the default value, you don't have to specify it.

src/notes/note.entity.ts
DerMolly: Because we didn't create this path yet

davidmehren: Yep, we will most certainly need to create Notes without an owner, so it doesn't make much sense to remove this now, only to add it later.

src/groups/group.entity.ts
davidmehren: After more discussion with @InnayTool, we came to the conclusion that including `special` here is preferable, as the `create` function will only be used to create the two hard-coded groups and to handle API requests at one or two places.

InnayTool (approved):

PR #1710, opened 2021-10-05
Add mergeOnEmail option

hrenard:
### Component/Part
Auth/Backend

### Description
This PR adds a mergeOnEmail option.

By default, there is no behavior changes. But if the option is enabled, when logging in with OAuth, the user will be found by profileid or by email. This allows a seamless migration from email authentication to OAuth.

### Steps

- [x] Added implementation
- [ ] Added / updated tests
- [ ] Added changelog snippet
  made sure that:


hrenard:
> Thanks for the PR @hrenard! Unfortunately, to quote our README

I understand. It's there for those who need it anyway.

> I have looked at this code and found some serious bugs. Like: A merge of profiles would kill the history and ownerships of at least one of the profiles because you would find only ONE profile on login.

This is why it is disabled by default. The idea is to facilitate the migration from local authentication to oauth (by disabling email authentication and enabling oauth). And therefore not to create 2 accounts in the first place.

The remaining problem is that if local authentication is not disabled and someone who has already registered with oauth, creates a local account with the same email.

Issue #1642, opened 2021-09-16
Lazy-load all relations


davidmehren:
**Which part of the project should be enhanced?**
Entity definitions

**Is your enhancement request related to a problem? Please describe.**
We currently need to ensure that each entity is always loaded from the database with the same `relations` option (which eagerly loads relations), to ensure consistent behavior at runtime. This is tedious and prone to error.

**Describe the solution you'd like**
All relations should be refactored to be lazily loaded. This allows us to remove all `relations` options.

**Describe alternatives you've considered**
Continue using the current method

**Additional context**

PR #1827, opened 2021-11-16
Lazy-load all relations

davidmehren:
### Component/Part
entities

### Description
This PR refactors the entity classes to lazy-load all relations.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
Closes https://github.com/hedgedoc/hedgedoc/issues/1642

### DIFF
--- a/src/auth/auth.service.ts
+++ b/src/auth/auth.service.ts
@@ -53,7 +53,11 @@ export class AuthService {
     }
     const accessToken = await this.getAuthTokenAndValidate(keyId, secret);
     await this.setLastUsedToken(keyId);
-    return await this.usersService.getUserByUsername(accessToken.user.username);
+    return await this.usersService.getUserByUsername(
+      (
+        await accessToken.user
+      ).username,
+    );
   }
 
   async createTokenForUser(
--- a/src/auth/auth.service.ts
+++ b/src/auth/auth.service.ts
@@ -65,9 +65,9 @@ export class AuthService {
     identifier: string,
     validUntil: TimestampMillis,
   ): Promise<AuthTokenWithSecretDto> {
-    user.authTokens = await this.getTokensByUser(user);
+    user.authTokens = this.getTokensByUser(user);
 
-    if (user.authTokens.length >= 200) {
+    if ((await user.authTokens).length >= 200) {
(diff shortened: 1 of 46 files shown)


InnayTool (approved):

src/auth/auth.service.ts
```diff
@@ -53,17 +53,21 @@ export class AuthService {
...
+    );
   }
 
   async createTokenForUser(
     user: User,
     identifier: string,
     validUntil: TimestampMillis,
   ): Promise<AuthTokenWithSecretDto> {
-    user.authTokens = await this.getTokensByUser(user);
+    user.authTokens = this.getTokensByUser(user);
```
DerMolly, line 68: Maybe a type hint would be useful here?

src/notes/alias.service.ts
```diff
@@ -87,7 +87,7 @@ export class AliasService {
     let oldPrimaryId = '';
     let newPrimaryId = '';
 
-    for (const anAlias of note.aliases) {
+    for (const anAlias of await note.aliases) {
```
DerMolly, line 90: Maybe it would be wise to await this once and then put it in a variable.

DerMolly (changes requested):

src/auth/auth.service.ts
davidmehren: I'm not declaring a variable here (the `User` class defines the type of the property), so I can't an explicit type.

src/notes/alias.service.ts
davidmehren: I'm pretty sure that this awaits `note.aliases` only once :smile:

src/media/media.service.spec.ts
```diff
@@ -222,9 +222,9 @@ describe('MediaService', () => {
         const mockMediaUploadEntry = {
           id: 'testMediaUpload',
           backendData: 'testBackendData',
-          user: {
+          user: Promise.resolve({
             username: 'hardcoded',
```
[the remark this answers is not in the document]
davidmehren: done

DerMolly (approved):

PR #1844, opened 2021-11-21
Add permissions guard and refactor controller methods accordingly

DerMolly:
### Component/Part
public and private api

### Description
This PR adds a permissions guard and refactors controller methods accordingly

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### DIFF
--- a/src/api/utils/permissions.guard.ts
+++ b/src/api/utils/permissions.guard.ts
@@ -1,0 +1,66 @@
+/*
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+import { CanActivate, ExecutionContext, Injectable } from '@nestjs/common';
+import { Reflector } from '@nestjs/core';
+import { Request } from 'express';
+
+import { ConsoleLoggerService } from '../../logger/console-logger.service';
+import { NotesService } from '../../notes/notes.service';
+import { Permission } from '../../permissions/permissions.enum';
+import { PermissionsService } from '../../permissions/permissions.service';
+import { User } from '../../users/user.entity';
+import { getNote } from './get-note.pipe';
+
+/**
+ * This guards controller methods from access, if the user has not the appropriate permissions.
+ * The permissions are set via the {@link Permissions} decorator in addition to this guard.
+ */
+@Injectable()
+export class PermissionsGuard implements CanActivate {
+  constructor(
+    private readonly logger: ConsoleLoggerService,
+    private reflector: Reflector,
+    private permissionsService: PermissionsService,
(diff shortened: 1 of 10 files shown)


DerMolly:
This has one very strang failing test and I'm not remotely sure why this happens.

InnayTool (approved):

src/api/utils/permissions.guard.ts
```diff
@@ -0,0 +1,60 @@
...
+    private permissionsService: PermissionsService,
+    private noteService: NotesService,
+  ) {}
+
+  async canActivate(context: ExecutionContext): Promise<boolean> {
+    const permissions = this.reflector.get<Permission[]>(
+      'permissions',
+      context.getHandler(),
+    );
+    // If no permissions are set this is probably an error and this guard should not let the request pass
```
davidmehren, line 34: Could we at least debug-log this case?
Might even warrant a higher log level, as hitting this should be a (severe?) bug.

src/permissions/permissions.decorator.ts
```diff
@@ -0,0 +1,17 @@
...
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+import { CustomDecorator, SetMetadata } from '@nestjs/common';
+
+import { Permission } from './permissions.enum';
+
+/**
+ * This decorator gathers the {@link Permission Permission} a user must hold for the {@link PermissionsGuard}
+ * @param permissions - an array of permissions. In praxis this should always contain exactly one {@link Permission}
```
davidmehren, line 12:
```suggestion
 * @param permissions - an array of permissions. In practice this should always contain exactly one {@link Permission}
```

davidmehren (changes requested):

src/api/utils/permissions.guard.ts
DerMolly: This should only ever occur, because someone broke the code. We could add log message "This should never happen. If you see this, please open an issue at https://github.com/hedgedoc/hedgedoc/issues". Do you think this would be appropriate?

src/api/utils/permissions.guard.ts
```diff
@@ -0,0 +1,66 @@
...
+    if (!permissions) {
+      this.logger.error(
+        'This should never happen. If you see this, please open an issue at https://github.com/hedgedoc/hedgedoc/issues',
```
davidmehren, line 40:
```suggestion
        'Could not find permission metadata. This should never happen. If you see this, please open an issue at https://github.com/hedgedoc/hedgedoc/issues',
        new Error().stack,
        'canActivate',
```

davidmehren (changes requested):

davidmehren (approved):

Issue #1881, opened 2021-12-09
Authenticated requests to the public API are too slow


davidmehren:
### Description

HedgeDoc 2 currently manages to process about 4 authenticated requests to the public API per second.

### Steps to reproduce

1. Benchmark requests to `/api/v2/me`
2. Observe that 1000 requests take about 260 seconds.

### Expected behaviour

Requests take, I don't know, less than 50 ms?

### Your Setup

- Host OS: Fedora 35
- NodeJS version: 17.2
- HedgeDoc version: `develop`

davidmehren:
Load testing with `artillery` results in
```
http.request_rate: .......................................... 10/sec
http.requests: .............................................. 600
http.codes.200: ............................................. 600
http.responses: ............................................. 600
http.response_time:
  min: ...................................................... 249
  max: ...................................................... 647
  median: ................................................... 347.3
  p95: ...................................................... 608
  p99: ...................................................... 632.8
  ```
  
  So we at least don't drop any requests at a rate of 10 per second. But 350 ms per API request is still too slow.

davidmehren:
According to my profiling, we seem to spend most of the time in `bcrypt`.
This is because we treat every API key as a password and check it with bcrypt. We should probably revisit that decision.

PR #1882, opened 2021-12-09
fix(auth): use sha-512 for auth tokens

davidmehren:
### Component/Part
Auth Tokens

### Description
Bcrypt hashes are too slow to be validated on every request.
As our tokens are random and have a fixed length, it is reasonable
to use SHA-512 instead.

SHA-512 is recommended by the BSI:
https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-1.pdf?__blob=publicationFile

Fixes https://github.com/hedgedoc/hedgedoc/issues/1881

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### DIFF
--- a/docs/content/dev/2.0.md
+++ b/docs/content/dev/2.0.md
@@ -39,3 +39,28 @@ Because we need to have empty constructors in our entity classes for TypeORM to 
 - Should either return a complete and fully useable instance or return a Pick/Omit type.
 - Exceptions to these rules are allowed, if they are mentioned in the method documentation
 
+## Auth tokens for the public API
+The public API uses bearer tokens for authentication.
+
+When a new token is requested via the private API, the backend generates a 64 bytes-long secret of
+cryptographically secure data and returns it as a base64url-encoded string, along with an identifier.
+That string can then be used by clients as a bearer token.
+
+A SHA-512 hash of the secret is stored in the database. To validate tokens, the backend computes the hash of the provided
+secret and checks it against the stored hash for the provided identifier.
+
+### Choosing a hash function
+Unfortunately, there does not seem to be any explicit documentation about our exact use-case.
+Most docs describe classic password-saving scenarios and recommend bcrypt, scrypt or argon2.
+These hashing functions are slow to stop brute-force or dictionary attacks, which would expose the original,
+user-provided password, that may have been reused across multiple services.
+
+We have a very different scenario:
+Our API tokens are 64 bytes of cryptographically strong pseudorandom data.
+Brute-force or dictionary attacks are therefore virtually impossible, and tokens are not reused across multiple services.
+We therefore need to only guard against one scenario:
+An attacker gains read-only access to the database. Saving only hashes in the database prevents the attacker
+from authenticating themselves as a user. The hash-function does not need to be very slow, 
+as the randomness of the original token prevents inverting the hash. The function actually needs to be reasonably fast,
+as the hash must be computed on every request to the public API.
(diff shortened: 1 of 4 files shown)


src/auth/auth.service.ts
```diff
@@ -70,9 +66,12 @@ export class AuthService {
         `User '${user.username}' has already 200 tokens and can't have anymore`,
       );
     }
-    const secret = bufferToBase64Url(randomBytes(54));
+    const secret = bufferToBase64Url(randomBytes(64));
```
DerMolly, line 69: This change is currently a bit unexplained. I'd suggest changing that.

DerMolly (approved):
LGTM

SISheogorath:
> SHA-512 is recommended by the BSI:
https://www.bsi.bund.de/SharedDocs/Downloads/EN/BSI/Publications/TechGuidelines/TG02102/BSI-TR-02102-1.pdf?__blob=publicationFile

Uhm, which section are you refering to? I don't think they talk about our use-case anywhere in that document. Section 6.1 recommends to do challenge-response for authentication with a shared secret. Otherwise it's quite silent on this part.

src/auth/auth.service.ts
```diff
@@ -70,9 +66,12 @@ export class AuthService {
...
     const keyId = bufferToBase64Url(randomBytes(8));
-    const accessToken = await hashPassword(secret);
+    const accessToken = crypto
```
InnayTool, line 71: I think that it should be named `accessTokenHash` to make clearer it is not the actual token.

InnayTool (changes requested):

davidmehren:
I updated the PR according to the reviews and expanded some comments

davidmehren:
Regarding the use of SHA-512:
Unfortunately, I could not find any explicit documentation about our exact use-case.
Most docs describe classic password-saving scenarios and recommend bcrypt, scrypt or argon2. These hashing functions are slow to stop brute-force or dictionary attacks, which would expose the original, user-provided password, which might have been reused across multiple services.

We have a very different scenario: Our API tokens are 64 bytes of cryptographically strong pseudorandom data. Brute-force or dictionary attacks are therefore virtually impossible, and they are not reused across multiple services. We therefore need to only guard against one scenario: An attacker gains read-only access to the database. Saving only hashes in the database prevents the attacker from authenticating themselves as a user. The hash-function does not need to be very slow, as the randomness of the original token prevents inverting the hash. The function actually needs to be reasonably fast, as the hash must be computed on every request to the public API. SHA-512 (or alternatively SHA3) fits this use-case.

SISheogorath:
@davidmehren 👍🏻 This sounds better :)

davidmehren:
@SISheogorath Great idea, I added the explanation to the docs.

docs/content/dev/2.0.md
```diff
@@ -39,3 +39,28 @@ Because we need to have empty constructors in our entity classes for TypeORM to
 - Should either return a complete and fully useable instance or return a Pick/Omit type.
 - Exceptions to these rules are allowed, if they are mentioned in the method documentation
 
+## Auth tokens for the public API
+The public uses bearer tokens for authentication.
```
InnayTool, line 43:
```suggestion
The public API uses bearer tokens for authentication.
```

InnayTool (changes requested):

InnayTool (approved):

PR #1935, opened 2021-12-28
Auth: Require old password for password change

ErikMichelson:
### Component/Part
Auth -> internal / "local"

### Description
This PR adds a check for the "old" password of the user when requesting a password change.
By checking the "old" password of the user prior to a password change, the password change function is more secured against abuse.

### Steps

- [x] Added implementation
- [x] Updated tests
  made sure that:

### Related Issue(s)
none

### DIFF
--- a/src/api/private/auth/auth.controller.ts
+++ b/src/api/private/auth/auth.controller.ts
@@ -9,10 +9,10 @@ import {
   ConflictException,
   Controller,
   Delete,
-  NotFoundException,
   Post,
   Put,
   Req,
+  UnauthorizedException,
   UseGuards,
 } from '@nestjs/common';
 import { Session } from 'express-session';
@@ -70,6 +70,10 @@ export class AuthController {
     @Body() changePasswordDto: UpdatePasswordDto,
   ): Promise<void> {
     try {
+      await this.identityService.loginWithLocalIdentity(
+        user,
+        changePasswordDto.currentPassword,
+      );
       await this.identityService.updateLocalPassword(
         user,
         changePasswordDto.newPassword,
@@ -77,7 +81,9 @@ export class AuthController {
       return;
     } catch (e) {
       if (e instanceof NotInDBError) {
-        throw new NotFoundException(e.message);
+        throw new UnauthorizedException(
+          'Verifying your identity with the current password did not work.',
+        );
       }
       throw e;
     }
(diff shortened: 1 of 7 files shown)


src/api/private/auth/auth.controller.ts
```diff
@@ -70,14 +70,20 @@ export class AuthController {
...
       await this.identityService.updateLocalPassword(
         user,
         changePasswordDto.newPassword,
       );
       return;
     } catch (e) {
       if (e instanceof NotInDBError) {
-        throw new NotFoundException(e.message);
+        throw new UnauthorizedException(
+          'Verifying your identity with the current password did not work.',
```
DerMolly, line 85: Isn't it bad practice to tell people that their problem lies with the password, even if we literally do that every time here even when the user does not exist?

DerMolly (changes requested):

src/api/private/auth/auth.controller.ts
ErikMichelson: No, this is the password change not the login.
The user must exist because it is retrieved from the session, that is checked by the guard.
The user also does not enter their username but only their current password and their new password. Therefore only the current password could be wrong.

ErikMichelson: One could argue however whether it is appropriate that the `identityService.loginWithLocalIdentity` method throws a `NotInDBError` in case of invalid username or password. A `InvalidCredentialsError` seems more meaningful to me.

DerMolly: I'd suggest adding an issue for this?

DerMolly (approved):

src/api/private/auth/auth.controller.ts
ErikMichelson: #1936

src/api/private/auth/auth.controller.ts
```diff
@@ -70,14 +70,20 @@ export class AuthController {
     @Body() changePasswordDto: UpdatePasswordDto,
   ): Promise<void> {
     try {
+      await this.identityService.loginWithLocalIdentity(
```
InnayTool, line 73: I think we should rename this function if we want to use it here, maybe something like `checkLocalPassword`?

davidmehren (changes requested):
Looks nice, thanks!

IMO, renaming `loginWithLocalIdentity` to `checkLocalPassword` is reasonable, as the function really does not 'log in' anybody, but only checks the password.

src/api/private/auth/auth.controller.ts
ErikMichelson: Done

davidmehren (approved):

InnayTool (approved):

Issue #1933, opened 2021-12-27
[1.9] Ignore health check endpoint logging


licy183:
### Which part of the project should be enhanced?

logger

### Is your enhancement request related to a problem? Please describe.

Hello, I upgraded hedgedoc to the latest version (v1.9.2), deployed it using Docker, and found that the health checking request was logging all the time. A lot of space will be used to save these useless logs.

```
2021-12-27T19:19:23.134Z info:    127.0.0.1 - - [27/Dec/2021:19:19:23 +0000] "GET /status HTTP/1.1" 200 268 "-" "hedgedoc-container-healthcheck/1.0"
2021-12-27T19:19:28.320Z info:    127.0.0.1 - - [27/Dec/2021:19:19:28 +0000] "GET /status HTTP/1.1" 200 268 "-" "hedgedoc-container-healthcheck/1.0"
```

### Describe the solution you'd like

Do not log the health check request.

### Describe alternatives you've considered

The [`skip`](https://github.com/expressjs/morgan#skip) option of library `morgan` may be helpful. 

We can detect this kind of request using `User-Agent`, or create a new route for health check.

Related issue: expressjs/morgan#119

DerMolly:
As an fixing PR was already merged, I think this issue can be closed…

PR #1938, opened 2021-12-28
Fix(logger): Ignore docker health check requests

licy183:
### Component/Part
logger

### Description
This patch disables logging of HTTP requests containing health-check UA from the local IP address. These logs can take up a lot of disk space but are of little use.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
#1933

### DIFF
--- a/app.js
+++ b/app.js
@@ -63,8 +63,14 @@ if (!config.useSSL && config.protocolUseSSL) {
   app.set('trust proxy', 1)
 }
 
+// check if the request is from container healthcheck
+function isContainerHealthCheck (req, _) {
+  return req.headers['user-agent'] === 'hedgedoc-container-healthcheck/1.0' && req.ip === '127.0.0.1'
+}
+
 // logger
 app.use(morgan('combined', {
+  skip: isContainerHealthCheck,
   stream: logger.stream
 }))
 


mrdrogdrog:
The request is always sent from 127.0.0.1. So why have you implemented all these edge cases?

licy183:
> The request is always sent from 127.0.0.1. So why have you implemented all these edge cases?

If the checker's client IP address is always 127.0.0.1, it is indeed unnecessary to implement these test cases. Should I submit a commit to delete these test cases and use 127.0.0.1 only?

app.js
```diff
@@ -65,6 +65,7 @@ if (!config.useSSL && config.protocolUseSSL) {
 
 // logger
 app.use(morgan('combined', {
+  skip: (req, _) => req.headers['user-agent'] === 'hedgedoc-container-healthcheck/1.0' && req.ip === '127.0.0.1',
```
mrdrogdrog, line 68: Please extract the check into a separate function. Like `IsRequestFromContainerHealthCheck()` or something.

mrdrogdrog (changes requested):

app.js
licy183: OK, I'll fix it.

mrdrogdrog (approved):

davidmehren (approved):
Thanks!

DerMolly (approved):

InnayTool (approved):

ErikMichelson (approved):

PR #2037, opened 2022-01-23
Add error mapping

### LINKED ISSUE #1640: Add exception filter for our custom errors
Exception handling

**Is your enhancement request related to a problem? Please describe.**
Having long `catch` blocks everywhere makes files unnecessarily long and verbose and duplicates code.

Implement a global [exception filter](https://docs.nestjs.com/exception-filters) that converts our custom errors like `AlreadyInDBError` to HTTP errors.

**Describe alternatives you've considered**
Continue having long `catch` blocks.

**Additional context**

### LINKED ISSUE #1983: We should not return Bad Request when an alias is already in use

Both APIs

Both the public and the private API return 400 Bad Request, when an alias is already used. This is a very generic error code.

The APIs should return 409 Conflict instead.

Not changing the API.

### Additional context

_No response_

DerMolly:
### Component/Part
public and private api

### Description
This PR adds an error mapping from internal HedgeDoc Error to HttpExceptions. While doing this I noticed that some error codes are not documented in the openapi docs and added the appropriate annotations. 

### Steps

- [x] Added implementation
  made sure that:

Fixes #1983
Fixes #1640

### DIFF
--- a/src/api/private/alias/alias.controller.ts
+++ b/src/api/private/alias/alias.controller.ts
@@ -9,7 +9,6 @@ import {
   Controller,
   Delete,
   HttpCode,
-  NotFoundException,
   Param,
   Post,
   Put,
@@ -19,11 +18,7 @@ import {
 import { ApiTags } from '@nestjs/swagger';
 
 import {
-  AlreadyInDBError,
-  ForbiddenIdError,
-  NotInDBError,
-  PrimaryAliasDeletionForbiddenError,
-} from '../../../errors/errors';
+
 import { SessionGuard } from '../../../identity/session.guard';
 import { ConsoleLoggerService } from '../../../logger/console-logger.service';
 import { AliasCreateDto } from '../../../notes/alias-create.dto';
@@ -49,32 +44,23 @@ export class AliasController {
   ) {
     this.logger.setContext(AliasController.name);
   }
+
   @Post()
   async addAlias(
     @RequestUser() user: User,
     @Body() newAliasDto: AliasCreateDto,
   ): Promise<AliasDto> {
-    try {
-      const note = await this.noteService.getNoteByIdOrAlias(
-        newAliasDto.noteIdOrAlias,
-      );
-      if (!(await this.permissionsService.isOwner(user, note))) {
-        throw new UnauthorizedException('Reading note denied!');
-      }
-      const updatedAlias = await this.aliasService.addAlias(
-        note,
-        newAliasDto.newAlias,
-      );
-      return this.aliasService.toAliasDto(updatedAlias, note);
-    } catch (e) {
-      if (e instanceof AlreadyInDBError) {
-        throw new BadRequestException(e.message);
-      }
-      if (e instanceof ForbiddenIdError) {
-        throw new BadRequestException(e.message);
-      }
-      throw e;
+    const note = await this.noteService.getNoteByIdOrAlias(
+      newAliasDto.noteIdOrAlias,
+    );
+    if (!(await this.permissionsService.isOwner(user, note))) {
+      throw new UnauthorizedException('Reading note denied!');
     }
+    const updatedAlias = await this.aliasService.addAlias(
+      note,
+      newAliasDto.newAlias,
+    );
+    return this.aliasService.toAliasDto(updatedAlias, note);
   }
 
   @Put(':alias')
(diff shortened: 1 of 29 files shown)


test/private-api/alias.e2e-spec.ts
```diff
@@ -83,7 +83,7 @@ describe('Alias', () => {
           .post(`/api/private/alias`)
           .set('Content-Type', 'application/json')
           .send(newAliasDto)
-          .expect(400)
+          .expect(409)
```
davidmehren, line 86: Hmmm, actually, `Bad Request` seems to be a better error message for a forbidden note ID than `Conflict`. The best error would probably be `Forbidden` in this case.

test/private-api/alias.e2e-spec.ts
```diff
@@ -156,7 +156,7 @@ describe('Alias', () => {
           .put(`/api/private/alias/${forbiddenNoteId}`)
           .set('Content-Type', 'application/json')
           .send(changeAliasDto)
-          .expect(400)
+          .expect(404)
```
davidmehren, line 159: 404 when requesting a forbidden ID is also weird. IMO it should return `Forbidden`.

src/errors/error-mapping.ts
```diff
@@ -0,0 +1,90 @@
...
+      'TokenNotValidError',
+      (object): HttpException => new UnauthorizedException(object),
+    ],
+    [
+      'TooManyTokensError',
+      (object): HttpException => new BadRequestException(object),
+    ],
+    [
+      'PermissionsUpdateInconsistentError',
+      (object): HttpException => new UnprocessableEntityException(object),
```
davidmehren, line 51: According to Wikipedia, HTTP 422 is WebDAV-specific

davidmehren (changes requested):

src/errors/error-mapping.ts
ErikMichelson: Wikipedia is right. HTTP 422 is defined in [RFC 4918](https://datatracker.ietf.org/doc/html/rfc4918#section-11.2) which is the RFC defining WebDAV.

test/private-api/alias.e2e-spec.ts
ErikMichelson: The question is: Do we want to let the end-user know which alias is forbidden?
In this case, I'd say okay 403 Forbidden is fitting.

But if we want no distinguishing between forbidden aliases and already taken aliases, it makes sense to map both to 409 Conflict.

test/private-api/alias.e2e-spec.ts
ErikMichelson: In this case 403 Forbidden seems right.

But speaking in strict terms it's "updating" and not "requesting" a forbidden alias.
For *requesting* a forbidden alias, 404 Not Found seems right.

src/api/private/alias/alias.controller.ts
```diff
@@ -49,35 +56,32 @@ export class AliasController {
   ) {
     this.logger.setContext(AliasController.name);
   }
+
   @Post()
+  @ApiConflictResponse({ description: conflictDescription })
+  @ApiUnauthorizedResponse({ description: unauthorizedDescription })
+  @ApiNotFoundResponse({ description: notFoundDescription })
```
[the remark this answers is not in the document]
DerMolly: But `FullApi` is
```
ApiForbiddenResponse({ description: forbiddenDescription }),
ApiNotFoundResponse({ description: notFoundDescription }),
ApiUnauthorizedResponse({ description: forbiddenDescription }),
```

src/api/private/me/history/history.controller.ts
```diff
@@ -26,11 +29,16 @@ import { SessionGuard } from '../../../../identity/session.guard';
...
+import {
+  notFoundDescription,
+  unauthorizedDescription,
+} from '../../../utils/descriptions';
 import { GetNoteInterceptor } from '../../../utils/get-note.interceptor';
 import { RequestNote } from '../../../utils/request-note.decorator';
 import { RequestUser } from '../../../utils/request-user.decorator';
 
 @UseGuards(SessionGuard)
+@UseFilters(ErrorExceptionMapping)
```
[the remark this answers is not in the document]
DerMolly: It wasn't always. This is remian of that. Fixed

src/errors/error-mapping.ts
DerMolly: Well that's not good then, should we replace it with `BadRequest` or something else?

src/api/private/alias/alias.controller.ts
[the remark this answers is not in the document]
davidmehren: Right. What I'd like is to merge each regularly used combination of individual decorators into a new one to reduce clutter.

src/errors/error-mapping.ts
DerMolly: `BadRequest` will be used

src/api/private/me/history/history.controller.ts
```diff
@@ -42,84 +50,57 @@ export class HistoryController {
   }
 
   @Get()
+  @ApiUnauthorizedResponse({ description: unauthorizedDescription })
+  @ApiNotFoundResponse({ description: notFoundDescription })
```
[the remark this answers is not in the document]
DerMolly: This will be fixed in a further PR

davidmehren (approved):

src/api/private/auth/auth.controller.ts
```diff
@@ -49,55 +52,38 @@ export class AuthController {
 
   @UseGuards(RegistrationEnabledGuard)
```
ErikMichelson, line 53: This guard can throw a BadRequestException which should get an ApiBadRequestResponse decorator.

DerMolly:
Annotations will be improved by #2076

src/api/private/auth/auth.controller.ts
DerMolly: fixed

ErikMichelson (approved):

InnayTool (approved):

PR #2128, opened 2022-02-23
Add Traefik as reversy proxy

crstian19:
Signed-off-by: Crstian19 <kiyoma19@gmail.com>

### Component/Part
Docs

### Description
Added to docs Traefik config as reverse proxy.

### Steps

- [ ] Added implementation
- [ ] Added / updated tests
- [ ] Added changelog snippet
  made sure that:

### DIFF
--- a/docs/content/guides/reverse-proxy.md
+++ b/docs/content/guides/reverse-proxy.md
@@ -114,3 +114,25 @@ Here is an example config snippet:
   Include /etc/letsencrypt/options-ssl-apache.conf
 </VirtualHost>
 ```
+### Traefik
+
+In traefik we have to add the traefik labels and it is also very important: add our domain to "CMD_DOMAIN", "CMD_URL_ADDPORT=false", "CMD_PROTOCOL_USESSL=true".
+
+``` yaml
+  app:
+    # Make sure to use the latest release from https://hedgedoc.org/latest-release
+    image: quay.io/hedgedoc/hedgedoc:1.9.2
+    environment:
+      - CMD_DB_URL=postgres://hedgedoc:password@database:5432/hedgedoc
+      - CMD_DOMAIN=yourdomain.com
+      - CMD_URL_ADDPORT=false
+      - CMD_PROTOCOL_USESSL=true
+    volumes:
+      - uploads:/hedgedoc/public/uploads
+    labels:
+      - traefik.enable=true
+      - traefik.http.routers.hedgedoc.entryPoints=web-secure
+      - traefik.http.routers.hedgedoc.rule=Host(`yourdomain.com`)
+      - traefik.http.routers.hedgedoc.tls.certresolver=default
+      - traefik.http.services.hedgedoc.loadbalancer.server.port=3000
+```


DerMolly (approved):
Thanks @crstian19!

This looks good to me, but maybe someone with more traefik experience want to have a look at that, too.

mrdrogdrog:
@DerMolly sure, I'll take a look later

docs/content/guides/reverse-proxy.md
```diff
@@ -114,3 +114,25 @@ Here is an example config snippet:
   Include /etc/letsencrypt/options-ssl-apache.conf
 </VirtualHost>
 ```
+### Traefik
+
+In traefik we have to add the traefik labels and it is also very important: add our domain to "CMD_DOMAIN", "CMD_URL_ADDPORT=false", "CMD_PROTOCOL_USESSL=true".
```
mrdrogdrog, line 119: - We don't use "we" in the docs. We use "you".
- `add our domain to "CMD_DOMAIN", "CMD_URL_ADDPORT=false", "CMD_PROTOCOL_USESSL=true".` This sentence alone doesn't make much sense here. Where should it be added? To the hedgedoc config? If so, then it's not necessary because that is already covered by e.g. https://docs.hedgedoc.org/guides/reverse-proxy/#configure-asset-link-generation.

docs/content/guides/reverse-proxy.md
```diff
@@ -114,3 +114,25 @@ Here is an example config snippet:
...
+
+``` yaml
```
mrdrogdrog, line 121: Please note that this is only an example. Maybe even remove the non-traefik related stuff here and add a placeholder like `[...]`

docs/content/guides/reverse-proxy.md
```diff
@@ -114,3 +114,25 @@ Here is an example config snippet:
...
+      - CMD_DOMAIN=yourdomain.com
+      - CMD_URL_ADDPORT=false
+      - CMD_PROTOCOL_USESSL=true
+    volumes:
+      - uploads:/hedgedoc/public/uploads
+    labels:
+      - traefik.enable=true
+      - traefik.http.routers.hedgedoc.entryPoints=web-secure
+      - traefik.http.routers.hedgedoc.rule=Host(`yourdomain.com`)
+      - traefik.http.routers.hedgedoc.tls.certresolver=default
```
mrdrogdrog, line 136: Please add a warning before the code block that explains that the entrypoint name and certresolver name are examples and should be replaced with the user's own values.

docs/content/guides/reverse-proxy.md
mrdrogdrog, line 119: Which traefik labels? All of them? Please write down the absolute necessary labels that are needed to make hedgedoc work with hedgedoc.

mrdrogdrog (changes requested):

ErikMichelson:
Hi @crstian19,

do you still want to work on this?

crstian19:
> Hi @crstian19,
> 
> do you still want to work on this?

Yes, but I haven't had time because of my work, to make the necessary modifications, as soon as I can I will make them.

docs/content/guides/reverse-proxy.md
```diff
@@ -114,3 +114,25 @@ Here is an example config snippet:
...
+In traefik we have to add the traefik labels and it is also very important: add our domain to "CMD_DOMAIN", "CMD_URL_ADDPORT=false", "CMD_PROTOCOL_USESSL=true".
+
+``` yaml
+  app:
+    # Make sure to use the latest release from https://hedgedoc.org/latest-release
+    image: quay.io/hedgedoc/hedgedoc:1.9.2
+    environment:
+      - CMD_DB_URL=postgres://hedgedoc:password@database:5432/hedgedoc
+      - CMD_DOMAIN=yourdomain.com
+      - CMD_URL_ADDPORT=false
```
funkyfuture, line 128: in my setup this isn't required. i'd stick to what is necessary in an example.

docs/content/guides/reverse-proxy.md
funkyfuture: both values might often be not necessary at all to define in a specific service.

docs/content/guides/reverse-proxy.md
mrdrogdrog: CMD_DOMAIN and CMD_URL_ADDPORT are necessary if you have a reverse proxy.
https://docs.hedgedoc.org/guides/reverse-proxy/#configure-asset-link-generation

funkyfuture: at least not in practice, this is the full extend of my configuration that runs for quiet a while now:

```yaml
services:
  app:
    environment:
      CMD_ALLOW_EMAIL_REGISTER: "false"
      CMD_DB_URL: postgres://hedgedoc:hedgedoc@db:5432/hedgedoc
      CMD_DOMAIN: ${DOMAIN:?err}
      CMD_PROTOCOL_USESSL: "true"
```

PR #2244, opened 2022-04-03
Fix GitLab snippet export

### LINKED ISSUE #2208: Gitlab snippet export doesn't work
### Description

Gitlab snippet export doesn't work with the current API ->  "files, content are missing, exactly one parameter must be provided"

### Steps to reproduce

1. Install Gitlab Version 14.9.1 CE
2. Connect the hedgedoc application
3. Test Gitlab snippet export

### Expected behaviour

The snippet is transferred.

### Logs

```shell
400
Bad Request
{"error":"files, content are missing, exactly one parameter must be provided"}
```

### Config

_No response_

### Your Setup

- Gitlab Version 14.9.1 CE
- linuxserver/hedgedoc Version 1.9.2

### Additional context

I think in the current API you have to specify **files** in the req-body, **file_name** is not sufficient.
https://docs.gitlab.com/ee/api/snippets.html
An alternative or additional feature would be to push the file directly into a repository 
(however, a branch and commit message is required here). https://docs.gitlab.com/ee/api/repository_files.html

ErikMichelson:
### Component/Part
editor -> GitLab snippet export

### Description
This PR fixes the GitLab snippet export.
The snippet export broke due to two reasons.
First of all, the request to GitLab fail in the default configuration due to the CSP not being set properly. This commit adds the configured GitLab base url to the connect-src directives. The second problem is a change in the GitLab API spec. Instead of `code` and `file_name` the GitLab API now requires an `files` array with `content` and `file_path` entries per snippet.

### Steps

- [x] Added implementation
- [x] Added / updated tests
- [x] Added changelog snippet
  made sure that:

### Related Issue(s)
Fixes #2208

### DIFF
--- a/public/js/index.js
+++ b/public/js/index.js
@@ -1918,6 +1918,8 @@ $('#snippetExportModalConfirm').click(function () {
   const accesstoken = $('#snippetExportModalAccessToken').val()
   const baseURL = $('#snippetExportModalBaseURL').val()
   const version = $('#snippetExportModalVersion').val()
+  const projectId = $('#snippetExportModalProjects').val()
+  const visibilityValue = $('#snippetExportModalVisibility').val()
 
   const data = {
     title: $('#snippetExportModalTitle').val(),
@@ -1928,9 +1930,9 @@ $('#snippetExportModalConfirm').click(function () {
       }
     ],
     visibility:
-      $('#snippetExportModalVisibility').val() === '0'
+      visibilityValue === '0'
         ? 'private'
-        : $('#snippetExportModalVisibility').val() === '10'
+        : visibilityValue === '10'
           ? 'internal'
           : 'private'
   }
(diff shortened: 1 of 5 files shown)


public/js/index.js
```diff
@@ -1933,40 +1936,27 @@ $('#snippetExportModalConfirm').click(function () {
 
   if (
     !data.title ||
-    !data.file_name ||
-    !data.code ||
-    !data.visibility_level ||
+    !data.files[0].file_path ||
+    !data.files[0].content ||
     !$('#snippetExportModalProjects').val()
```
mrdrogdrog, line 1941: You execute `$('#snippetExportModalProjects').val()` two times. Here and in L1944.

mrdrogdrog (changes requested):

davidmehren (approved):

mrdrogdrog (approved):

Issue #2248, opened 2022-04-08
Inline authorship coloration doesn't work - hex2rgb missing


EBendinelli:
### Description

The authorship coloration feature (underline in different color per author in the same paragraph) doesn't work

### Steps to reproduce

1. Create new pad
2. Author 1 writes a paragraph 
3. Author 2 opens pad and writes something within the same paragraph

### Expected behaviour

Author 2 modifications to the paragraph should be underlined with the author colour. We recently migrated from codiMD and that's the behaviour we were getting. 

Expected behaviour:
![image](https://user-images.githubusercontent.com/12803082/162405453-dc798e09-10f5-45ed-bbaf-8b77c79ef84f.png)

Observed behaviour:
![image](https://user-images.githubusercontent.com/12803082/162405830-d4ab51a1-89e3-4dec-abf7-7ca25c0b4bdc.png)

Error message:
![image](https://user-images.githubusercontent.com/12803082/162405169-e3ec8ee3-8039-4c8a-a5c7-51004dcc804c.png)

### Logs

```shell
Browser consols throw this error message: Uncaught ReferenceError: hex2rgb is not defined
```

### Your Setup

- Host OS: Alpine 3.15.2
- NodeJS version: v16.14.2
- HedgeDoc version: 1.9.2

### Additional context

Hedgedoc was installed manually. There doesn't seem to be any reference to hex2rbg in Hedgedoc codebase. CodiMD has three references to it, including a dedicated function with that name: https://github.com/hackmdio/codimd/search?q=hex2rgb

davidmehren:
Thanks for reporting this!

We actually include the code (https://github.com/hedgedoc/hedgedoc/blob/master/public/vendor/ot/codemirror-adapter.js#L381), but have a bug in our build scripts, so it doesn't get loaded properly.

davidmehren:
A fix is merged and will be contained in the next release

EBendinelli:
Thanks a lot for the quick fix! Looking forward to the next release

mrdrogdrog:
@EBendinelli This fix is included in the release we made yesterday 😉

PR #2249, opened 2022-04-08
Fix missing inline authorship colors

davidmehren:
### Component/Part
Frontend JS

### Description
The `hex2rgb` function seems to previously have been available globally.
It probably got lost in the great Webpack refactoring and nobody noticed
 that.

This copies the function into its own file (to make importing it easy)
 and adds an import in index.js.

Fixes https://github.com/hedgedoc/hedgedoc/issues/2248

### Steps

- [x] Added implementation
- [x] Added changelog snippet
  made sure that:

### Related Issue(s)
(https://github.com/hedgedoc/hedgedoc/issues/2248)

### DIFF
--- a/public/.eslintrc.js
+++ b/public/.eslintrc.js
@@ -20,7 +20,6 @@ module.exports = {
     FilePicker: false,
     ot: false,
     MediaUploader: false,
-    hex2rgb: false,
     num_loaded: false,
     Visibility: false,
     inlineAttachment: false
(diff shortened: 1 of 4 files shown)


mrdrogdrog (approved):

ErikMichelson (approved):
Inline authorship colors work again with this fix.

However if this function should be used somewhere else too, we should be careful as it does not accept HEX colors with alpha channel.

PR #2278, opened 2022-04-18
Actually use configured database

davidmehren:
### Component/Part
AppModule, logging, db config

### Description
This PR
- Uses the database from the config in `AppModule`
- Adapts our config options to follow TypeORM terminology
- Adds a logging adapter for TypeORM, which logs queries on the DEBUG level

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/src/config/mock/database.config.mock.ts
+++ b/src/config/mock/database.config.mock.ts
@@ -5,18 +5,18 @@
  */
 import { registerAs } from '@nestjs/config';
 
-import { DatabaseDialect } from '../database-dialect.enum';
+import { DatabaseType } from '../database-type.enum';
 import { DatabaseConfig } from '../database.config';
 
 export default registerAs(
   'databaseConfig',
   (): DatabaseConfig => ({
-    dialect: (process.env.HEDGEDOC_TEST_DB_TYPE || 'sqlite') as DatabaseDialect,
+    type: (process.env.HEDGEDOC_TEST_DB_TYPE ||
+      DatabaseType.SQLITE) as DatabaseType,
     database: 'hedgedoc',
     password: 'hedgedoc',
     host: 'localhost',
     port: 0,
-    storage: '',
     username: 'hedgedoc',
   }),
 );
--- a/src/logger/console-logger.service.ts
+++ b/src/logger/console-logger.service.ts
@@ -23,6 +23,7 @@ import DateTimeFormatOptions = Intl.DateTimeFormatOptions;
 export class ConsoleLoggerService implements LoggerService {
   private classContext: string | undefined;
   private lastTimestamp: number;
+  private skipColor = false;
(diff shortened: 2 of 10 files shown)


src/config/mock/database.config.mock.ts
```diff
@@ -5,18 +5,17 @@
...
 
-import { DatabaseDialect } from '../database-dialect.enum';
+import { DatabaseType } from '../database-dialect.enum';
 import { DatabaseConfig } from '../database.config';
 
 export default registerAs(
   'databaseConfig',
   (): DatabaseConfig => ({
-    dialect: (process.env.HEDGEDOC_TEST_DB_TYPE || 'sqlite') as DatabaseDialect,
+    type: (process.env.HEDGEDOC_TEST_DB_TYPE || 'sqlite') as DatabaseType,
```
DerMolly, line 14:
```suggestion
    type: (process.env.HEDGEDOC_TEST_DB_TYPE || DatabaseType.SQLITE) as DatabaseType,
```

src/config/database-dialect.enum.ts
```diff
@@ -4,7 +4,7 @@
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 
-export enum DatabaseDialect {
+export enum DatabaseType {
```
DerMolly, line 7: the file should be renamed, too…

src/logger/console-logger.service.ts
```diff
@@ -118,12 +123,22 @@ export class ConsoleLoggerService implements LoggerService {
     isTimeDiffEnabled?: boolean,
   ): void {
     let output;
-    if (isObject(message)) {
-      output = `${color('Object:')}\n${ConsoleLoggerService.sanitize(
-        JSON.stringify(message, null, 2),
-      )}\n`;
+
+    // if skipColor is set, we do not use colors and skip sanitizing
```
DerMolly, line 127: Maybe I'm not fully understanding what this is supposed to do, but this seems to me more lika a `skipSanetization` instead of a `skipColor`.

DerMolly (changes requested):
Maybe we need to remove the `sanetize` method from the Logger class alltogether and instead use it as an helper anywhere we put user-input (and maybe stuff from the DB) in logs.
Probably a good rule of thumb would be any non-static strings.

src/logger/console-logger.service.ts
davidmehren: The (current) main use of `skipColor` is to disable the coloration of log messages by the ConsoleLogger, to allow the caller to provide its own color. For the latter, sanitization must also be disabled, as that would strip the color escape-codes.

InnayTool (approved):

DerMolly (approved):

PR #2315, opened 2022-05-04
[1.x] fix(migrations): Remove unexpected shell call

SISheogorath:
### Component/Part
migrations

### Description
This patch removes the call of `/usr/bin/env` when calling the migration
script in favour of using the processes own nodejs invocation path.

This should drop the requirement for `/usr/bin/env` to exist on a
system/in a container that runs hedgedoc.

### Steps

- [x] Added implementation
- [x] Added changelog snippet
  made sure that:

### Related Issue(s)
Required for https://github.com/hedgedoc/container/pull/313

### DIFF
--- a/lib/migrations/20200321153000-fix-account-deletion.js
+++ b/lib/migrations/20200321153000-fix-account-deletion.js
@@ -3,7 +3,7 @@ const { spawnSync } = require('child_process')
 const path = require('path')
 module.exports = {
   up: function (queryInterface, Sequelize) {
-    const cleanup = spawnSync('./bin/cleanup', { cwd: path.resolve(__dirname, '../../') })
+    const cleanup = spawnSync(process.argv[0], ['./bin/cleanup'], { cwd: path.resolve(__dirname, '../../') })
     if (cleanup.status !== 0) {
       throw new Error('Unable to cleanup')
     }
--- a/public/docs/release-notes.md
+++ b/public/docs/release-notes.md
@@ -2,6 +2,9 @@
 
 ## <i class="fa fa-tag"></i> 1.x.x <i class="fa fa-calendar-o"></i> UNRELEASED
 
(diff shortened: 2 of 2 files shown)


lib/migrations/20200321153000-fix-account-deletion.js
```diff
@@ -3,7 +3,7 @@ const { spawnSync } = require('child_process')
 const path = require('path')
 module.exports = {
   up: function (queryInterface, Sequelize) {
-    const cleanup = spawnSync('./bin/cleanup', { cwd: path.resolve(__dirname, '../../') })
+    const cleanup = spawnSync(process.argv[0], ['./bin/cleanup'], { cwd: path.resolve(__dirname, '../../') })
```
ErikMichelson, line 6: Is this secure to rely on `process.argv[0]`? Or better said: Why not hardcode `node` directly?
If I remember correctly there was once a security problem where argv[0] could be overridden.

SISheogorath: Fair point, it could be overwritten. The problem with putting `"node"` there, is that it assumes that PATH evaluation is done and that the binary is actually called `node` and not `nodejs` or `node16`. On the other hand, this wasn't handled by the currently used script either, so I guess it's fair to try 🤔

On the other hand, this is called before any user-workload is processed, therefore if this was manipulated, it would have happened in our own code or one of our dependencies, which already has the ability to run code anyway 👀

ErikMichelson: >  if this was manipulated, it would have happened in our own code or one of our dependencies, which already has the ability to run code anyway

Well that's true, then probably my point is negligible.

DerMolly (approved):

ErikMichelson (approved):

Issue #1594, opened 2021-08-29
Media APIs: Implement a decorator to convert header to Note instance


davidmehren:
**Which part of the project should be enhanced?**
The `MediaController` in both the public and private APIs

**Describe the solution you'd like**
When uploading media, the ID of the note the upload belongs to is transferred in a header.
Currently, the business logic uses `await this.noteService.getNoteByIdOrAlias(noteId);` to create a Note instance from that. We should implement a custom parameter decorator that uses the `GetNotePipe` to get the Note instance.

**Additional context**
https://stackoverflow.com/a/55561022

atkac:
Hello @davidmehren , I'd like to work on this issue.

davidmehren:
This was implemented by @lautaroalvarez 🎉

PR #2383, opened 2022-05-25
refactor(media-apis): Implement a decorator to convert header to Note instance

lautaroalvarez:
### Component/Part
Media APIs routes (both private and public)

### Description
This PR is a refactor in the upload file routes. I have followed the issue comments (https://github.com/hedgedoc/hedgedoc/issues/1594#issuecomment-942629384) so I added a guard to find the related note to a new uploaded file and call it in both routes.

Its my first contribution in this project so please help me if I am doing something in a wrong way.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
https://github.com/hedgedoc/hedgedoc/issues/1594

### DIFF
--- a/src/api/private/media/media.controller.ts
+++ b/src/api/private/media/media.controller.ts
@@ -6,7 +6,6 @@
 import {
   Controller,
   Delete,
-  Headers,
   Param,
   Post,
   UploadedFile,
@@ -25,7 +24,9 @@ import { MulterFile } from '../../../media/multer-file.interface';
 import { Note } from '../../../notes/note.entity';
 import { NotesService } from '../../../notes/notes.service';
 import { User } from '../../../users/user.entity';
+import { NoteHeaderInterceptor } from '../../utils/note-header.interceptor';
 import { OpenApi } from '../../utils/openapi.decorator';
+import { RequestNote } from '../../utils/request-note.decorator';
 import { RequestUser } from '../../utils/request-user.decorator';
 
 @UseGuards(SessionGuard)
@@ -59,6 +60,7 @@ export class MediaController {
     description: 'ID or alias of the parent note',
   })
   @UseInterceptors(FileInterceptor('file'))
+  @UseInterceptors(NoteHeaderInterceptor)
   @OpenApi(
     {
       code: 201,
@@ -72,13 +74,11 @@ export class MediaController {
   )
   async uploadMedia(
     @UploadedFile() file: MulterFile,
-    @Headers('HedgeDoc-Note') noteId: string,
+    @RequestNote() note: Note,
     @RequestUser() user: User,
   ): Promise<MediaUploadDto> {
-    // TODO: Move getting the Note object into a decorator
-    const note: Note = await this.noteService.getNoteByIdOrAlias(noteId);
     this.logger.debug(
-      `Recieved filename '${file.originalname}' for note '${noteId}' from user '${user.username}'`,
+      `Recieved filename '${file.originalname}' for note '${note.id}' from user '${user.username}'`,
       'uploadMedia',
     );
     const upload = await this.mediaService.saveFile(file.buffer, user, note);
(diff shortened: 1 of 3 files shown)


src/api/utils/note-header.guard.ts
```diff
@@ -0,0 +1,33 @@
...
+  constructor(
+    private readonly logger: ConsoleLoggerService,
+    private noteService: NotesService,
+  ) {
+    this.logger.setContext(NoteHeaderGuard.name);
+  }
+
+  async canActivate(context: ExecutionContext): Promise<boolean> {
+    const request: Request & {
+      headers: { 'hedgedoc-note': string }; // eslint-disable-line
```
ErikMichelson, line 26: I'm not sure whether it makes a difference if the header is written lowercase here, where it was written as `HedgeDoc-Note` previously.

@davidmehren

lautaroalvarez: Inside `request.headers` all header names are in lowercase. I don't know if we can use the `@Headers()` function from nest, Its my first time with nest and I didn't find it on documentation.

I have to add this "eslint disable" because eslint doesn't like keys with "-", I try using `hedgedocNote` but it doesn't work.

davidmehren: Header names are always lowercased according to the [node docs](https://nodejs.org/api/http.html#messageheaders), so that should be alright.

src/api/utils/note-header.guard.ts
```diff
@@ -0,0 +1,33 @@
...
+
+import { ConsoleLoggerService } from '../../logger/console-logger.service';
+import { Note } from '../../notes/note.entity';
+import { NotesService } from '../../notes/notes.service';
+
+/**
+ * This guard take the note's id in header and find the Note in database.
+ */
+@Injectable()
+export class NoteHeaderGuard implements CanActivate {
```
davidmehren, line 16: This should probably be an Interceptor instead of a Guard. Guards decide if a request is allowed to proceed, but this guard always returns `true` after manipulating the request object.

You can take a look at this interceptor we already implemented: https://github.com/hedgedoc/hedgedoc/blob/develop/src/api/utils/get-note.interceptor.ts

src/api/utils/note-header.guard.ts
```diff
@@ -0,0 +1,33 @@
+/*
+ * SPDX-FileCopyrightText: 2021 The HedgeDoc developers (see AUTHORS file)
```
davidmehren, line 2:
```suggestion
 * SPDX-FileCopyrightText: 2022 The HedgeDoc developers (see AUTHORS file)
```

davidmehren (changes requested):

src/api/utils/note-header.guard.ts
lautaroalvarez: Sure. I will change it to Interceptor

src/api/private/media/media.controller.ts
```diff
@@ -60,7 +60,7 @@ export class MediaController {
     description: 'ID or alias of the parent note',
   })
   @UseInterceptors(FileInterceptor('file'))
```
DerMolly, line 62:
```suggestion
  @UseInterceptors(FileInterceptor('file'), NoteHeaderInterceptor)
```

src/api/public/media/media.controller.ts
```diff
@@ -78,7 +78,7 @@ export class MediaController {
     500,
   )
   @UseInterceptors(FileInterceptor('file'))
```
DerMolly, line 80:
```suggestion
  @UseInterceptors(FileInterceptor('file'), NoteHeaderInterceptor)
```

DerMolly (changes requested):
You can bind multiple interceptors with one `@UseInterceptors`

DerMolly:
Currently the lint job is failing, @lautaroalvarez. This can be fixed by running `yarn lint:fix`.

davidmehren (approved):

davidmehren:
The CI fails for unrelated reasons, so if everybody else is happy, I'll force-merge this.

InnayTool (approved):

DerMolly (approved):

PR #2519, opened 2022-07-18
doc: fix socket.io reverse-proxy

Stunkymonkey:
### Description
I tried it without and it did not work. So I guess the documentation has a little problem.

### Steps

- [ ] Added implementation
- [ ] Added / updated tests
- [ ] Added changelog snippet
  made sure that:

### DIFF
--- a/docs/content/guides/reverse-proxy.md
+++ b/docs/content/guides/reverse-proxy.md
@@ -64,7 +64,7 @@ server {
         server_name hedgedoc.example.com;
 
         location / {
-                proxy_pass http://127.0.0.1:3000;
+                proxy_pass http://127.0.0.1:3000/;
                 proxy_set_header Host $host; 
                 proxy_set_header X-Real-IP $remote_addr; 
                 proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for; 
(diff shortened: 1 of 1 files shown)


davidmehren:
The [relevant nginx docs](https://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_pass) say the following:

> If the proxy_pass directive is specified with a URI, then when a request is passed to the server, the part of a [normalized](https://nginx.org/en/docs/http/ngx_http_core_module.html#location) request URI matching the location is replaced by a URI specified in the directive:

In our case, the URI specified in the `location` is exactly the same as in the proxy_pass directive. Adding the paths to the `proxy_pass` should therefore have no effect.

Stunkymonkey:
this was quite some time ago. So I mostly forgot why.
I can only suggest you try it yourself. But using nginx with NixOS i can report this is the config that worked for me.
If you know it is not working please close the PR

davidmehren:
> If you know it is not working please close the PR

Well, I don't *know* that your changes don't work. But the current config does work for me and, according to the nginx docs, your changes should not change any behavior.

I'm therefore going to close this for now, if anybody else finds the current nginx config nonworking, we can revisit this.

Issue #1984, opened 2022-01-16
Realtime messages


ErikMichelson:
## Scope
Realtime API over WebSocket

The following message types need to be implemented as messages over the HEDGEDOC channel of the WebSocket.

## Required messages
- [ ] Note content updates (via yjs) #1780 
- [ ] Permission updates
- [ ] Note deletion (to inform currently editing users)
- [ ] Server version upgrade (to show reload for new features alert)

InnayTool:
Should we close this issue because realtime is mainly implemented?

mrdrogdrog:
The pr that implements the messages isn't merged yet.

PR #2638, opened 2022-09-04
feat: add realtime announcements for permission changes and note deletion

DerMolly:
### Component/Part
realtime communication

### Description
This PR adds realtime announcements for permission changes and note deletion.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
fixes https://github.com/hedgedoc/hedgedoc/issues/1984


DerMolly:
This should only be reviewed after https://github.com/hedgedoc/hedgedoc/pull/2637 is merged…

InnayTool (approved):

davidmehren (approved):

InnayTool (approved):

mrdrogdrog:
I don't like that you're adding another forward reference between the services to notify the realtime users. Not because of THIS reference but because this won't be the last time we'll implement a notify for realtime users. So if we keep it this way, then every time we add another message we increase the complexity and maybe even create more circular dependencies which make it harder to understand or test the app.

How about using the nestjs event system instead? https://docs.nestjs.com/techniques/events

InnayTool:
How about merging this PR and implement the event system in another one?

mrdrogdrog (changes requested):
The new public methods aren't tested

mrdrogdrog (approved):

PR #2689, opened 2022-09-25
feat: check password strength for local login 

### LINKED ISSUE #1846: Configurable password policies for internal auth
### Is your feature request related to a problem? Please describe.

Short or weak passwords can be a security risk to the instance's data.

Two ideas:
 - Instance owners can define a minimum password length for internal login.
 - Instance owners can define a regex that needs to be fulfilled for passwords of internal login.

Find sane default values and don't let the instance owner define them. However, currently even empty passwords containing only of spaces would be accepted and that doesn't seem right.

### Additional context

_No response_

DerMolly:
### Component/Part
local login

### Description
This PR adds a minimal strength level to local passwords.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
closes #1846


InnayTool (approved):

davidmehren:
The linked issue talked about enforcing a minimum password length, this is quite a different solution and introduces a dependency to a password-strength-estimation lib, which spits out a mysterious score.

I'm not really convinced of this approach. In my opinion, we should keep all security-related code as lean as possible to reduce attack surface. If anyone wants a fancy check algorithm for their users passwords, they can always deploy Keycloak or any other auth solution.

docs/content/config/index.md
```diff
@@ -45,7 +45,17 @@ If the renderer is provided by another domain, it's way harder to manipulate Hed
 
 ## Authentication
 
-**ToDo:** Add Authentication docs
+### Local
+
+HedgeDoc provides local accounts, handled and maintained by HedgeDoc. These local accounts can be used, but it's strongly recommended to use one of the other authentication methods.
```
davidmehren, line 50:
```suggestion
HedgeDoc provides local accounts, handled internally. This feature only provides basic functionality, so for most environments we recommend using an external authentication mechanism. This also enables more secure authentication like 2FA or WebAuthn.
```

DerMolly: fixed

InnayTool (approved):

davidmehren (approved):

PR #2719, opened 2022-10-02
ci: add paths command to skip unnecessary GitHub actions

DerMolly:
### Component/Part
CI

### Description
This PR skips unnecessary GitHub actions via paths command.

### Steps

- [x] Added implementation
  made sure that:


DerMolly:
It appears we can't skip required test and be able to merge the PR without admin privileges. See also https://docs.github.com/en/actions/managing-workflow-runs/skipping-workflow-runs. So either we remove the required flag from all those workflows (I think that's a bad idea) or we only use it in the not required CodeQL workflow…

PR #2970, opened 2022-11-27
Yarn 3.3.0 + Yarn Workspaces

mrdrogdrog:
### Component/Part
Yarn configuration

### Description
This PR updates yarn to 3.3.0 and enables workspaces

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:


ErikMichelson (approved):

InnayTool (approved):

mrdrogdrog:
TBH i'm not sure how to continue.

> What happened?

So far I can install and run all components. Everything lints and starts. BUT. I can't create a production build of the frontend.

> Whats the problem with the frontend?

I can't create a production build of the frontend because the next.js standalone packaging process, which creates a minified build and tree-shakes the dependencies, seem to rely on a single relative node_modules directory instead of gathering dependencies using the node package linker (e.g. using `require`). Unusual locations of `node_modules` confuse this logic. That's also the same reason why we can't use yarn pnp which doesn't use a node_modules directory at all.

> So yarn is the problem because it's doing something weird again??

Well. Actually yarn isn't doing anything wrong here. Having multiple node_modules directories (or placing them somewhere else than in the project directory) isn't prohibited. The same problem would also occur if you e.g. use packages from a system-wide node_modules directory.

> What about the problems we had last time?

I solved every problem regarding eslint, prettier, etc. The current status is, that frontend and backend can run in dev mode. The mystery is the last one (so far).

> Why do we need workspaces at all then??

The next step after this would be to import the [commons lib](https://github.com/hedgedoc/commons) into the mono-repository and use it without compiling and uploading it to npm. Yarn workspaces include other sibling packages directly through the linker.

> So.. what now? 

No idea so far. please stand by.

mrdrogdrog:
The problem with the frontend build is solved. The solution was to set `outputFileTracingRoot` in the next config so it detects the root of our monorepo correctly.

davidmehren (approved):

PR #2973, opened 2022-11-29
Fix translation login button

ImaCrea:
Signed-off-by: ImaCrea <maximeguedj@pm.me>

### Component/Part
UI button

### Description
The translation for the login button in French was not easy as it was translating only "Register". So people wanting to sign in were lost and looking for a button to login. Best solution found is to insert a translation for "Register | Log in" in French : "S'inscrire | Se connecter".

Here's a screenshot done for testing if it looks well:
Capture d’écran 2022-11-29 à 19 43 58

Tested with "average" user and it does the job. :)

### Steps
Simply updates a translation.

### DIFF
--- a/locales/fr.json
+++ b/locales/fr.json
@@ -8,7 +8,7 @@
     "Collaborate with URL": "Collaboration en temps-réel",
     "Support charts and MathJax": "Gère les graphiques et MathJax",
     "Support slide mode": "Gère le mode présentation",
-    "Sign In": "S'inscrire",
+    "Sign In": "S'inscrire | Se connecter",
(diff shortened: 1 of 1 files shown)


mrdrogdrog:
Hello, thanks for your help. 
Every time we create a hedgedoc release we import the translations from POEditor and overwrite the language files in the repository. 
So if you want to change translations then please do this on https://translate.hedgedoc.org/ . Otherweise your change would be lost.

ImaCrea:
Thanks @mrdrogdrog didn't know translation was done there. I signed up on POEditor and changed it.
Cheers,

PR #3089, opened 2022-12-28
fix(i18n): clarify where to find /features for new users

lukas-mertens:
Signed-off-by: Lukas Mertens <git@lukas-mertens.de>

### Component/Part
Editor

### Description
This PR changes the default placeholder to be more user-friendly. As a new user 

### Steps

- [x] Added implementation
- [ ] Added / updated tests
- [x] Added changelog snippet
  made sure that:

image

### DIFF
--- a/frontend/src/components/editor-page/editor-pane/editor-pane.tsx
+++ b/frontend/src/components/editor-page/editor-pane/editor-pane.tsx
@@ -4,6 +4,7 @@
  * SPDX-License-Identifier: AGPL-3.0-only
  */
 import { useApplicationState } from '../../../hooks/common/use-application-state'
+import { useBaseUrl, ORIGIN } from '../../../hooks/common/use-base-url'
 import { useDarkModeState } from '../../../hooks/common/use-dark-mode-state'
 import { cypressAttribute, cypressId } from '../../../utils/cypress-attribute'
 import { findLanguageByCodeBlockName } from '../../markdown-renderer/extensions/base/code-block-markdown-extension/find-language-by-code-block-name'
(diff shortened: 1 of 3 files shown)


frontend/src/components/editor-page/editor-pane/editor-pane.tsx
```diff
@@ -140,7 +140,7 @@ export const EditorPane: React.FC<ScrollProps> = ({ scrollState, onScroll, onMak
       <ToolBar />
       <ReactCodeMirror
         editable={firstUpdateHappened && connectionSynced}
-        placeholder={t('editor.placeholder') ?? ''}
+        placeholder={t('editor.placeholder', { host: window.location.host }) ?? ''}
```
lukas-mertens, line 143: If there is a better way to do this in react, please let me know.

mrdrogdrog: This breaks in server side rendering because the window object doesn't exist there. Also this doesn't include port, protocol or subdirectory.
 
You should use the `useBaseUrl` hook to get the editor base url.

```typescript
const editorOrigin = useBaseUrl(ORIGIN.EDITOR)
```

mrdrogdrog (changes requested):

frontend/src/components/editor-page/editor-pane/editor-pane.tsx
lukas-mertens: Thanks. Please review again

mrdrogdrog (changes requested):
Don't use the hook directly in the object. Please save the value in a const first and use this.

lukas-mertens:
> Don't use the hook directly in the object. Please save the value in a const first and use this.

I changed that, although I don't quite understand why yet. Is it because of performance-reasons because this way the hook is called less often on updates?

I am currently having problems figuring out why the linter is failing in the pipeline. Locally executing `prettier -c src/components/editor-page/editor-pane/editor-pane.tsx --config .prettierrc.json` works.

mrdrogdrog:
It's just because of code style.

lukas-mertens:
I see. I rebased as it was outdated 👍

mrdrogdrog (approved):

mrdrogdrog:
Thanks @lukas-mertens :)

Issue #2574, opened 2022-08-17
Metrics and status exposed by default


davidc:
### Which part of the project should be enhanced?

metrics/status API

### Is your enhancement request related to a problem? Please describe.

Confidential information is exposed at /metrics and /status by default and this is neither documented nor does there seem to be an easy way to disable it.

### Describe the solution you'd like

1) A configuration option to disable public exposure of metrics (Yes I can block /status and /metrics in the reverse proxy, but those are the only ones I know about, maybe there are more, maybe there will be more in future)

and either:

2) This option set by default so that this information is not exposed
or
3) Very clear and prominent information in all three installation documents warning people that it is enabled by default

### Describe alternatives you've considered

Obviously now I am aware of it I can block these specific URLs in the reverse proxy, however what about the many other people who may not be aware they have been leaking information for months?

SISheogorath:
May I ask which of the information exposed there is considered sensitive and in which context?

ErikMichelson:
> Additionally /metrics is exposing URLs which could include PII.

No, the /metrics route does only include the routes as set up in the source code. For example `/:noteId`  instead of actual note ids. So there is no PII exposed in that way as all the routes are publicly visible in the source code.

almereyda:
I would second the general idea of the original posting, that the `/metrics` and `/status` endpoints provide observability information specific to an instance and primarily used for orchestration and observation purposes.

A possible way to mitigate their exposure by default, would be to have them listen on a separate interface/port, optionally on `localhost`, in order to separate the concerns of the routes cleanly. This would still allow to retrieve them where the process is running, but avoids to have them readily available in the open internet.

davidc:
> > Additionally /metrics is exposing URLs which could include PII.
> 
> No, the /metrics route does only include the routes as set up in the source code. For example `/:noteId` instead of actual note ids. So there is no PII exposed in that way as all the routes are publicly visible in the source code.

This may be the intention but it is not the reality, if you visit /metrics you see thousands of URLs.

mrdrogdrog:
> thousands of URLs

But none of these contain PII

ErikMichelson:
This was implemented with #4165.

PR #3219, opened 2023-01-23
config: Add a flag to control the /metrics and /status endpoints

freesteph:
### Component/Part
Application

### Description
It can be a security concern in some environments to expose system capabilities even though they don't expose any PII. Add a couple flags to control whether the `/metrics` and `/status` (and anything in the StatusRouter) are exposed.

### Steps

- [X] Added implementation
- [ ] Added / updated tests
- [X] Added changelog snippet
- [X] I read the [contribution documentation](https://github.com/hedgedoc/hedgedoc/blob/develop/CONTRIBUTING.md) and
  made sure that:

### Related Issue(s)
https://github.com/hedgedoc/hedgedoc/issues/2574


docs/content/configuration.md
```diff
@@ -76,24 +76,27 @@ these are rarely used for various reasons.
...
+| `cookiePolicy`                | `CMD_COOKIE_POLICY`                | **`lax`**, `strict` or `none`                                                             | Set a SameSite policy whether cookies are send from cross-origin. Be careful: setting a SameSite value of none without https breaks the editor.                           |
+| `observability.exposeMetrics  | `CMD_OBSERVABILITY_EXPOSE_METRICS` | **true** or `false`                                                                       | Controls whether the /metrics endpoint is exposed.                                                                                                                        |
+| `observability.exposeMetrics  | `CMD_OBSERVABILITY_EXPOSE_STATUS`  | **true** or `false`                                                                       | Controls whether the /status routes are exposed.                                                                                                                          |
```
DerMolly, line 98:
```suggestion
| `observability.exposeStatus  | `CMD_OBSERVABILITY_EXPOSE_STATUS`  | **true** or `false`                                                                       | Controls whether the /status routes are exposed.                                                                                                                          |
```

DerMolly (changes requested):

mrdrogdrog (changes requested):
Will approve as soon as docs are fixed

freesteph:
All done I believe. Thanks for looking into this!

ErikMichelson (changes requested):
Two points:
- The config option is quite long with the term “observability” inside. In my opinion, `CMD_EXPOSE_METRICS` would be good enough.
- Having the status endpoint disabled, will break the [docker image's health check](https://github.com/hedgedoc/container/blob/master/resources/healthcheck.mjs). Therefore, I suggest that queries from localhost will always be allowed, no matter the config value.

DerMolly (approved):

mrdrogdrog:
Regarding the docker healthcheck we could use a new route that just returns "ok".

freesteph:
> Two points:
> 
>     * The config option is quite long with the term “observability” inside. In my opinion, `CMD_EXPOSE_METRICS` would be good enough. This is not very important to me, but I wanted to note it anyway.
> 
>     * Having the status endpoint disabled, will break the [docker image's health check](https://github.com/hedgedoc/container/blob/master/resources/healthcheck.mjs). Therefore, I suggest that queries from localhost will always be allowed, no matter the config value.

Thanks for the feedback. I think it makes sense to group these two options into an `observability` scope, it allows grouping them which is nice. Up to you though.

I've added a commit to keep `/status` returning 200, let me know if that's good enough.

ErikMichelson (approved):

mrdrogdrog (approved):

InnayTool (approved):

PR #3268, opened 2023-02-05
fix(config): Replace HD_DOMAIN and HD_EDITOR_BASE_URL with HD_BASE_URL

mrdrogdrog:
### Component/Part
Configuration of backend and frontend

### Description
This PR replaces the env variables `HD_DOMAIN` and `HD_EDITOR_BASE_URL` with `HD_BASE_URL`.
This is done because HD_DOMAIN doesn't contain only a domain and to unify the configuration.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:


backend/src/config/app.config.spec.ts
```diff
@@ -9,9 +9,9 @@ import appConfig from './app.config';
 import { Loglevel } from './loglevel.enum';
 
 describe('appConfig', () => {
-  const domain = 'https://example.com';
-  const invalidDomain = 'localhost';
-  const rendererBaseUrl = 'https://render.example.com';
+  const baseUrl = 'https://example.com/';
```
ErikMichelson, line 12: Instead of requiring the trailing slash everywhere, what about adding it automatically where it's needed if it is absent?

mrdrogdrog: We had that discussion before and my argument is still the same: I don't want the software to assume settings.

mrdrogdrog: The base URL can also contain a subdirectory.
If you don't require the slash then the software doesn't know if `https://example.org/asd` means "the admin forgot to remove the file and means `https://example.org/` or "the admin forgot to add the slash and means `https://example.org/asd/`.

mrdrogdrog: Okay. During chat we came to the compromise that we won't enforce the trailing slash if the user is giving only a protocol and domain. But trailing slashes are mandatory if they specify a subpath.

docs/content/config/index.md
```diff
@@ -21,9 +21,9 @@ We also provide an `.env.example` file containing a minimal configuration in the
...
-| `HD_DOMAIN`              | -         | `https://md.example.com`    | The URL the HedgeDoc instance runs on.                                                                                                                 |
+| `HD_BASE_URL`              | -         | `https://md.example.com`    | The URL the HedgeDoc instance runs on.                                                                                                                 |
 | `PORT`                   | 3000      |                             | The port the HedgeDoc instance runs on.                                                                                                                |
-| `HD_RENDERER_BASE_URL`   | HD_DOMAIN |                             | The URL the renderer runs on. If omitted this will be same as `HD_DOMAIN`.                                                                             |
+| `HD_RENDERER_BASE_URL`   | HD_BASE_URL |                             | The URL the renderer runs on. If omitted this will be same as `HD_BASE_URL`.                                                                             |
```
DerMolly, line 26:
```suggestion
| `HD_RENDERER_BASE_URL`   | `HD_BASE_URL` |                             | The URL the renderer runs on. If omitted this will be same as `HD_BASE_URL`.                                                                             |
```

ErikMichelson (approved):

DerMolly (approved):

ErikMichelson (approved):

PR #3327, opened 2023-02-12
Refactor DefaultAccessLevel & NotesService

davidmehren:
### Component/Part
DefaultAccessLevel & NotesService

### Description
This PR contains two refactorings extracted from my WIP E2E-tests branch:
- The `DefaultAccessPermission` enum is renamed to `DefaultAccessLevel`, to reduce confusion between the Permission class and the enum
- `NoteService.createNote` is rejiggled to make it easier to understand, comments are added

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/backend/src/notes/notes.service.ts
+++ b/backend/src/notes/notes.service.ts
@@ -9,7 +9,7 @@ import { EventEmitter2 } from '@nestjs/event-emitter';
 import { InjectRepository } from '@nestjs/typeorm';
 import { Repository } from 'typeorm';
 
-import { DefaultAccessPermission } from '../config/default-access-permission.enum';
+import { DefaultAccessLevel } from '../config/default-access-level.enum';
 import noteConfiguration, { NoteConfig } from '../config/note.config';
 import {
   AlreadyInDBError,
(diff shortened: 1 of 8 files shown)


backend/src/notes/notes.service.ts
```diff
@@ -92,37 +92,52 @@ export class NotesService {
...
       this.checkNoteIdOrAlias(alias);
     }
-    const newNote = Note.create(owner, alias);
     if (noteContent.length > this.noteConfig.maxDocumentLength) {
       throw new MaximumDocumentLengthExceededError();
     }
-    //TODO: Calculate patch
+
+    // Create new note and set the initial properties
+    // We cast to a note early to keep the later code clean
```
mrdrogdrog: redundant comment

backend/src/notes/notes.service.ts
```diff
@@ -92,37 +92,52 @@ export class NotesService {
...
+
+    let everyoneAccessLevel;
+
     if (owner) {
+      // When we know an owner, an initial history entry is created
       newNote.historyEntries = Promise.resolve([
-        HistoryEntry.create(owner, newNote as Note) as HistoryEntry,
+        HistoryEntry.create(owner, newNote) as HistoryEntry,
       ]);
+      // Use the default access level from the config
```
mrdrogdrog, line 117: Why are you describing the next line?

backend/src/notes/notes.service.ts
```diff
@@ -92,37 +92,52 @@ export class NotesService {
...
     newNote.revisions = Promise.resolve([
-      Revision.create(noteContent, noteContent, newNote as Note) as Revision,
+      //TODO: Calculate patch
+      Revision.create(noteContent, noteContent, newNote) as Revision,
     ]);
+
+    let everyoneAccessLevel;
+
     if (owner) {
+      // When we know an owner, an initial history entry is created
```
mrdrogdrog, line 113: Why are you describing this?

backend/src/notes/notes.service.ts
```diff
@@ -92,37 +92,51 @@ export class NotesService {
...
-      newNote as Note,
+      newNote,
       await this.groupsService.getLoggedInGroup(),
       this.noteConfig.permissions.default.loggedIn,
     );
 
+    // Merge into permissions array
     newNote.groupPermissions = Promise.resolve([
       ...Optional.ofNullable(everyonePermission).wrapInArray(),
       ...Optional.ofNullable(loggedInPermission).wrapInArray(),
```
mrdrogdrog, line 142: instead of creating blocks using new lines and comments you could extract these part-functionalities into functions.

InnayTool (approved):

mrdrogdrog (changes requested):

backend/src/notes/notes.service.ts
```diff
@@ -148,15 +162,17 @@ export class NotesService {
...
-        );
+    if (accessLevel === DefaultAccessLevel.NONE) {
+      return null;
+    } else {
+      return NoteGroupPermission.create(
+        group,
+        note,
+        accessLevel === DefaultAccessLevel.WRITE,
+      );
+    }
```
ErikMichelson, line 175: The else branch does not require parentheses as the if-condition already does a return.
Removing the parentheses makes the code more readable.

ErikMichelson (changes requested):

backend/src/notes/notes.service.ts
mrdrogdrog: In this case I would prefer a ternary operator or keep the indentation so both returns are on one level.

backend/src/notes/notes.service.ts
ErikMichelson: It's true, however as it is not a simple one-line statement but instead three lines, I wouldn't say this needs to be removed explicitly.

DerMolly (approved):

ErikMichelson (approved):

mrdrogdrog (approved):

PR #3607, opened 2023-03-24
refactor: save ydoc state in the database, so it can be restored easier

mrdrogdrog:
### Component/Part
Realtime communication

### Description
By storing the ydoc state in the database we can reconnect lost clients easier and enable offline editing because we continue using the crdt data that has been used by the client before the connection loss.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:


frontend/src/components/editor-page/editor-pane/hooks/yjs/use-code-mirror-yjs-extension.ts
```diff
@@ -8,30 +8,33 @@ import { YTextSyncViewPlugin } from '../../codemirror-extensions/document-sync/y
...
 import type { YDocSyncClientAdapter } from '@hedgedoc/commons'
+import type { RealtimeDoc } from '@hedgedoc/commons'
 import { useEffect, useMemo, useState } from 'react'
-import type { Text as YText } from 'yjs'
 
 /**
  * Creates a {@link Extension code mirror extension} that synchronizes an editor with the given {@link YText ytext}.
  *
- * @param yText The source and target for the editor content
+ * @param doc The {@link RealtimeDoc realtime doc} that contains the markdown contentn text channel
```
ErikMichelson, line 17:
```suggestion
 * @param doc The {@link RealtimeDoc realtime doc} that contains the markdown content text channel
```

backend/src/realtime/realtime-note/realtime-note.ts
```diff
@@ -28,12 +28,13 @@ export class RealtimeNote extends EventEmitter2<RealtimeNoteEventMap> {
   private readonly clients = new Set<RealtimeConnection>();
   private isClosing = false;
 
-  constructor(private readonly note: Note, initialContent: string) {
+  constructor(private readonly note: Note, initialState: string | number[]) {
```
DerMolly, line 31: Maybe we could use multiple contructors for those two different cases?

commons/src/y-doc-sync/realtime-doc.ts
```diff
@@ -3,27 +3,53 @@
...
+
   /**
    * Creates a new instance.
    *
    * The new instance is filled with the given initial content.
    *
    * @param initialContent - the initial content of the {@link Doc YDoc}
    */
-  constructor(initialContent?: string) {
+  constructor(initialContent?: string | number[]) {
```
DerMolly, line 39: same here

commons/src/y-doc-sync/realtime-doc.ts
```diff
@@ -3,27 +3,53 @@
...
     super()
     if (initialContent) {
-      this.getMarkdownContentChannel().insert(0, initialContent)
+      if (typeof initialContent === 'string') {
```
davidmehren, line 42: I'm a bit unhappy with this union type and the following `typeof`.

Actually, with the whole concept of mixing "a string of initial content" with "an arbitrary binary blob of yjs state" that gets introduced with this PR. I think these two types of data should be more cleanly separated.

DerMolly (changes requested):

backend/src/realtime/realtime-note/realtime-note.ts
mrdrogdrog: Not possible in typescript. The way to go here is a union type.

ErikMichelson: You could add two separate static methods to the class as factory methods

davidmehren (approved):

DerMolly (approved):

ErikMichelson (approved):

PR #3613, opened 2023-03-24
fix(landing-layout): links in the user-dropdown not working

ErikMichelson:
### Component/Part
Landing layout

### Description
This PR fixes the links in the user dropdown.

We use a custom component for rendering the link (Dropdown.Item), therefore we need to set both passHref and legacyBehavior

See: https://nextjs.org/docs/api-reference/next/link#if-the-child-is-a-custom-component-that-wraps-an-a-tag

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
none

### DIFF
--- a/frontend/src/components/landing-layout/navigation/user-dropdown.tsx
+++ b/frontend/src/components/landing-layout/navigation/user-dropdown.tsx
@@ -33,13 +33,13 @@ export const UserDropdown: React.FC = () => {
       </Dropdown.Toggle>
 
       <Dropdown.Menu className='text-start'>
-        <Link href={'/n/features'} passHref={true}>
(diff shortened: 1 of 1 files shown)


mrdrogdrog:
Have you tested if it works without the legacy behaviour if you put the link tag into the item?

ErikMichelson:
Not yet, however the problem was an `<a>` tag inside an  `<a>` tag. I don't think that the other way around this will be functional. But I'll test it.

ErikMichelson:
> Have you tested if it works without the legacy behaviour if you put the link tag into the item?

As expected, that's broken.

image

mrdrogdrog (approved):

PR #3617, opened 2023-03-24
fix(cursor): cursor text being hidden when at start of line

ErikMichelson:
### Component/Part
frontend -> editor -> remote cursors

### Description
This PR fixes the cursor text not being visible when the cursor is at the start of the line.
Additionally, the cursor's font was not defined and therefore was styled in the browser's default font

image

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
none

### DIFF
--- a/frontend/src/components/editor-page/editor-pane/codemirror-extensions/remote-cursors/remote-cursor-marker.ts
+++ b/frontend/src/components/editor-page/editor-pane/codemirror-extensions/remote-cursors/remote-cursor-marker.ts
@@ -18,7 +18,8 @@ export class RemoteCursorMarker implements LayerMarker {
     private top: number,
     private height: number,
     private name: string,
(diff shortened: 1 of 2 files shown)


mrdrogdrog:
Now the cursor is hidden when you're at the end of the line. Doesn't really solve the problem.

mrdrogdrog:
To do it correctly we would need to decide if the tag should be left or right aligned.
we could either calculate it based on the tag width or just by deciding that it's right aligned if the position is over 50% of the codemirror.

DerMolly (approved):

mrdrogdrog (approved):

mrdrogdrog:
image

This happens in the first line. We should fix this.

DerMolly (approved):

PR #3643, opened 2023-03-26
feat: check permissions in realtime code and frontend

DerMolly:
### Component/Part
frontend / commons

### Description
This PR adds permission checks in the realtime code and the frontend

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:


commons/src/utils/permissions.ts
```diff
@@ -0,0 +1,32 @@
+/*
+ * SPDX-FileCopyrightText: 2023 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+import { NotePermissions, SpecialGroup } from './permissions.types.js'
+
```
ErikMichelson, line 7:
```suggestion

/**
 * Checks if the given user is the owner of a note.
 *
 * @param permissions The permissions of the note to check
 * @param user The username of the user
 * @return True if the user is the owner of the note
 */
```

commons/src/utils/permissions.ts
```diff
@@ -0,0 +1,32 @@
...
+export const userIsOwner = (
+  permissions: NotePermissions,
+  user?: string
+): boolean => {
+  return !!user && permissions.owner === user
+}
+export const userCanEdit = (permissions: NotePermissions, user?: string) => {
```
ErikMichelson, line 14:
```suggestion
/**
 * Checks if the given user may edit a note.
 *
 * @param permissions The permissions of the note to check
 * @param user The username of the user
 * @return True if the user has the permission to edit the note
 */
export const userCanEdit = (permissions: NotePermissions, user?: string): boolean => {
```

commons/src/utils/permissions.ts
```diff
@@ -0,0 +1,32 @@
...
+  const isOwner = userIsOwner(permissions, user)
+  const mayWriteViaUserPermission = permissions.sharedToUsers.some(
+    (value) => value.canEdit && value.username === user
+  )
+  const mayWriteViaGroupPermission = permissions.sharedToGroups.some(
+    (value) =>
+      value.groupName === SpecialGroup.LOGGED_IN && value.canEdit && !!user
+  )
```
ErikMichelson, line 22: This checks for `!!user` in every iteration of the loop that runs over the sharedToGroups array. You could change this to the following:

```suggestion
  const mayWriteViaGroupPermission = !!user && permissions.sharedToGroups.some(
    (value) =>
      value.groupName === SpecialGroup.LOGGED_IN && value.canEdit
  )
```

ErikMichelson (changes requested):

commons/src/utils/permissions.ts
ErikMichelson: The boolean return type is still missing

ErikMichelson (changes requested):

mrdrogdrog:
Well depends. Should they see the other people cursors or not? Because that's coupled.

DerMolly:
I think they should. Other people should not see the cursor of the read-only user, but the read-only users should see what happens in the pad.

mrdrogdrog (approved):

ErikMichelson (approved):

PR #3689, opened 2023-03-29
fix(profile): show correct error messages

ErikMichelson:
### Component/Part
profile page -> change password

### Description
This PR fixes the error messages that will be displayed on the password change form.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
none

### DIFF
--- a/frontend/src/api/common/error-to-i18n-key-mapper.spec.ts
+++ b/frontend/src/api/common/error-to-i18n-key-mapper.spec.ts
@@ -1,0 +1,57 @@
+/*
+ * SPDX-FileCopyrightText: 2023 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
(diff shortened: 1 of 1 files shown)


ErikMichelson:
There are various things (which are all coupled together thus only one commit):

- The success message was always shown for the password change thing
- The ErrorToI18nKeyMapper had an error when using the fallback key

Yes, one could move the error message to a more common namespace. However, I guess there could be other possible places in the future where it makes sense to not only restrict the key to a certain namespace.

DerMolly (approved):

mrdrogdrog:
I mean yes, I see why the fixes are necessary. But i'm still not sure if the `treatAsAbsoluteKey` is the right approach here. I would rather gather the translations for the errors under one namespace per API. But we can discuss this later.

mrdrogdrog (commented):
Oh just because you're already on it. Would you be so kind to add tests for the mapper?

Issue #3706, opened 2023-03-31
Refactor Permissions test file.


Avi98:
### Which part of the project should be enhanced?

permissions.service.spec.ts

### Is your enhancement request related to a problem? Please describe.

in the permission service test file has some redundant test mocks that needs to refactor need

### Describe the solution you'd like

Refactor test

### Describe alternatives you've considered

Don't refactor the test. And live with redundant tests

PR #3711, opened 2023-03-31
refactor: create mock function for noteRepo and eventemiter

Avi98:
### Description
Fixes [3706](https://github.com/hedgedoc/hedgedoc/issues/3706)

### Steps

- [x] Added implementation
- [ ] Added / updated tests
- [ ] Added changelog snippet
  made sure that:

### Related Issue(s)
[3706](https://github.com/hedgedoc/hedgedoc/issues/3706)

### DIFF
--- a/backend/src/permissions/permissions.service.spec.ts
+++ b/backend/src/permissions/permissions.service.spec.ts
@@ -46,6 +46,23 @@ import { NoteUserPermission } from './note-user-permission.entity';
 import { PermissionsModule } from './permissions.module';
 import { PermissionsService } from './permissions.service';
 
+function mockedEventEmitter(eventEmitter: EventEmitter2) {
+  return jest.spyOn(eventEmitter, 'emit').mockImplementationOnce((event) => {
+    expect(event).toEqual(NoteEvent.PERMISSION_CHANGE);
+    return true;
+  });
+}
+
+function mockNoteRepo(noteRepo: Repository<Note>) {
+  jest
+    .spyOn(noteRepo, 'save')
+    // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+    // @ts-ignore
+    .mockImplementationOnce(async (entry: Note) => {
+      return entry;
+    });
+}
+
 describe('PermissionsService', () => {
   let service: PermissionsService;
   let notes: Note[];
@@ -53,6 +70,7 @@ describe('PermissionsService', () => {
   let userRepo: Repository<User>;
   let groupRepo: Repository<Group>;
   let eventEmitter: EventEmitter2;
+  let eventEmitterEmitSpy: jest.SpyInstance;
   const noteMockConfig: NoteConfig = createDefaultMockNoteConfig();
 
   beforeAll(async () => {
@@ -61,6 +79,7 @@ describe('PermissionsService', () => {
      * array and the overrideProvider call, as otherwise we have two instances
      * and the mock of createQueryBuilder replaces the wrong one
      * **/
+
     userRepo = new Repository<User>(
       '',
       new EntityManager(
@@ -81,6 +100,7 @@ describe('PermissionsService', () => {
       ),
       undefined,
     );
+
     const module: TestingModule = await Test.createTestingModule({
       providers: [
         PermissionsService,
@@ -149,6 +169,10 @@ describe('PermissionsService', () => {
     eventEmitter = module.get<EventEmitter2>(EventEmitter2);
   });
 
+  beforeEach(() => {
+    mockNoteRepo(noteRepo);
+    eventEmitterEmitSpy = mockedEventEmitter(eventEmitter);
+  });
   afterEach(() => {
     jest.clearAllMocks();
   });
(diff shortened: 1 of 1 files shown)


mrdrogdrog:
Is this ready for review?

Avi98:
> Is this ready for review?

Yes @mrdrogdrog, have I missed anything ?

backend/src/permissions/permissions.service.spec.ts
```diff
@@ -171,6 +181,27 @@ describe('PermissionsService', () => {
...
+    return spyOnNote(methodName).mockImplementationOnce(implementation);
+  }
+
+  function mockEventEmitterMethods(
+    methodName: MethodType<typeof eventEmitter>,
+    implementation: ImplementationType,
+  ) {
+    return spyOnEventEmitter(methodName).mockImplementationOnce(implementation);
+  }
+
```
mrdrogdrog, line 204: This extraction is a bit overengineered. You use each of this functions for exactly one purpose which would be okay if the extracted code is very complex or long. But in this case it's not really helpful :/

mrdrogdrog (changes requested):

backend/src/permissions/permissions.service.spec.ts
Avi98: @mrdrogdrog I have removed unnecessary abstraction.

backend/src/permissions/permissions.service.spec.ts
```diff
@@ -171,6 +181,24 @@ describe('PermissionsService', () => {
...
+  function mockEventEmitterMethods(
+    methodName: MethodType<typeof eventEmitter>,
+    implementation: ImplementationType,
+  ) {
+    return jest
+      .spyOn(eventEmitter, methodName)
+      .mockImplementationOnce(implementation);
+  }
+
```
mrdrogdrog, line 201: Almost, but not exactly. You use mockNoteMethods only for 'save' and mockEventEmitterMethods only for 'emit'.
I don't see a reason for a method-level abstraction if it's not used for other methods.

mrdrogdrog (changes requested):

backend/src/permissions/permissions.service.spec.ts
Avi98: @mrdrogdrog  removed different method level abstraction. Now I'm using a single abstraction for both methods.

mrdrogdrog: Sry but i still don't get why you create this abstraction at all. Everywhere where you use these two methods they're used in the exact same way. You always use the mockNoteMethods for mocking the save function and mockEventEmitterMethods to mock the emit function so it expects a certain call.
That's an indication that the parameters are not needed. 
It would be much easier if you create just two functions. One that mocks the save function of the repository and one that mocks the event emitter to expect a certain event. There is no need for more complex abstraction.

You could even move the mock of the save function into the beforeAll because it is done in all tests that need it.

Avi98: @mrdrogdrog, I have removed the abstraction. I only have one mock function for event emitter and one for note repository. 
Please let me know if i have missed anything

backend/src/permissions/permissions.service.spec.ts
```diff
@@ -727,35 +751,19 @@ describe('PermissionsService', () => {
...
         sharedToGroups: [],
       });
-      expect(mockedEventEmitter).toHaveBeenCalled();
+      expect(emitEvent).toHaveBeenCalled();
     });
     describe('works', () => {
+      beforeEach(() => {
+        mockNoteRepo(noteRepo);
+        emitEvent;
+      });
```
mrdrogdrog, line 765: This is already handled by the previous beforeEach you've inserted.

backend/src/permissions/permissions.service.spec.ts
```diff
@@ -46,13 +46,31 @@ import { NoteUserPermission } from './note-user-permission.entity';
...
+}
+
 describe('PermissionsService', () => {
   let service: PermissionsService;
   let notes: Note[];
   let noteRepo: Repository<Note>;
   let userRepo: Repository<User>;
   let groupRepo: Repository<Group>;
   let eventEmitter: EventEmitter2;
+  let emitEvent: jest.SpyInstance;
```
mrdrogdrog, line 73: The name isn't quite fitting. The variable doesn't contain an Event.
Something like `eventEmitterEmitSpy` would be more precise.

mrdrogdrog (changes requested):

mrdrogdrog:
Almost there :)

backend/src/permissions/permissions.service.spec.ts
Avi98: removed

backend/src/permissions/permissions.service.spec.ts
Avi98: done

mrdrogdrog (approved):

ErikMichelson (approved):

mrdrogdrog:
Thank you for your contribution ❤️

PR #3757, opened 2023-04-08
feat: move first heading title extraction into an app extension

mrdrogdrog:
### Component/Part
Frontend

### Description
This PR moves the first title extraction into a markdown renderer extension.

### Steps

- [x] Added implementation
  made sure that:


commons/src/note-frontmatter-parser/parse-tags.ts
```diff
@@ -0,0 +1,17 @@
...
+/**
+ * Parses the given value as tags array.
+ *
+ * @param rawTags The raw value to parse
+ * @return the parsed tags
+ */
+export const parseTags = (rawTags: string | string[]): string[] => {
+  return (Array.isArray(rawTags) ? rawTags : rawTags.split(','))
+    .map((entry) => entry.trim())
+    .filter((tag) => !!tag)
```
Avi98, line 16: we can also use Boolean constructor 

```suggestion
    .filter(Boolean)
```

mrdrogdrog: Personal preference 🤷

InnayTool (changes requested):

backend/src/revisions/revision.entity.ts
```diff
@@ -79,13 +108,60 @@ export class Revision {
...
+    newRevision.title = title;
+    newRevision.description = description;
+    newRevision.tags = Promise.resolve(tags);
     newRevision.note = Promise.resolve(note);
     newRevision.edits = Promise.resolve([]);
     newRevision.yjsStateVector = yjsStateVector ?? null;
     return newRevision;
   }
+
+  private static createTagEntities(frontmatter?: NoteFrontmatter): Tag[] {
```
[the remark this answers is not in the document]
mrdrogdrog: I can't call services from entities because of missing service injection.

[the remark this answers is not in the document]
mrdrogdrog: But then I would need to move the whole calculation logic into the service because the tags are generated from the content.

backend/src/notes/notes.service.ts
```diff
@@ -102,9 +103,10 @@ export class NotesService {
 
     // We cast to a note early to keep the later code clean
     const newNote = Note.create(owner, alias) as Note;
+
+    const patch = createPatch(newNote.publicId, '', noteContent);
     newNote.revisions = Promise.resolve([
-      //TODO: Calculate patch
-      Revision.create(noteContent, noteContent, newNote) as Revision,
+      Revision.create(noteContent, patch, newNote) as Revision,
```
InnayTool, line 109: shouldn't we call the `revisionsService` here?

backend/src/revisions/revision.entity.ts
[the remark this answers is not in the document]
InnayTool: Yes, that's how it is done in other services, I think.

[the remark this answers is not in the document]
mrdrogdrog: I moved the logic into another file

backend/src/revisions/revisions.service.ts
```diff
@@ -158,7 +164,13 @@ export class RevisionsService {
...
       newContent,
     );
-    const revision = Revision.create(newContent, patch, note, yjsStateVector);
+    const revision = Revision.create(
+      newContent,
+      patch,
+      note,
+      yjsStateVector,
+    ) as Revision;
+    updateRevisionMetadataFromContent(revision);
```
InnayTool, line 173: Is it valid to create a revision without metadata?

backend/src/notes/notes.service.ts
mrdrogdrog: I've been thinking about that too. But there are some other problems.
- The revision service method saves the creates revision entity, which is not possible at this point because the note doesnt exist.
- getLatestRevision throws an error if the provided note doesn't have an id or any revisions. (The whole "lets cast the ENTITY_WITHOUT_ID to ENTITY" behaviour is firing back now :sweat_smile: )

I would prefer to keep it this way and do a major refactoring of the whole `.create() as ENTITY` code.

backend/src/revisions/revisions.service.ts
mrdrogdrog: No. that would create inconsistency because the metadata would be out of sync with the content.

InnayTool: So shouldn't we put a parameter for metadata in the create method of the revision?

mrdrogdrog: I restructured and separated the code to do this

DerMolly (approved):

backend/src/revisions/revision-metadata.dto.ts
```diff
@@ -50,4 +50,31 @@ export class RevisionMetadataDto extends BaseDto {
...
+   * Does not contain any markup but might be empty
+   * @example Everything I want to buy
+   */
+  @IsString()
+  @ApiProperty()
+  description: string;
+
+  /**
+   * List of tags assigned to this note
+   * @example "['shopping', 'personal']
```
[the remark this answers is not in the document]
mrdrogdrog: the tags never contain markup. :thinking:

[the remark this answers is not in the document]
DerMolly: But the description have?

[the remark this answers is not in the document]
mrdrogdrog: No.

InnayTool (approved):

davidmehren (approved):

backend/src/revisions/revisions.service.spec.ts
```diff
@@ -197,13 +198,122 @@ describe('RevisionsService', () => {
...
+              "username": "mockusername",
+            },
+          ],
+          "id": 3246,
+          "length": 1854,
+          "patch": "mockPatch",
+          "tags": [
+            "mockTag",
+          ],
+          "title": "mockTitle",
```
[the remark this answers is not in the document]
ErikMichelson: But that seems rather complicated since one is a JS object, and this is an inline snapshot only as a string.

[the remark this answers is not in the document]
mrdrogdrog: I chose the snapshot because they're better with promises, dates and more complicated data structures.

backend/src/revisions/utils/extract-revision-metadata-from-content.ts
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * Parses the frontmatter of the given content and extracts the metadata that are necessary to create a new revision..
+ *
+ * @param {string} content the revision content that contains the frontmatter.
+ */
+export function extractRevisionMetadataFromContent(content: string): {
+  title: string;
+  description: string;
+  tags: string[];
```
ErikMichelson, line 26: Shouldn't this be declared as a type for consistency?

mrdrogdrog: What? The string array?

ErikMichelson (approved):

PR #3781, opened 2023-04-12
fix: suppress yjs double import warning

mrdrogdrog:
### Component/Part
YJS Import

### Description

This warning should indicate that you mix cjs and mjs imports. While this is a nice feature it has a weird side effect: Jest triggers this check and floods the console every time a test imports yjs.
This happens between all test suites and every time you call `jest.resetModules()`.
Nothing breaks here. Jest does what it should.
Therefore this PR suppresses this warning in test mode.

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/.yarn/patches/yjs-npm-13.5.52-a5c64d9068.patch
+++ b/.yarn/patches/yjs-npm-13.5.52-a5c64d9068.patch
@@ -1,0 +1,33 @@
+# This warning should indicate that you mix cjs and mjs imports.
+# While this is a nice feature it has a weird side effect:
+# Jest triggers this check and floods the console every time a test imports yjs.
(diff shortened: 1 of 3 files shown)


.yarn/patches/yjs-npm-13.5.52-a5c64d9068.patch
ErikMichelson, line 1: Please insert a [comment](https://unix.stackexchange.com/a/618787) in the patch file with roughly the same content as your PR description. This will save future developers some time to understand why this patch is required or at least better for us.

mrdrogdrog: I inserted it into the commit

DerMolly (approved):

PR #3805, opened 2023-04-14
improve S3 backend

mrdrogdrog:
### Component/Part
S3 backend

### Description

The minio lib already handles port fallbacks.
It also expects "endpoint" to be a single host instead of a path. (see https://github.com/minio/minio-js/blob/master/src/main/minio.js#L269)

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/backend/src/media/backends/s3-backend.ts
+++ b/backend/src/media/backends/s3-backend.ts
@@ -59,7 +59,7 @@ export class S3Backend implements MediaBackend {
     }
   }
 
-  async deleteFile(fileName: string, _: BackendData): Promise<void> {
+  async deleteFile(fileName: string): Promise<void> {
     try {
       await this.client.removeObject(this.config.bucket, fileName);
       const url = this.getUrl(fileName);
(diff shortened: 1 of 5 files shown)


backend/src/media/backends/s3-backend.ts
```diff
@@ -25,23 +25,24 @@ export class S3Backend implements MediaBackend {
...
+      useSSL: isSecure,
+      accessKey: this.config.accessKeyId,
+      secretKey: this.config.secretAccessKey,
+    });
+  }
+
+  private determinePort(url: URL): number | undefined {
+    const port = parseInt(url.port);
+    return isNaN(port) ? undefined : port;
   }
```
ErikMichelson, line 46: Really simple method, but for the coverage a unit-test would be good

ErikMichelson (changes requested):

backend/src/media/backends/s3-backend.ts
mrdrogdrog: It's a private method. Therefore i would need to add unit tests for the whole class. That's not a bad thing, but could take some more time.

mrdrogdrog: I've added a unit test for the s3-backend

DerMolly (approved):

ErikMichelson (approved):

Issue #2953, opened 2022-07-07
Show count of guest users on a revision


ErikMichelson:
**Which part of the project should be enhanced?**
Revision viewer

**Describe the solution you'd like**
The current revision viewer shows an icon per user that participated in that revision (see screenshot).
There's also a field `anonymousAuthorCount` in the API response that could be used to show some generic people icon with a hover title of "x guests".

![image](https://user-images.githubusercontent.com/52606896/177804186-fe325690-b574-43a7-8c8c-52f32ca79b57.png)

**Describe alternatives you've considered**
Only show non-guest users in the revision list

**Additional context**
none

ErikMichelson:
This has been implemented with #3817.

PR #3813, opened 2023-04-15
Only show non-guest users in the sidebar

GGH-WeyboZetO:
### Component/Part
Frontend

### Description
Only show non-guest users in the sidebar

### Steps
- [x] Added implementation
  made sure that:
  - Signed-off-by: Shan Michel <yuplasmas@gmail.com>

### Related Issue(s)
#2953


mrdrogdrog:
Hello,

thanks for your contribution, but I guess you misunderstood something here. With "sidebar" @ErikMichelson meant the sidebar in the revisions dialog. Or in other words: Don't list the guest users in the list items.

While this would be the easiest solution, it isn't the preferred one. The alternative in an issue should only be implemented if the requested feature/improvement is not possible.

GGH-WeyboZetO:
Hello,

I think I understand now what he meant now.
REvision
I'm doing my tests with yarn start:dev with localhost but I don't see any "mock icon" like that:
hw
How can I get something similar with 1-2 mock icons or have that feature not been deleted?
The issue is to get the name of the user displayed when we hover our mouse over the icon right?

Thank you.

mrdrogdrog:
> I'm doing my tests with yarn start:dev with localhost but I don't see any "mock icon" like that:

that's because the backend doesn't save author information right now. It simply just doesn't know who edited what. 

>  or have that feature not been deleted

You can't. if you run the frontend in mock mode then there is NO data persistence. If you need other data then you have to change the mock data in `src/pages/api/private/notes/feature/revisions/`.


> The issue is to get the name of the user displayed when we hover our mouse over the icon right

No. That is already working. The original issue #2953 is about showing how many guest users have participated in a revision.

GGH-WeyboZetO:
I tried to git grep "revisions" and got that result but since it's a frontend issue I guess none of them are right since they are all backend.
grepdd
Can I ask where exactly is located the done doing thoses boxes?
lcasx

GGH-WeyboZetO:
modifnombreauteur

frontend/src/components/editor-page/document-bar/revisions/revision-list-entry.tsx
```diff
@@ -74,6 +75,10 @@ export const RevisionListEntry: React.FC<RevisionListEntryProps> = ({ active, on
         </ShowIf>
         <ShowIf condition={!revisionAuthors.error && !revisionAuthors.loading}>{revisionAuthors.value}</ShowIf>
       </span>
+      <span>
+        <UiIcon icon={IconPersonPlus} className='mx-2' />
+        <Trans i18nKey={'Number of author'} />: {revision.anonymousAuthorCount}
```
mrdrogdrog, line 80: We don't use actual sentences as i18n keys but fqdn like in line 69.

mrdrogdrog (changes requested):
Please squash your commits.

Issue #3273, opened 2023-02-05
Validate YAML frontmatter


mrdrogdrog:
### Which part of the project should be enhanced?

Frontend Frontmatter Parser

### Is your enhancement request related to a problem? Please describe.

Right now we just parse the YAML frontmatter and type cast it. This could lead to problems and possible security issues if users e.g. insert a list in the title field. 

### Describe the solution you'd like

Use a scheme validator like JOI to validate this external user input.

hsedr:
Hey @mrdrogdrog,

we are a group of students who want to contribute to this project. This issue seems like an interesting starting point, but we need further clarification in order to understand your goal here. 

As of now we understand that the `load`-Function, which is used to parse the raw frontmatter into a an Object, poses a security issue, as nested JS-Scripts could be executed while parsing it.
So the goal here should be to avoid this function and instead use the JOI-validator?

Are we on the right track here, or did we miss something? Further informationen would be very appreciated.

best regards

mrdrogdrog:
Hello,

almost right. The yaml load function doesn't execute parsed code and it also doesn't produce functions in the resulting object. The problem here isn't arbitrary code execution but just type safety.
Because of the used types our code expects e.g. the title to be a string. But if someone writes a frontmatter where the title is a list of numbers then this list is passed to the processing code and at some point it might crash.

So yes, a JOI validation would be nice.

mrdrogdrog:
@hsedr Are you still working on this?

hsedr:
Hi, yes we are working on it. What deadline should we retain?

mrdrogdrog:
Already solved by #4107

PR #3835, opened 2023-04-17
Frontmatter validator

hsedr:
### Component/Part
Frontend -> Frontmatter

### Description
This PR adds a JOI scheme validator to validate external user input/frontmatter.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
#3273 

Sorry, I messed up the last pull request. Rookie Mistake.


hsedr:
Regarding the DCO Checks. 
When I'm following the instructions locally too many commits are signed off/changed. 
The fork synchronization was done on Github, that's why the commits are signed off incorrectly.

What is the right measure I should undertake here.

Best Regards

mrdrogdrog:
The sign-off has nothing to do with cryptographically signing using gpg and isn't changed by github.

The sign-off is a line in the commit message that states that you've read the [developer certificate of origin](https://github.com/hedgedoc/hedgedoc/blob/develop/docs/content/legal/developer-certificate-of-origin.txt) and agree to it. If you follow the [sign off instructions](https://github.com/hedgedoc/hedgedoc/pull/3835/checks?check_run_id=12814477067) it should be okay.

mrdrogdrog:
Please drop the commit with the .env file and the vscode file. They're out of scope for this PR.

seredenkodenis:
@mrdrogdrog Thanks for your help! We will do it now

frontend/package.json
```diff
@@ -55,6 +55,7 @@
     "@redux-devtools/core": "3.13.1",
     "@reduxjs/toolkit": "1.9.4",
     "@svgr/webpack": "7.0.0",
+    "@types/joi": "^17.2.3",
```
mrdrogdrog, line 58: Not necessary. The JOI package provides its own types.

frontend/package.json
```diff
@@ -80,6 +81,7 @@
     "i18next-browser-languagedetector": "7.0.1",
     "i18next-resources-to-backend": "1.1.3",
     "isomorphic-ws": "5.0.0",
+    "joi": "^17.9.1",
```
mrdrogdrog, line 84: We use pinned dependencies.

mrdrogdrog (changes requested):

seredenkodenis:
@mrdrogdrog Thanks for your help! I dropped commit and pinned dependency. Am I right that we have to run DCO now?

frontend/src/redux/note-details/raw-note-frontmatter-parser/parser.ts
```diff
@@ -51,6 +52,40 @@ const parseRawNoteFrontmatter = (rawData: RawNoteFrontmatter): NoteFrontmatter =
...
+  robots: Joi.string().optional(),
+  lang: Joi.string().optional(),
+  dir: Joi.string().optional(),
+  breaks: Joi.boolean().optional(),
+  GA: Joi.string().optional(),
+  disqus: Joi.string().optional(),
+  license: Joi.string().optional(),
+  type: Joi.string().optional(),
+  slideOptions: Joi.object().allow(null).pattern(/.*/, Joi.string().allow('')),
+  opengraph: Joi.object().allow(null).pattern(/.*/, Joi.string().allow(''))
```
mrdrogdrog, line 71: Why should objects validate against a regex pattern?
You could verify their structure as well. 
There is also a type for SlideOptions in our code. 

Is there a reason why you're allowing null here instead of optional() ?

mrdrogdrog (changes requested):

frontend/src/redux/note-details/raw-note-frontmatter-parser/parser.ts
seredenkodenis: Hi, sorry for no response. 
I am not sure but there is one test that checks that title is null (when setted by default). And optional checks only if field undefined. 

So if you provide null to title and check it with joy => then error

mrdrogdrog: which test?

mrdrogdrog: anyway. slideOptions and opengraph shouldnt be empty. They should be optional.  I don't know why the rawfrontmatter type accepts null. It should be optional. Maybe we need to change the types a bit. But not in this PR. anyway. Empty string shouldn't be a valid value for opengraph or slideoptions because they're optional objects.

mrdrogdrog:
Hello, how is the status here? :)

hsedr:
Regarding the `opengraph`-field. I couldn't find any specification defining which key fields the opengraph could or should contain. 
The only fields that are used and tested are `title` and `image`. Additionally, I added the `unknown` option to the opengraph because key fields like "image:type" would otherwise result in an error.  

Test2

This test was run without the `unknown`-Option.

A safer way would be of course to define a schema for the opengraph protocol. What are your thoughts on this?

mrdrogdrog:
You messed up the sign-off again and try to merge 80 commits.

mrdrogdrog:
The unknown option is right here. But the properties that are known should be checked.

seredenkodenis:
I am so sorry, we have no experience with DCO. How can we fix all this problems?

mrdrogdrog:
How do you usually sign-off? What command are you using?

seredenkodenis:
git commit -s -m ...

mrdrogdrog:
Well. That signs off the commit you're creating in that moment. But you somehow change the other commits as well.

seredenkodenis:
By @hsedr after pull requests everytime there is merge commit and it is not signed:(

mrdrogdrog:
I fixed your branch. You should delete your LOCAL branch and pull it again to avoid conflicts.

Don't use force push if it's not necessary.
Always pull before you start working. Otherwise git needs to create merge commits between your local commits and the remote commits.

mrdrogdrog:
also: as long as you're working on the branch it's okay to have these merge commits. You can clean this up later.

seredenkodenis:
Thank you very much for your help!

mrdrogdrog:
Hey, I don't wanna put any pressure, but do you have an idea when you gonna work on this again? Do you need help?
Because the JOI scheme is blocking a feature we've planned for the 2.0 alpha release :/

mrdrogdrog:
Already solved by https://github.com/hedgedoc/hedgedoc/pull/4107

PR #3853, opened 2023-04-20
fix: sticky position of toc sidebar

mrdrogdrog:
### Component/Part
Frontend

### Description
This PR fixes the sticky position of the toc sidebar

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/frontend/global-styles/index.scss
+++ b/frontend/global-styles/index.scss
@@ -30,18 +30,6 @@ body {
   background-color: $dark;
 }
 
-#__next {
-  height: 100vh;
-}
-
-html {
-  height: 100%;
-}
-
-body {
-  min-height: 100%;
-}
-
 *:focus {
   outline: 0 !important;
 }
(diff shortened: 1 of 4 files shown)


DerMolly (approved):

frontend/global-styles/index.scss
```diff
@@ -30,18 +30,6 @@ body {
...
-  height: 100vh;
-}
-
-html {
-  height: 100%;
-}
-
-body {
-  min-height: 100%;
-}
```
ErikMichelson, line 43: Why did we add this in the first place? Does something break when removing this?

Blame only shows the initial squash commit from the original private repo for the frontend.

mrdrogdrog: Because we needed it but now we dont.

Seems like no.

mrdrogdrog: IIRC this is either from the beginning of the bootstrap 3 usage when we recreated the first layouts.. or from the old HD code. anyway. I checked the pages and they still look okay

ErikMichelson (approved):

PR #3924, opened 2023-05-01
Fix premature note cleanup on error

### LINKED ISSUE #1992: Online status bug
### Description

Online status show I'm Online, when I click on it I don't see my name

### Steps to reproduce

Since the version 1.9.2 we have several troubles coworking on hedgedoc documents: updates sometimes are done on other user ones, you can loose your updates :'(
It's hard to reproduce and we can't see anything interesting in the logs.
We have noticed that the Online icon can be blue with the user numbers but we can be not in the list. We have to refresh the page (F5) to be "really" online. This can be the origin of our problems: you think you are online (blue icon), but you are not. You work on the doc, changes appears on your page, but you are not online and they are lost.
Sometimes F5 is not enough...
 Is it possible to make the Online icon red If we are not Online and refuse updates if it is the case ?

### Expected behaviour

Not lose my work when I have the Online icon

### Logs

_No response_

### Config

_No response_

### Your Setup

- Host OS: Debian bullseye
- NodeJS version: v16.13.2
- HedgeDoc version:  1.9.2

### Additional context

_No response_

### LINKED ISSUE #3894: Connection forbidden error causes inconsistent pad state and data loss
### Description

Several hours of changes to a pad were lost after reloading the page. Some of the connected users reported not seeing some of the changes (while being "online") or reported garbled pad content. HedgeDoc did not save any revisions during this period. At the time of the last revision, HedgeDoc logged the first error message (see Logs).

### Steps to reproduce

Browser A is logged into Hedgedoc, browser B is not logged in (guest access).

1. Open a pad with "freely" permission in both browsers, wait until it is fully loaded and online
2. Browser B: Use Developer Tools to enable offline mode and wait until the pad is offline
    - Note that in Chrome the Developer Tools Network tab must be open on page load so Websocket connections are properly cut off
3. Browser A: Change pad permission to "restricted"
4. Browser B: Disable offline mode and wait until HedgeDoc connects, receives the permission error and redirects to the index page
    - At this point HedgeDoc logs a "connection forbidden" error
5. Browser A: Add some text to the pad
    - Note that the live preview is not updated (in side-by-side view)
6. Browser A: Reload the page
    - Note that the changes are missing and the pad has the content from before the "connection forbidden" error

If a third (authenticated) browser is connected to the pad, the behaviour gets weirder: After the "connection forbidden" error both browsers are "online" and allow changes to the pad, but the changes are not visible in the other browser. If one reloads the page, cursor movements and changes are visible again, however, the pad content is inconsistent. Some changes are visible, others are not. Some changes cause a "Both operations have to have the same base length" error. The live preview updates on changes, but its content is inconsistent with the Markdown input.

### Expected behaviour

The permission error should not interrupt and break the other existing (and permitted) connections to the pad.

### Logs

```shell
2023-04-24T14:05:04.699Z error:         connection forbidden
2023-04-24T15:17:00.158Z error:         connection forbidden
2023-04-24T15:34:26.965Z error:         Both operations have to have the same base length
```

### Config

_No response_

### Your Setup

- Host OS: Debian 11
- NodeJS version: v16.20.0
- HedgeDoc version: 1.9.7

### Additional context

The steps in "Steps to reproduce" are my attempt at reproducing a "connection forbidden" error. In our case the pad's permission was not changed at all. So I'm not sure what caused the error on our system. My best guess is a session timeout while having the pad open followed by a reconnect due to a bad internet connection.

This is the first time in serveral years that users of our system reported data loss due to this error, although we see this error message about once per month.

This issue might be related to https://github.com/hedgedoc/hedgedoc/issues/1992.

---

I looked into the code where the "connection forbidden" error originates from (in `lib/realtime.js`):

```js
function finishConnection (socket, noteId, socketId) {
  // if no valid info provided will drop the client
  if (!socket || !notes[noteId] || !users[socketId]) {
    return interruptConnection(socket, noteId, socketId)
  }
  // check view permission
  if (!checkViewPermission(socket.request, notes[noteId])) {
    interruptConnection(socket, noteId, socketId)
    return failConnection(403, 'connection forbidden', socket)
  }
  // [...]
```

I believe the `interruptConnection` function is what causes this bug:

```js
function interruptConnection (socket, noteId, socketId) {
  if (notes[noteId]) delete notes[noteId]
  if (users[socketId]) delete users[socketId]
  if (socket) { clearSocketQueue(connectionSocketQueue, socket) } else { connectionSocketQueue.shift() }
  connectNextSocket()
}
```

`interruptConnection` is called when a SocketIO connection is established without having the permission to access a pad. It cleans up the user/socket state, but it also cleans up the state for the pad in the `notes` global.

For comparison, the `disconnect` function checks if there are other connected users before deleting `notes[noteId]`:

```js
function disconnect (socket) {
    /* ... */
    // remove note in notes if no user inside
    if (Object.keys(note.users).length <= 0) {
      if (note.server.isDirty) {
        updateNote(note, function (err, _note) {
          if (err) return logger.error('disconnect note failed: ' + err)
          // clear server before delete to avoid memory leaks
          note.server.document = ''
          note.server.operations = []
          delete note.server
          delete notes[noteId]
    /* ... */
```

I guess adding such a check to `interruptConnection` would fix the issue, but I did not verify that.

As far as I understand, both updating the pad state in the database and creating revisions relies on the `notes` global. So removing the pad state from `notes` prematurely could explain the behaviour.

julianrother:
### Component/Part
Backend (realtime)

### Description

This PR fixes that `interruptConnection()` cleans up note state without first checking if other clients are still connected to the note.

The call to `interruptConnection()` in case of a permission error causes the behavior described in #3894. Since `failConnection()` is also called in this case and since it performs the same cleanup as `interruptConnection()`, but with properly checking for other clients, I removed the call to `interruptConnection()`.

It took me some time to understand why the other call to `interruptConnection()` was added in the first place, whether/when it is executed and if it could cause the same problem as the other call:

```js
function finishConnection (socket, noteId, socketId) {
  // if no valid info provided will drop the client
  if (!socket || !notes[noteId] || !users[socketId]) {
    return interruptConnection(socket, noteId, socketId)
  }
  // ...
```

`!socket` can never be true otherwise the function call `finishConnection(socket, noteId, socket.id)` would have already crashed. `!notes[noteId]` can never be true, because right before the call to `finishConnection()`, `notes[noteId]` is either populated or it is estabilshed that `notes[noteId]` is already populated.

However, if the client is the first client for a note, `startConnection()` fetches the note object asynchronously. If the client disconnects before the promise is resolved, `disconnect()` would have cleaned up `users[socketId]` before the call to `finishConnection()` and the third condition would be true. This also matches the [commit that introduced this check](https://github.com/hedgedoc/hedgedoc/commit/d3a23ad72f1cdde4e7b87e8d79c678b27d98b8c2) ("Fixed realtime.js finishConnection user might be undefined issue"). So in this case cleanup of the `notes` global and `clearSocketQueue()`/`connectNextSocket()` are neccesary.

Anyway, I don't think this call to `interruptConnection()` (and the unconditional `delete notes[noteId]`) does any harm, since it only happens if the client is the first client for a note, but I couldn't resist refactoring it.

### Steps

- [x] Added implementation
- [x] Added changelog snippet
  made sure that:

### Related Issue(s)
Fixes #3894, probably fixes #1992

### DIFF
--- a/public/docs/release-notes.md
+++ b/public/docs/release-notes.md
@@ -10,6 +10,9 @@ You now need Node 16 to run HedgeDoc. We don't support more recent versions of N
 - Allow setting of `documentMaxLength` via `CMD_DOCUMENT_MAX_LENGTH` environment variable.
 - Add dedicated healthcheck endpoint at /_health that is less resource intensive than /status.
 
+### Bugfixes
+- Fix that permission errors can break existing connections to a note, causing inconsistent note content and changes not being saved
+
 ## <i class="fa fa-tag"></i> 1.9.7 <i class="fa fa-calendar-o"></i> 2023-02-19
 
 ### Bugfixes
--- a/lib/realtime.js
+++ b/lib/realtime.js
@@ -363,13 +363,6 @@ function connectNextSocket () {
   }, 1)
 }
 
-function interruptConnection (socket, noteId, socketId) {
-  if (notes[noteId]) delete notes[noteId]
-  if (users[socketId]) delete users[socketId]
-  if (socket) { clearSocketQueue(connectionSocketQueue, socket) } else { connectionSocketQueue.shift() }
-  connectNextSocket()
-}
-
 function checkViewPermission (req, note) {
   if (note.permission === 'private') {
     if (req.user && req.user.logged_in && req.user.id === note.owner) { return true } else { return false }
@@ -386,13 +379,8 @@ let isDisconnectBusy = false
 const disconnectSocketQueue = []
 
 function finishConnection (socket, noteId, socketId) {
-  // if no valid info provided will drop the client
-  if (!socket || !notes[noteId] || !users[socketId]) {
-    return interruptConnection(socket, noteId, socketId)
-  }
   // check view permission
   if (!checkViewPermission(socket.request, notes[noteId])) {
-    interruptConnection(socket, noteId, socketId)
     return failConnection(403, 'connection forbidden', socket)
   }
   const note = notes[noteId]
@@ -460,6 +448,13 @@ function startConnection (socket) {
       },
       include
     }).then(function (note) {
+      // if client disconnected while we waited for the note, disconnect() cleaned up users[socket.id]
+      if (!users[socket.id]) {
+        clearSocketQueue(connectionSocketQueue, socket)
+        connectNextSocket()
+        return
+      }
+
       if (!note) {
         return failConnection(404, 'note not found', socket)
       }


mrdrogdrog (approved):
Thanks for your contribution. It looks good and works.

Also congratulation for surviving the HedgeDoc 1.x spaghetti code. :yum:

public/docs/release-notes.md
```diff
@@ -7,6 +7,9 @@
 - Allow setting of `documentMaxLength` via `CMD_DOCUMENT_MAX_LENGTH` environment variable.
 - Add dedicated healthcheck endpoint at /_health that is less resource intensive than /status.
 
+### Bugfixes
+- Fix that permission errors can break existing connections to a pad, cause inconsistent pad content and changes not being saved
```
davidmehren, line 11:
```suggestion
- Fix that permission errors can break existing connections to a note, causing inconsistent note content and changes not being saved
```

davidmehren (changes requested):
Just nitpicking on the changelog entry, everything else looks very sensible 👍🏻

DerMolly (approved):

InnayTool (approved):

mrdrogdrog (changes requested):
Please squash the docs commit into the other one.

julianrother:
Thank you for the feedback! I've now squashed the docs commit and rebased everything onto the current master, so feel free to merge :)

Do you have plans for a 1.x release with this fix?

mrdrogdrog (approved):

mrdrogdrog:
Maybe today. A new 1.x release is overdue.

davidmehren (approved):
Thank you for contributing!

PR #4024, opened 2023-05-11
refactor: permissions

mrdrogdrog:
### Component/Part
Backend

### Description
This PR refactors the permission service and many things that are connected to it.
Check commit messages for more information.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:


backend/src/api/private/notes/notes.controller.ts
```diff
@@ -261,9 +261,9 @@ export class NotesController {
     return await this.noteService.toNotePermissionsDto(returnedNote);
   }
 
-  @UseInterceptors(GetNoteInterceptor)
-  @Permissions(Permission.OWNER)
   @UseGuards(PermissionsGuard)
```
DerMolly, line 264: I'm just noticing this is not needed here as we already have the annotation at the controller level…

backend/src/permissions/permissions.guard.ts
```diff
@@ -0,0 +1,95 @@
...
+    context: ExecutionContext,
+  ): RequiredPermission | undefined {
+    const requiredAccessLevel = this.reflector.get<RequiredPermission>(
+      permissionMetadataKey,
+      context.getHandler(),
+    );
+    // If no requiredAccessLevel are set this is probably an error and this guard should not let the request pass
+    if (!requiredAccessLevel) {
+      this.logger.error(
+        'Could not find requiredAccessLevel metadata. This should never happen. If you see this, please open an issue at https://github.com/hedgedoc/hedgedoc/issues',
```
DerMolly, line 75: Maybe we want to extract the "This should never happen"-text as a variable for easy changes in the future?

mrdrogdrog: Huh sounds like a good idea

DerMolly (approved):

davidmehren (changes requested):

davidmehren (approved):

PR #4035, opened 2023-05-13
fix(frontend): new note button doesn't use /new anymore

DerMolly:
### Component/Part
new note button

### Description
This led to problems, if user clicked the back button in their browser. This implementation doesn't add functional routes in between new notes, but pushes the new note directly in the history.

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/frontend/src/components/common/new-note-button/new-note-button.tsx
+++ b/frontend/src/components/common/new-note-button/new-note-button.tsx
@@ -3,10 +3,12 @@
  *
  * SPDX-License-Identifier: AGPL-3.0-only
  */
+import { createNote } from '../../../api/notes'
 import { cypressId } from '../../../utils/cypress-attribute'
+import { useUiNotifications } from '../../notifications/ui-notification-boundary'
 import { IconButton } from '../icon-button/icon-button'
-import Link from 'next/link'
-import React from 'react'
+import { useRouter } from 'next/router'
+import React, { useCallback } from 'react'
 import { FileEarmarkPlus as IconPlus } from 'react-bootstrap-icons'
 import { Trans } from 'react-i18next'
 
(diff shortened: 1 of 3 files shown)


frontend/src/components/common/new-note-button/new-note-button.tsx
```diff
@@ -3,22 +3,39 @@
...
-        <Trans i18nKey='navigation.newNote' />
-      </IconButton>
-    </Link>
+    <IconButton
+      {...cypressId('new-note-button')}
+      iconSize={1.5}
+      size={'sm'}
+      icon={IconPlus}
+      onClick={() => {
+        void createNewNoteAndRedirect()
```
mrdrogdrog, line 36: Don't use void here! If anything goes wrong then the whole app crashes the error isn't covered. Either use `useAsyncFn` or make the callback not async.

mrdrogdrog (changes requested):

frontend/src/components/common/new-note-button/new-note-button.tsx
```diff
@@ -3,22 +3,44 @@
...
  * Links to the "new note" endpoint
  */
 export const NewNoteButton: React.FC = () => {
+  const { showErrorNotification } = useUiNotifications()
+  const router = useRouter()
+  const createNewNoteAndRedirect = useCallback((): void => {
+    createNote('')
+      .then((note) => {
+        const to = `/n/${note.metadata.primaryAddress}`
+        router?.push(to).catch((error: Error) => {
```
mrdrogdrog, line 25: Return the promise from the push here without the catch block. The next catch block is the same and will handle it

frontend/src/components/common/new-note-button/new-note-button.tsx
```diff
@@ -3,22 +3,44 @@
...
-        <Trans i18nKey='navigation.newNote' />
-      </IconButton>
-    </Link>
+    <IconButton
+      {...cypressId('new-note-button')}
+      iconSize={1.5}
+      size={'sm'}
+      icon={IconPlus}
+      onClick={() => {
+        void createNewNoteAndRedirect()
```
mrdrogdrog, line 41: No need for the void here or the lambda

mrdrogdrog (changes requested):

frontend/src/components/common/new-note-button/new-note-button.tsx
```diff
@@ -3,22 +3,42 @@
...
-      </IconButton>
-    </Link>
+    <IconButton
+      {...cypressId('new-note-button')}
+      iconSize={1.5}
+      size={'sm'}
+      icon={IconPlus}
+      onClick={() => {
+        createNewNoteAndRedirect()
+      }}>
```
mrdrogdrog, line 40: no need for the lamba. this only creates unnecessary re-renders

mrdrogdrog (changes requested):

mrdrogdrog (approved):

PR #4036, opened 2023-05-13
feat(backend): handle username always in lowercase

DerMolly:
### Component/Part
user 

### Description
This should make all usernames of new users into lowercase. Usernames are also searched in the DB as lowercase.

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/frontend/src/components/login-page/auth/fields/username-field.tsx
+++ b/frontend/src/components/login-page/auth/fields/username-field.tsx
@@ -8,19 +8,26 @@ import React from 'react'
 import { Form } from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
 
+export interface UsernameFieldProps extends AuthFieldProps {
+  value: string
+}
+
+//TODO: This should be replaced with the common username component. See https://github.com/hedgedoc/hedgedoc/issues/4128
 /**
  * Renders an input field for a username.
  *
  * @param onChange Hook that is called when the input is changed.
  * @param invalid True indicates that the username is invalid, false otherwise.
+ * @param value the username value
  */
-export const UsernameField: React.FC<AuthFieldProps> = ({ onChange, invalid }) => {
+export const UsernameField: React.FC<UsernameFieldProps> = ({ onChange, invalid, value }) => {
   const { t } = useTranslation()
 
   return (
     <Form.Group>
       <Form.Control
         isInvalid={invalid}
+        value={value}
         type='text'
         size='sm'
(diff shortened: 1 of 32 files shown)


ErikMichelson (commented):
Maybe it's nice to convert uppercase letters to lowercase in the register page while typing as well?

davidmehren:
This has potential for unintuitive behavior, when the username is magically converted without a user noticing.

IMO, the proper way is to reject usernames with capital letters and return a corresponding error.

mrdrogdrog:
This could also clash with e.g. LDAP.

DerMolly:
> This has potential for unintuitive behavior, when the username is magically converted without a user noticing.

How so? It's common that `DerMolly`, `dermolly` and `DeRmOlLy` reference the same account, is it not?

I think it's a good idea to implement that the user input gets transformed on the frontend's side, so that the user directly sees that different capitalization don't work. This would affect:
- username on internal login
- user search in the permissions

> IMO, the proper way is to reject usernames with capital letters and return a corresponding error.

Given we transform all names in the frontend, we could respond with errors if people hold the api the wrong way.

davidmehren:
According to my google-fu, LDAP `uid`s are case-insensitive by default. So *most* LDAPs should not be affected.

mrdrogdrog:
We agreed on the following points:
  - Users can't be created if they have usernames with uppercase letters. The API should throw an error.
  - LDAP username will be lowercased on creation
  - The frontend should lowercase usernames automatically while typing

davidmehren (approved):

frontend/src/components/login-page/auth/fields/username-field.tsx
```diff
@@ -8,19 +8,26 @@ import React from 'react'
 import { Form } from 'react-bootstrap'
 import { useTranslation } from 'react-i18next'
 
+export interface UsernameFieldProps extends AuthFieldProps {
+  value: string
+}
+
+//TODO: This should be replaced with the common username component
```
ErikMichelson, line 15: Is there an issue for this?

ErikMichelson (commented):
Otherwise LGTM

frontend/src/components/login-page/auth/fields/username-field.tsx
mrdrogdrog: https://github.com/hedgedoc/hedgedoc/issues/4128

ErikMichelson (approved):

InnayTool (approved):

Issue #4128, opened 2023-05-31
Deduplicate username field component


mrdrogdrog:
### Which part of the project should be enhanced?

Frontend

### Is your enhancement request related to a problem? Please describe.

Right now we have two components for usernames. One in `frontend/src/components/login-page/auth/fields/username-field.tsx` and one in `frontend/src/components/common/fields/username-field.tsx` .

### Describe the solution you'd like

Deduplicate the code to only use the component in commons.

### Describe alternatives you've considered

-

PR #4137, opened 2023-06-01
refactor: deduplicated username field

Avi98:
### Component/Part
Frontend

### Description
This PR fixes [#4128](https://github.com/hedgedoc/hedgedoc/issues/4128)

### Steps

- [x] Added implementation
- [ ] Added / updated tests
- [ ] Added changelog snippet
  made sure that:

### Related Issue(s)

[#4128](https://github.com/hedgedoc/hedgedoc/issues/4128)


mrdrogdrog (changes requested):
Your implementation is way to complicated for what it should do.

1. The username component is just wrapping the form.control component which allows both `isInvalid` and `isValid` at the same time. Just make them optional and let the actual form.control element handle the case if both are provided which will probably never happen. 
2. Remove the conditional show of labels. The correct react way-of-doing is to create an intermediate component that uses the username component together with the texts.

Avi98:
Got the first point. 

> 2. Remove the conditional show of labels. The correct react way-of-doing is to create an intermediate component that uses the username component together with the texts. The conditional show should only be used if a value could actually change.

For this do i need to create a intermediate component as we are using label only in register component. Can i add label in register form only.

mrdrogdrog:
The idea is that this intermediate component has the labels in any case and is used in register.tsx

frontend/src/components/common/fields/username-field.tsx
```diff
@@ -4,43 +4,39 @@
...
+
+export interface UsernameFieldProps extends CommonFieldProps {
+  isInvalid?: boolean
+  isValid?: boolean
+}
 
 /**
  * Renders an input field for the username when registering.
  *
  * @param onChange Hook that is called when the entered username changes.
```
mrdrogdrog, line 19: onChange isn't a hook. It's a callback.

frontend/src/components/common/fields/username-label-field.tsx
```diff
@@ -0,0 +1,32 @@
...
+import { UsernameField } from './username-field'
+import type { UsernameFieldProps } from './username-field'
+import React from 'react'
+import { Form } from 'react-bootstrap'
+import { Trans } from 'react-i18next'
+
+/**
+ * Wraps and contains label and info for UsernameField
+ *
+ * @param onChange Hook that is called when the entered username changes.
```
mrdrogdrog, line 15: see above

mrdrogdrog (changes requested):

mrdrogdrog (approved):

frontend/src/components/common/fields/username-label-field.tsx
```diff
@@ -0,0 +1,32 @@
...
+
+/**
+ * Wraps and contains label and info for UsernameField
+ *
+ * @param onChange Callback that is called when the entered username changes.
+ * @param value The currently entered username.
+ * @param isValid  Is a valid field or not
+ * @param isInvalid Adds error style to label
+ */
+export const UsernameLabelField = (props: UsernameFieldProps) => {
```
ErikMichelson, line 20: Please add a type definition.

```suggestion
export const UsernameLabelField: React.FC<UsernameFieldProps> = (props: UsernameFieldProps) => {
```

frontend/src/pages/register.tsx
```diff
@@ -86,7 +86,7 @@ export const RegisterPage: NextPage = () => {
             <Card className='bg-dark mb-4 text-start'>
               <Card.Body>
                 <Form onSubmit={doRegisterSubmit} className={'d-flex flex-column gap-3'}>
-                  <UsernameField onChange={onUsernameChange} value={username} />
+                  <UsernameLabelField onChange={onUsernameChange} value={username} isValid={Boolean(username.trim())} />
```
ErikMichelson, line 89: `Boolean(username.trim())` is not a statically evaluated expression.
To improve performance, it should therefore be extracted and wrapped in a `useMemo`.

ErikMichelson (changes requested):

frontend/src/components/common/fields/username-label-field.tsx
Avi98: Thanks @ErikMichelson, done

frontend/src/pages/register.tsx
Avi98: sure @ErikMichelson, done

mrdrogdrog:
Rebase please

ErikMichelson (approved):

mrdrogdrog (approved):

mrdrogdrog:
Thanks for your contribution!

Issue #4306, opened 2023-06-25
Backend log flushing behavior is confusing


davidmehren:
### Description

When the HD2 backend cannot reach its database, it does not print any output for about a minute. Then, all error messages are dumped at the same time.

### Steps to reproduce

- Run HD2 in a docker setup where the backend cannot reach the database
- Observe no log output for a minute

### Expected behaviour

The log starts to print immediately after the container is started and is not buffered somewhere.

### Logs

```shell
hedgedoc2-backend-1   | 2023-06-25T18:00:17.790303124Z [Nest] 7  - 06/25/2023, 6:00:17 PM     LOG [NestFactory] Starting Nest application...
hedgedoc2-backend-1   | 2023-06-25T18:00:17.790977383Z [Nest] 7  - 06/25/2023, 6:00:17 PM     LOG [InstanceLoader] TypeOrmModule dependencies initialized +2ms
hedgedoc2-backend-1   | 2023-06-25T18:00:17.791196034Z [Nest] 7  - 06/25/2023, 6:00:17 PM     LOG [InstanceLoader] AuthorsModule dependencies initialized +0ms
hedgedoc2-backend-1   | 2023-06-25T18:00:17.791286124Z [Nest] 7  - 06/25/2023, 6:00:17 PM     LOG [InstanceLoader] MonitoringModule dependencies initialized +1ms
hedgedoc2-backend-1   | 2023-06-25T18:00:17.791308581Z [Nest] 7  - 06/25/2023, 6:00:17 PM     LOG [InstanceLoader] PassportModule dependencies initialized +0ms
hedgedoc2-backend-1   | 2023-06-25T18:00:17.791436170Z [Nest] 7  - 06/25/2023, 6:00:17 PM     LOG [InstanceLoader] RouterModule dependencies initialized +0ms
hedgedoc2-backend-1   | 2023-06-25T18:00:17.791462477Z [Nest] 7  - 06/25/2023, 6:00:17 PM     LOG [InstanceLoader] DiscoveryModule dependencies initialized +0ms
hedgedoc2-backend-1   | 2023-06-25T18:00:17.791565200Z [Nest] 7  - 06/25/2023, 6:00:17 PM     LOG [InstanceLoader] ConfigHostModule dependencies initialized +0ms
[...]

PR #4308, opened 2023-06-25
fix(backend): immediately use our custom logger

davidmehren:
### Component/Part
app init

### Description

While the DI and database initialization is running, NestJSs default logger is normally used. Our custom logger was only being initialized after DI setup is complete.
 Errors encountered during DI setup were buffered and only printed after DI init was complete, or the app exited on error.
 This led to the app not printing anything for a minute in certain cases.

 This commit replaces the initial logger with our ConsoleLoggerService that logs everything.
 After DI init is complete, that logger is replaced with a normal instance of ConsoleLoggerService that uses the real config from DI.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
Fixes https://github.com/hedgedoc/hedgedoc/issues/4306

### DIFF
--- a/backend/src/main.ts
+++ b/backend/src/main.ts
@@ -3,7 +3,6 @@
  *
  * SPDX-License-Identifier: AGPL-3.0-only
  */
-import { LogLevel } from '@nestjs/common';
 import { ConfigService } from '@nestjs/config';
 import { NestFactory } from '@nestjs/core';
 import { NestExpressApplication } from '@nestjs/platform-express';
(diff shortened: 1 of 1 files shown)


DerMolly (approved):

backend/src/main.ts
```diff
@@ -12,14 +11,16 @@ import { setupApp } from './app-init';
...
+import { Loglevel } from './config/loglevel.enum';
 import { MediaConfig } from './config/media.config';
 import { ConsoleLoggerService } from './logger/console-logger.service';
 
 async function bootstrap(): Promise<void> {
   // Initialize AppModule
   const app = await NestFactory.create<NestExpressApplication>(AppModule, {
-    logger: ['error', 'warn', 'log'] as LogLevel[],
-    bufferLogs: true,
+    // ConsoleLoggerService only users the loglevel, so we can give it an incomplete AppConfig to log everything
```
mrdrogdrog, line 21:
```suggestion
    // ConsoleLoggerService only uses the loglevel, so we can give it an incomplete AppConfig to log everything
```

InnayTool (approved):

PR #4398, opened 2023-07-05
Increase Frontend Performance

mrdrogdrog:
### Component/Part
Renderer

### Description
This PR increases the performance of the frontend by using `useDeferredValue`. It's a new hook in React 18 and renders a state change in the background and only renders the result if the browser can handle it right now. If the value changes quickly and often then the result is that not every change is visible.

It also adds performance markers to measure how long the frontend needs to render a markdown text.

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/dev-reverse-proxy/package.json
+++ b/dev-reverse-proxy/package.json
@@ -4,6 +4,7 @@
   "private": true,
   "license": "AGPL-3.0",
   "scripts": {
-    "start:dev": "caddy run || ./caddy run"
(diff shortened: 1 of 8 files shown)


ErikMichelson (approved):

ErikMichelson (changes requested):
`Error: Uncaught [TypeError: performance.mark is not a function]` in jest 👀

ErikMichelson (approved):

DerMolly (approved):

PR #4539, opened 2023-07-30
fix(commons): replace microbundle with bash based compile script

mrdrogdrog:
### Component/Part
Building of commons package

### Description
This PR fixes the package generation of the commons package.

The commons package ships wrong types because it is refering to the same files for the ESM and the CJS build.
See https://arethetypeswrong.github.io/?p=%40mrdrogdrog%2Foptional%401.1.0

This happens because microbundle can handle the generation of `.mjs` and `.cjs` from files itself but delegates the generation of types entirely to typescript by running it once. Microbundle uses the "type" field from the package.json to know if a `.js` file is meant to be mjs or cjs and generates the other type by using the specific file extension `.cjs` and `.mjs` (so if your package is a `type: module`, then `.js` file are interpreted as ECMAModule and if you have a commonjs file you need to name it `.cjs`).  But this causes a problem with typescript. If you use typescript with the newer module resolver then it expects the type declarations to be named exactly like the file you wanna import. So if you have a `.js` file it will try to look up types in a `.d.ts` file. If it is resolving a `.mjs` file it is looking for a `.d.mts` file. 

This clashes with the types generated by microbundle because you can't use a `.mjs` file with a `.d.ts` file.

Running typescript multiple times can also be complicated.
When generating type declaration files, typescript takes a look at the source file extension. So a `.mts` file will generate a `.mjs` and a `.d.mts` file. A `.ts` will generate a `.js` and `.d.ts` file. It doesn't matter if you run microbundle on `.ts`, `.mts` or `.cjs` files, it will only generate the type declarations once.

How do you get the other type declaration? To solve this problem you either have to run typescript multiple times and manipulate the input or output data to have correct `.d.mts` / `.d.cts` files AND imports... or do what this PR changes. 

It runs typescript multiple times but places the complied files in different directories. It then places a package.json in both directories which declares if `.js` is commonjs or ESM. 
This way the resolver is happy because it can import `.js` files according to the package.json content and typescript is happy because it can find type declarations. And because package.json files are inheriting properties from other package.json files no necessary file is missing.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
This should solve https://github.com/hedgedoc/hedgedoc/pull/4334

### DIFF
--- a/commons/.eslintrc.cjs
+++ b/commons/.eslintrc.cjs
@@ -8,7 +8,7 @@ module.exports = {
     "parser": "@typescript-eslint/parser",
     "parserOptions": {
         "project": [
-            "./tsconfig.json"
+            "./tsconfig.test.json"
         ]
     },
     "plugins": [
--- a/commons/build.sh
+++ b/commons/build.sh
@@ -1,0 +1,33 @@
+#!/bin/bash
+#
+# SPDX-FileCopyrightText: 2023 The HedgeDoc developers (see AUTHORS file)
+#
+# SPDX-License-Identifier: AGPL-3.0-only
+#
+
+set -e
+
+echo "Clear dist directory.."
+rm -rf dist
+
+echo "Compile to CJS.."
+tsc --project tsconfig.cjs.json
+
+echo "Compile to ESM.."
+tsc --project tsconfig.esm.json
+
+echo "Fix CJS package.json.."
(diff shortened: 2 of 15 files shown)


davidmehren (approved):
Seems to make sense, tested with TS 5.2 Beta.

Issue #3734, opened 2023-04-05
Adding a bullet list to a colored citation make a tag icon appear


yazae:
### Description

Adding a bullet list to a colored citation make a tag icon appear.

### Steps to reproduce

When I try to use a citation who contains a bullet list, and that citation is colored with `[color=anyColor]`, a tag icon appear under the citation, colored with the color of the citation.

Here is the md code :

```md
> Citation with color
> [name=invader][color=purple]

> Citation
> - With bullets without color
> [name=invader]

> Citation
> - With bullets (with a `[color]` added)
> [name=invader][color=purple]
```

And it render as such in HedgeDoc :

![image](https://user-images.githubusercontent.com/88883564/230173206-388296ad-4c45-488c-bf41-3c9efed00012.png)

**Workaround**

As pointed by Tilman on the matrix channel, there is a notation who don’t make the icon tag appear :

```
> Citation
> - With bullets (with a `[color]` added)
> [name=invader]
> 
> [color=purple]
```

### Expected behaviour

The icon tag shouldn’t appear, it doesn’t mean anything, and it doesn’t make sense that it appear only on citation with bullets.

### Your Setup

- Host OS: idk
- NodeJS version: idk
- HedgeDoc version: 1.8.2 (but it is the same at http://demo.hedgedoc.org/ and even in hd2 apparently)

### Additional context

It’s an amazing project, many thanks to it’s mainteners o/

ErikMichelson:
I remember stumbling upon this weird tag icon when I re-implemented this feature for 2.0 (https://github.com/hedgedoc/react-client/pull/249#issuecomment-647626674)

Revisiting this now, I see no real use-case in this and would vote to drop it, therefore making the `[color=...]` syntax only a modifier for a blockquote.

DerMolly:
Hm, the color tags can be used to create legends (red is for $A, blue is for $B) etc. So I think we may want to keep this around.

ErikMichelson:
To avoid more syntax that is incompatible with other markdown parsers, I would rather encourage the use of the specific [unicode "circles" for this](https://stackoverflow.com/a/70616224).

yazae:
> Hm, the color tags can be used to create legends (red is for $A, blue is for $B) etc. So I think we may want to keep this around.

Well, for now it's impossible to know what it's for, and it's inconsistent with the normal blockquote, so it needs work before considering it a feature I think 🤔 
[...]

PR #4570, opened 2023-08-05
fix(blockquote): remove extra color tag

DerMolly:
### Component/Part
extra color tags

### Description
This PR removes the extra color tag in a way that still makes us able to use the color tag in the blockquote

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
#3734


ErikMichelson (approved):

mrdrogdrog (approved):
I would have prefered to replace the color tags with nothing to indicate that they are only used as control mechanism. But I guess this way people see that nothing special happens which is also okay.

DerMolly:
> I would have prefered to replace the color tags with nothing to indicate that they are only used as control mechanism. But I guess this way people see that nothing special happens which is also okay.

I thought of how to handle the case and we parse the color information, because we use it with the blockquote. First I just replaced the tag icon with the text, but that way the text disappeared when a non-valid color was given so this is the best solution I could find.

PR #4733, opened 2023-08-27
fix(frontend): netlify deployment

mrdrogdrog:
### Component/Part
Frontend

### Description
This PR fixes the netlify deployment

### Steps

- [x] Added implementation
  made sure that:

### DIFF
--- a/.github/workflows/frontend-netlify-deploy-main.yml
+++ b/.github/workflows/frontend-netlify-deploy-main.yml
@@ -44,7 +44,7 @@ jobs:
         run: echo "HD_BASE_URL=\"https://hedgedoc.dev/\"" >> .env.production
 
       - name: Build app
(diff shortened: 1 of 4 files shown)


mrdrogdrog:
Mayyybeee... Instead of fixing this we should throw away the netlify deployment and set-up a full-stack-demo (like demo.hedgedoc.org) that resets itself every day or so. Wdyt @ErikMichelson @DerMolly

ErikMichelson:
As already said on matrix, I'd favor such a solution where the instance database is reset once per day.

ErikMichelson:
We did set up a live instance on hedgedoc.dev instead of the netlify deployment. The database will be reset every day as it is a sqlite file in the container and the container gets recreated every day.

Therefore I'm closing this PR.

PR #4757, opened 2023-08-30
fix(oauth2): Fix crash in rolesClaim extraction

SISheogorath:
### Component/Part
auth/oauth2

### Description
This patch adds a try-catch around the rolesClaim extraction to prevent full crashes of HedgeDoc when a user profile is read, that doesn't contain any such claim, which can happen with some IdPs, like Keycloak, that omit the attribute when it's empty.

As a result an authorized user would crash the entire server, which is definitely unintended behaviour. The simply try-catch should resolve the issue and make sure that roles is always defined even if the `extractProfileAttribute` call fails.

### Steps

- [x] Added implementation
- [ ] Added changelog snippet
  made sure that:

### DIFF
--- a/lib/web/auth/oauth2/index.js
+++ b/lib/web/auth/oauth2/index.js
@@ -75,7 +75,13 @@ function checkAuthorization (data, done) {
       logger.error('oauth2: "accessRole" is configured, but "rolesClaim" is missing from the config. Can\'t check group membership!')
     } else {
       // parse and check role data
-      const roles = extractProfileAttribute(data, config.oauth2.rolesClaim)
+      let roles = []
+      try {
+        roles = extractProfileAttribute(data, config.oauth2.rolesClaim)
+      } catch (err) {
+        logger.warn(`oauth2: failed to extract rolesClaim '${config.oauth2.rolesClaim}' from user profile.`)
+        return done('Permission denied', null)
+      }
       if (!roles) {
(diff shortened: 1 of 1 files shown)


davidmehren (approved):

lib/web/auth/oauth2/index.js
```diff
@@ -75,7 +75,12 @@ function checkAuthorization (data, done) {
       logger.error('oauth2: "accessRole" is configured, but "rolesClaim" is missing from the config. Can\'t check group membership!')
     } else {
       // parse and check role data
-      const roles = extractProfileAttribute(data, config.oauth2.rolesClaim)
+      let roles = []
+      try {
+        roles = extractProfileAttribute(data, config.oauth2.rolesClaim)
+      } catch (err) {
+        logger.warn(`oauth2: failed to extract rolesClaim '${config.oauth2.rolesClaim}' from user profile.`)
+      }
```
mrdrogdrog, line 83: Please add a return in the catch block

mrdrogdrog (changes requested):

ErikMichelson:
@SISheogorath What's the state with this PR? :)

SISheogorath:
Looks like I missed the remark, otherwise it should be good to go :)

lib/web/auth/oauth2/index.js
SISheogorath: done :)

ErikMichelson (approved):

mrdrogdrog (approved):

PR #4829, opened 2023-09-10
feat(api/notes): prevent creating new notes with alias equal to an existing note ID

davidmehren:
### Component/Part
notes service

### Description
This PR adds a check to `createNote` to prevent creating new notes with an alias equal to an existing note ID.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### DIFF
--- a/backend/src/notes/notes.service.ts
+++ b/backend/src/notes/notes.service.ts
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: 2022 The HedgeDoc developers (see AUTHORS file)
+ * SPDX-FileCopyrightText: 2024 The HedgeDoc developers (see AUTHORS file)
  *
  * SPDX-License-Identifier: AGPL-3.0-only
  */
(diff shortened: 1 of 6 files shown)


backend/src/notes/notes.service.ts
```diff
@@ -279,6 +285,24 @@ export class NotesService {
...
+    // Check if we already have a note with this id
+    const note = await this.noteRepository
+      .createQueryBuilder()
+      .where('note.publicId = :noteId')
+      .orWhere('note.id = :noteId')
+      .setParameter('noteId', noteId)
+      .getOne();
+    if (note !== null) {
+      this.logger.debug(
+        `A note with the alias '${noteId}' already exists.`,
```
ErikMichelson, line 298: This is misleading and seems to be an implementation error. The method does only check publicIds and internal ids(?) but not aliases.

ErikMichelson (changes requested):

backend/src/notes/notes.service.ts
```diff
@@ -94,6 +94,12 @@ export class NotesService {
     // Check if new note doesn't violate application constraints
     if (alias) {
       this.checkNoteIdOrAlias(alias);
```
InnayTool, line 96: since we have multiple check methods. We should rename this so it is more obvious what is checked.

DerMolly (approved):

PR #5012, opened 2023-10-07
feature: add identicon generation to users without photo

### LINKED ISSUE #1202: Profile picture generation in the frontend
user service, photo generation

**Is your enhancement request related to a problem? Please describe.**
We need to create photos for user that don't have any.

In 1.x we did it this way (thanks to @ErikMichelson for the disection):

HedgeDoc 1.x uses three places where (sometimes different) avatar images are used:
 - The online status indicator
 - The information line above a note (note creator, last editor)
 - The user dropdown on the intro/history page

There is no possibility to upload an own avatar to HedgeDoc 1.x, however the image to use is determined like this:

- When a user account is created, the profile data of the external provider (usually the OAuth2 /userinfo endpoint) is stringified into the `profile` column of the database. Local (email) users have an empty `profile`.
- The `/me` route as well as the realtime code responsible for extracting the user that last changed a note both call the [`getProfile` method](https://github.com/hedgedoc/hedgedoc/blob/dcf72148/lib/models/user.js#L71) of an User object, that tries to parse the database `profile` column and then determines the profile picture from that as following:
    1. The `profile.provider` is set to an auth provider that serves the profile pictures of the user. In that case this will be just taken. (Applies to Facebook, Twitter, GitHub, GitLab, Mattermost, Google)
    2. The `profile.provider` is set to an auth provider that does not serve the users' profile picture OR the user has no profile picture on one of the previous providers OR `profile.provider` is not set. In that case [`generateAvatarURL` method](https://github.com/hedgedoc/hedgedoc/blob/dcf72148/lib/letter-avatars.js#L28) is called, which does the following differentiation:
        1. The user has an email address set (`profile.email`) AND `config.allowGravatar` is enabled. In that case the email address is md5'ed and given to Libravatar for retrieving the picture.
        2. The user has no email address AND `config.allowGravatar` is enabled. In that case the username is prepended before "@example.com" and given to Libravatar.
        3. Otherwise the API route `/user/<username>/avatar.svg` is returned. `<username>` in this case is either the localpart of the email address or the username if no email is set or the md5 hash of the email address (WHY?).

- The API route mentioned above does the following:
    1. Generate a random "dark" color with the username as seed for the generator.
    2. Take the first letter of the username and uppercase it.
    3. Create a SVG file that contains the color as background color and the letter as a text element on it.

- The status list icon inside the editor only checks if a user from the list of users returned by the [websocket action `online users`](https://github.com/hedgedoc/hedgedoc/blob/dcf72148/public/js/index.js#L2528) contains a `photo` property (which it only does if the login provider has provided one, see [`updateUserData` method](https://github.com/hedgedoc/hedgedoc/blob/dcf72148/lib/realtime.js#L610)). If so, it takes this, otherwise it uses a randomly generated color as background-image for the profile picture placeholder container.

---

Generating a data-URI with a svg in it, that shall be used for logged-in users without an image.

Anonymous users could, just give themselves (or rather the frontend give sthem) a random name and a random color. This way the could be clearly distinguished from logged-in users.

It seems reasonable to include a deterministic way of generating images, so a user always get's the same image. As such it probably should be based on the `userName`, because that won't change for a user…

Also we probably won't want to build our own image generation lib, because there are ton out there.

@MrDrogDrog suggested using either

<https://avatars.dicebear.com/> with [human](https://github.com/dicebear/dicebear), [avataaars](https://avataaars.com/) or [bottts](https://bottts.com/) (see [Licenses](https://avatars.dicebear.com/licenses))
**or**
<https://jdenticon.com/>, possibly limiting ourselves to the hedgedoc color scheme (see [here](https://jdenticon.com/))

**Describe alternatives you've considered**
Eh. I don't know. 

**Additional context**

Relevant code segments (1.x Code)
- https://github.com/hedgedoc/hedgedoc/blob/dcf72148cebd817399fc0e6067709281fdf2dac9/lib/web/userRouter.js#L125
- https://github.com/hedgedoc/hedgedoc/blob/dcf72148cebd817399fc0e6067709281fdf2dac9/lib/letter-avatars.js#L8
- https://github.com/hedgedoc/hedgedoc/blob/dcf72148cebd817399fc0e6067709281fdf2dac9/lib/letter-avatars.js#L28
- https://github.com/hedgedoc/hedgedoc/blob/dcf72148cebd817399fc0e6067709281fdf2dac9/lib/models/user.js#L93
- https://github.com/hedgedoc/hedgedoc/blob/dcf72148cebd817399fc0e6067709281fdf2dac9/public/js/cover.js#L74
- https://github.com/hedgedoc/hedgedoc/blob/dcf72148cebd817399fc0e6067709281fdf2dac9/public/js/index.js#L2753
- https://github.com/hedgedoc/hedgedoc/blob/dcf72148cebd817399fc0e6067709281fdf2dac9/public/js/extra.js#L73

ErikMichelson:
### Component/Part
user avatar

### Description
This PR adds a generated identicon (based on the display name) when no avatar is specified.
We cannot use the email address or something alike as we don't have that in many cases.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
fixes #1202


frontend/package.json
```diff
@@ -46,6 +46,8 @@
     "@codemirror/state": "6.2.1",
     "@codemirror/theme-one-dark": "6.1.2",
     "@codemirror/view": "6.20.2",
+    "@dicebear/collection": "7.0.1",
```
davidmehren, line 49: We can save our tree-shaking some work by just using identicon here

DerMolly (approved):

davidmehren (approved):

Issue #5167, opened 2023-10-20
Revision viewer showing error during revision loading


christian-heusel:
### Description

There is an error when you check a notes revision history, although it appears only very briefly.
Also there seems to be some empty field there:
![Screenshots2023-10-20_12:05:18](https://github.com/hedgedoc/hedgedoc/assets/26827864/51e57998-d746-4d39-af53-da02ae2df90c)

### Steps to reproduce

1. Go to any note
2. Make an edit
3. Check the revision history
4. Click on the last revision
5. Observe the above behaviour

### Expected behaviour

No errors for regular user interaction

### Your Setup

The hedgedoc beta setup over at 
https://hedgedoc.dev/

ErikMichelson:
Thanks for reporting.

Two separate things:

1. The empty fields are known. This is caused by #5016 still being open.
2. It seems that during the time of loading the revision from the API, there is no loading spinner but this error is shown. This could indeed be fixed.

ErikMichelson:
More context:
[This error](https://github.com/hedgedoc/hedgedoc/blob/develop/frontend/src/components/editor-page/sidebar/specific-sidebar-entries/revisions-sidebar-entry/revisions-modal/revision-viewer.tsx#L36) is probably fired when the component is initialized for the first time and the props are not evaluated yet. The method is then called again when the props (`noteId`, `selectedRevisionId`) are available, and thus the error vanishes.

We could extend the early component return by checking for `noteId` too and then remove this error by silently ignoring the case that the props are not there yet.

PR #5182, opened 2023-10-22
fix(frontend): error in review viewer

DerMolly:
### Component/Part
review viewer

### Description
This PR fixes the error in review viewer.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
#5167


ErikMichelson (changes requested):
The error is still thrown and displayed for a short period of time.
I guess this could be mitigated by changing the block from lines 35 to 39 to something like:

```typescript
if (noteId && selectedRevisionId !== undefined) {
  return await getRevision(noteId, selectedRevisionId)
}
```

This removes the `throw` which is not necessary as the component does not render anyway when the values are undefined. In contrast to the `noteId` field which is a string and can be checked implicitly, the `selectedRevisionId` is a number and therefore needs the explicit `!== undefined` as `0` evaluates to `false` when being checked implicitly.

ErikMichelson (approved):

mrdrogdrog (approved):

Issue #3727, opened 2023-04-04
Configurable note revision clean-up


ErikMichelson:
### Is your feature request related to a problem? Please describe.

Huge instances with many notes and many users tend to create a huge amount of data with note revisions. One can delete them manually via the UI or the API, however this is a tedious task.

### Describe the solution you'd like

A new config option `HD_REVISION_RETENTION_DAYS` that sets the number of days a revision should be kept. A regular nest cronjob will clean all note revisions that are due for deletion.
If the config option is not set or set to `0`, all revisions will be kept forever.

### Describe alternatives you've considered

Include a CLI tool for the admin to clean-up old note revisions/the database.

yamashush:
> Huge instances with many notes and many users tend to create a huge amount of data with note revisions.

I totally get it.
I have been operating hackmd & codimd for over 5 years.
I have experienced that the system's response has become very slow due to huge revision.

PR #5349, opened 2024-01-14
feat(backend revision): add clean-up note revisions job

yamashush:
### Component/Part

backend revision

### Description

This PR adds note revisions clean-up job.

> A new config option HD_REVISION_RETENTION_DAYS that sets the number of days a revision should be kept. A regular nest cronjob will clean all note revisions that are due for deletion.
If the config option is not set or set to 0, all revisions will be kept forever.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)

- https://github.com/hedgedoc/hedgedoc/issues/3727


backend/src/revisions/revisions.service.ts
```diff
@@ -219,4 +220,39 @@ export class RevisionsService {
...
+
+  // Delete all old revisions 5 sec after startup
+  @Timeout(5000)
+  async handleTimeout(): Promise<void> {
+    return await this.removeOldRevisions();
+  }
+
+  async removeOldRevisions(): Promise<void> {
+    const currentTime = new Date().getTime();
+    const revisionRetentionDays: number = parseInt(process.env.HD_REVISION_RETENTION_DAYS || '0')
```
ErikMichelson, line 238: I know this is still a WIP pull request and probably subject to change. However in case you haven't seen it, the environment variable processing and validation is better done via the config module.

yamashush: > via the config module.

Thank you for teaching me.
This is very helpful as I didn't realize that.

mrdrogdrog:
Please rebase your branch to the latest version of develop to fix the CI.

yamashush:
> Please rebase your branch to the latest version of develop to fix the CI.

Thank you for letting me know.
After rebase, the test seems to work now!

docs/content/references/config/revision.md
```diff
@@ -0,0 +1,8 @@
+# Revisions
+
+| environment variable         | default | example | description                                                                                                                    |
+| ---------------------------- | ------- | ------- | ------------------------------------------------------------------------------------------------------------------------------ |
+| `HD_REVISION_RETENTION_DAYS` | 0       |         | The number of days a revision should be kept. If the config option is not set or set to 0, all revisions will be kept forever. |
+<!--
+| `HD_IMAGE_PROXY` | -       | `https://image-proxy.example.com`   | **ToDo:** Add description  |
```
ErikMichelson, line 7: Why is this env variable in the revision section?

ErikMichelson (commented):
In general this is a good starting point, thanks for that!

However when deleting old revisions, we need to somehow compensate the missing information, don't we? A revision holds the diff (called patch in the entity) to its preceding revision. This is used by the revision viewer to show the differences. When deleting old revisions, this results in an invalid diff. So the oldest not removed revision needs its diff be updated.

Or should we do it differently, @hedgedoc/backend @hedgedoc/frontend ?

docs/content/references/config/revision.md
yamashush: I did this because I thought it would be better to separate it by entity.
But maybe it should be in the notes section?

yamashush:
> However when deleting old revisions, we need to somehow compensate the missing information, don't we? A revision holds the diff (called patch in the entity) to its preceding revision. This is used by the revision viewer to show the differences. When deleting old revisions, this results in an invalid diff. So the oldest not removed revision needs its diff be updated.

I see, revisions have not only content but also diff (patch).
Thank you for making me realize that!!

I was referring to the code below that is called from the revision modal.
Does it seem like the diff(patch) needs to be updated here as well?

https://github.com/hedgedoc/hedgedoc/blob/9e315033f2fe0a434b663135cead2a92a4673baa/backend/src/revisions/revisions.service.ts#L63

docs/content/references/config/revision.md
yamashush: @ErikMichelson I moved it from revesion section to note section!

yamashush:
@ErikMichelson I have implemented updating patch (content diff), so could you please review it again?

backend/src/revisions/revisions.service.spec.ts
```diff
@@ -410,4 +435,192 @@ describe('RevisionsService', () => {
...
+        @@ -0,0 +1,6 @@
+        +---
+        +title: new title
+        +description: new description
+        +tags: [ "tag1" ]
+        +---
+        +new content
+        "
+      `);
+    });
```
ErikMichelson, line 533: Like already stated in the other PR, I think extracting the embedded snapshots into their own "normal snapshot" file would be good.

backend/src/revisions/revisions.service.ts
```diff
@@ -219,4 +226,73 @@ export class RevisionsService {
       await this.revisionRepository.save(revision);
     }
   }
+
+  // Delete all old revisions everyday on 0:00 AM
+  @Cron('0 0 * * *')
+  async handleCron(): Promise<void> {
```
ErikMichelson, line 232: I think we should rename these methods (`handleCron` and `handleTimeout`) to indicate that they're used in the context of revision cleanup. Maybe something like `handleRevisionCleanupTimeout` etc?

backend/src/revisions/revisions.service.ts
```diff
@@ -219,4 +226,73 @@ export class RevisionsService {
...
+    return await this.removeOldRevisions();
+  }
+
+  // Delete all old revisions 5 sec after startup
+  @Timeout(5000)
+  async handleTimeout(): Promise<void> {
+    return await this.removeOldRevisions();
+  }
+
+  async removeOldRevisions(): Promise<void> {
```
ErikMichelson, line 242: It would be good to have a block of ESDoc here

backend/src/revisions/revisions.service.ts
```diff
@@ -219,4 +226,73 @@ export class RevisionsService {
...
+          (revision) =>
+            new Date(revision.createdAt).getTime() <=
+            currentTime - revisionRetentionDays * 24 * 60 * 60 * 1000,
+        );
+      const remainedRevisions = revisions.filter((val) => !oldRevisions.includes(val))
+
+      if (!oldRevisions.length) {
+        continue;
+
+      } else if (oldRevisions.length == revisions.length - 1 ){
```
ErikMichelson, line 272: Please use explicit comparisons always (`===`).

backend/src/revisions/revisions.service.ts
```diff
@@ -219,4 +226,73 @@ export class RevisionsService {
...
+          createdAt: "ASC",
+        },
+      });
+
+      const oldRevisions = revisions
+        .slice(0, -1) // always keep the latest revision
+        .filter(
+          (revision) =>
+            new Date(revision.createdAt).getTime() <=
+            currentTime - revisionRetentionDays * 24 * 60 * 60 * 1000,
```
ErikMichelson, line 265: To improve performance, this second value could be calculated outside the loop, so before the `.filter` method call.

backend/src/revisions/revisions.service.ts
```diff
@@ -219,4 +226,73 @@ export class RevisionsService {
...
+      const remainedRevisions = revisions.filter((val) => !oldRevisions.includes(val))
+
+      if (!oldRevisions.length) {
+        continue;
+
+      } else if (oldRevisions.length == revisions.length - 1 ){
+        const beUpdatedRevision = revisions.slice(-1)[0]
+        beUpdatedRevision.patch = createPatch(
+          note.publicId,
+          '', // donnt exist older revision
```
ErikMichelson, line 276:
```suggestion
          '', // there is no older revision
```

ErikMichelson (commented):
Thanks for your contribution and sorry again that this took a while.

I've annotated some minor things.

yamashush:
@ErikMichelson Thanks for your review and sorry again that this took a while 🙏 
I fixed!!

ErikMichelson (approved):
Works and looks good to me. Thanks for your contribution! ❤️

yamashush:
Thank you so much for your review ❤️

Issue #5036, opened 2023-10-07
Only show delete Note button if user is owner


InnayTool:
### Which part of the project should be enhanced?

Sidebar and History page

### Describe the solution you'd like

Only show delete Note button if user is owner by deactivating or not showing buttons

### Describe alternatives you've considered

--

Issue #2898, opened 2020-10-01
Re-add export menu functionality


DerMolly:
We need to add functionality to the export menu:

- [x] ~~Dropbox~~
- [x] Gist
- [x] Gitlab snippet (see https://github.com/codimd/server/issues/508)
- [x] Markdown (see hedgedoc/react-client#627)
- [ ] HTML (see hedgedoc/hedgedoc#2889)

ErikMichelson:
With only HTML export remaining, which has its own issue, this can be closed.

PR #5356, opened 2024-01-17
Sidebar: Fixes & GitHub Gist / GitLab snippet export

ErikMichelson:
### Component/Part
editor sidebar

### Description
This PR adds various features and fixes to the sidebar:

* Show deletion button only when being the note owner
* Remove useless import from clipboard button
* Add GitHub Gist export
* Add GitLab snippet export

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
#5036, #2898


DerMolly (approved):
I'm debating if it would be a good idea to save the gist or gitlab token in the local storage and to get it from there if it's existing. But I don't like the idea to store such a token in local storage…
What do you think @ErikMichelson?

ErikMichelson:
@DerMolly I thought about that too but decided against localStorage for security reasons. Instead this could be implemented with the [CredentialsContainer API](https://developer.mozilla.org/en-US/docs/Web/API/CredentialsContainer), which uses encrypted storage underneath (in most cases the OS keystore). However for a first implementation this seemed overkill to me and could in my opinion wait until after 2.0. What do you think?

Issue #3608, opened 2023-03-24
Don't delete realtime-notes immediately after the last user left


mrdrogdrog:
### Which part of the project should be enhanced?

Backend realtime code

### Is your enhancement request related to a problem? Please describe.

If a user just refreshs the page and is the last user on a note then the backend creates a new revision and writes it to the database and immediately loads it again. This is a bit overhead.

### Describe the solution you'd like

Add a timer that waits a few seconds before the realtime note gets deleted.

### Describe alternatives you've considered

Accept the overhead.

### Additional context

Keep in mind that on application shutdown the revision must be created in any case!

ErikMichelson:
Has been implemented thanks to @yamashush

yamashush:
@ErikMichelson I appreciate it too. Thank you for fixing and merging the code.

ErikMichelson:
You're welcome and thanks for your contributions too!

PR #5385, opened 2024-01-21
chore(backend): Add timer berfore deleting realtime note

yamashush:
### Component/Part

Backend realtime note

### Description

This PR adds a timer before `RealtimeNote` gets deleted.

- adds 3 seconds lifetime to `RealtimeNote` without clients
- prevents from creating revision even if the last user reloads the screen

Please see issue #3608 for details.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)

- #3608

### DIFF
--- a/backend/src/realtime/realtime-note/realtime-note.ts
+++ b/backend/src/realtime/realtime-note/realtime-note.ts
@@ -5,7 +5,7 @@
  */
 import { Message, MessageType, RealtimeDoc } from '@hedgedoc/commons';
 import { Logger } from '@nestjs/common';
-import { EventEmitter2, EventMap } from 'eventemitter2';
+import { EventEmitter2, EventMap, Listener } from 'eventemitter2';
 
 import { Note } from '../../notes/note.entity';
 import { RealtimeConnection } from './realtime-connection';
@@ -19,6 +19,8 @@ export interface RealtimeNoteEventMap extends EventMap {
   yDocUpdate: (update: number[], origin: unknown) => void;
 }
 
+const LIFETIME_WITHOUT_CLIENTS = 10 * 1000; // 10 seconds
+
 /**
  * Represents a note currently being edited by a number of clients.
  */
@@ -26,7 +28,9 @@ export class RealtimeNote extends EventEmitter2<RealtimeNoteEventMap> {
   protected logger: Logger;
   private readonly doc: RealtimeDoc;
   private readonly clients = new Set<RealtimeConnection>();
+  private readonly clientAddedListener: Listener;
   private isClosing = false;
+  private destroyEventTimer: NodeJS.Timeout | null = null;
 
   constructor(
     private readonly note: Note,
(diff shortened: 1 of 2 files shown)


backend/src/realtime/realtime-note/realtime-note.ts
```diff
@@ -150,4 +157,11 @@ export class RealtimeNote extends EventEmitter2<RealtimeNoteEventMap> {
       connection.getTransporter().sendMessage(content);
     });
   }
+
+  /**
+   * Indicates if realtime note is ready to destory.
```
mrdrogdrog, line 162:
```suggestion
   * Indicates if a realtime note is ready to get destroyed.
```

backend/src/realtime/realtime-note/realtime-note.ts
```diff
@@ -150,4 +157,11 @@ export class RealtimeNote extends EventEmitter2<RealtimeNoteEventMap> {
...
+   */
+  private canDestory(): boolean {
```
mrdrogdrog, line 164:
```suggestion
  private canBeDestroyed(): boolean {
```

mrdrogdrog (changes requested):

backend/src/realtime/realtime-note/realtime-note.ts
```diff
@@ -68,8 +70,13 @@ export class RealtimeNote extends EventEmitter2<RealtimeNoteEventMap> {
...
     this.emit('clientRemoved', client);
-    if (!this.hasConnections() && !this.isClosing) {
-      this.destroy();
+
+    if (this.canBeDestroyed()) {
+      setTimeout(() => {
+        if (this.canBeDestroyed()) {
+          this.destroy();
+        }
+      }, LIFETIME_WITHOUT_CLIENTS);
```
ErikMichelson, line 79: I'm not completely sure, but this looks like it could break due to the timeout not being cancelled on reconnects.

Let's look at the following scenario: A note is loaded with one client connected. The client disconnects and the timer starts to run. Within less than 3 seconds, the client re-connects and disconnects again. Now there are two timers running. As there is no client left, the realtime-note gets destroyed by the first timer. Shortly after, the second timer calls this.destroy() on the already destroyed note.

yamashush: Thanks for your review! !

In the case of the scenario you presented, it certainly doesn't work as expected.
Thank you for making me realize that 😄

yamashush: @ErikMichelson
Added a listener that cancels the timer when a client connects.
I think it would work in your scenario as well.

ErikMichelson (commented):
Tested (rebased onto develop) and works fine.

<del>However, I'd argue to increase the timeout even a bit more (like maybe 10 seconds) as a full refresh in low internet speed conditions could exceed 3 seconds. What do you think, @hedgedoc/frontend ?</del>

Edit: Changed to 10 seconds for mentioned reasons.

mrdrogdrog (changes requested):
clientAddedListener needs to be destroyed when the realtime note gets destroyed

mrdrogdrog (approved):

yamashush:
@ErikMichelson Thank you for your fixing 🙏 🙏 🙏

PR #5535, opened 2024-03-24
Media redirection API endpoint & Note media browser

### LINKED ISSUE #2908: List uploads of current note

### LINKED ISSUE #3704: Allow "named" usage of uploaded media instead of URL

Uploads / Media handler

When uploading media in HedgeDoc 1 (and current state of 2.0), the file is stored in the configured backend and the URL to that file is inserted into the document.

This might look like this in the markdown content:
```markdown
![image alt](/uploads/ee037d72-5872-4dc4-9086-c0cc6ed8d23d.png)
```

When changing the storage backend for uploads, the URLs in each note have to be adjusted in order to continue working. This requires a lot of manual work or SQL queries which potentially could change unwanted places as well.

When uploading a file, the user can choose an identifier for that upload. This identifier is stored alongside the metadata and storage backend URL for the upload in the database. By default the identifier could be a short random string.

Instead of inserting the storage backend URL into the note's content, the identifier will be used. This could look like this:
```markdown
![image alt](#sunflower)
```

By using a `#` as a prefix, the address of the image won't conflict with possible existing images.

Because each upload is associated with the note where the file was uploaded, the identifiers don't need to be unique per instance, but only unique per note.

Upon loading a note in the frontend, the frontend fetches the `/api/notes/:note/media` endpoint to get an overview of the uploaded media objects and their identifiers.
A renderer middleware could intercept all `` tags with an URL starting with `#` and replace those with their storage backend URL.

Should the storage backend change in the future, no note has to be touched. Instead, it is only required to update the URLs in the media table of the database.

Instead of inserting the storage backend URL into the note, a special endpoint could be introduced that acts as a proxy for the image and lets the server fetch and deliver the image through this endpoint from the storage backend.

### Additional context

_No response_

ErikMichelson:
### Component/Part
* backend -> API -> media handler
* frontend -> editor -> sidebar -> media browser

### Description
This PR adds two things:
* an endpoint for being redirected to a media object's storage-backend URL
* a media browser for the sidebar in the editor

Previous versions of HedgeDoc suffered from the problem that changing the media backend required manipulation of the media links in all created notes. We discussed in
#3704 that it's favourable to have an endpoint that redirects to the image's original URL. When changing the media backend, the link stays the same but just the redirect changes.

The media browser allows to delete uploads attached to a note as well as re-inserting their image link into the note.

Bildschirmfoto vom 2024-03-23 19-04-35

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
closes #3704
closes #2908
depends on #5356


backend/src/api/public/media/media.controller.ts
```diff
@@ -101,6 +104,17 @@ export class MediaController {
...
 
+  @Get(':filename')
+  @OpenApi(404, 500)
+  async getMedia(
+    @Param('filename') filename: string,
+    @Res() response: Response,
+  ): Promise<void> {
+    const mediaUpload = await this.mediaService.findUploadByFilename(filename);
+    const targetUrl = mediaUpload.fileUrl;
+    response.redirect(targetUrl);
```
DerMolly, line 115: Do we need to implement  some kind of check if the user has access to the media? Otherwise a user with an API Token could theoretically try to find files with predictable names (e.g Screenshot_$date.png).

DerMolly (commented):
Did you try to access a note with an image uploaded by a regular user as an anyomous user? Does that work?

backend/src/api/public/media/media.controller.ts
DerMolly: As the filename get's randomized, this is not really a concern…

DerMolly (approved):

Issue #1024, opened 2021-03-11
Redirection loop with subdirectory:  `"urlPath": "hedgedoc",` and nginx config


zeigerpuppy:
using v 1.7.2, I am attempting to run HedgeDoc on a subdirectory.
However, loading fails with the error `ERR_TOO_MANY_REDIRECTS`.

I have tried a number of different proxy configs and Hedgedoc works fine on a normal domain if I don't use subdirectory.
Perhaps there is something broken in the redirect code?

I have included my hedgedoc config file and nginx configs below (sanitized with domain changed to example.org).

Does anyone else have hedgedoc serving from a subdir working?

## Configs

config.json:

```json
{
    "production": {
        "domain": "example.org",
        "urlPath": "hedgedoc",
        "protocolUseSSL": true,
        "urlAddPort": false,
        "allowFreeURL": true,
        "loglevel": "info",
        "sessionSecret": "xxxxxxx",
        "hsts": {
            "enable": true,
            "maxAgeSeconds": 31536000,
            "includeSubdomains": true,
            "preload": true
        },
        "csp": {
            "enable": true,
            "directives": {
            },
            "upgradeInsecureRequests": "auto",
            "addDefaults": true,
            "addDisqus": true,
            "addGoogleAnalytics": false
        },
        "cookiePolicy": "lax",
        "db": {
            "username": "hedgedoc",
            "password": "XXXXXXXXXX",
            "database": "hedgedoc",
            "host": "localhost",
            "port": "5432",
            "dialect": "postgres"
        },
        "useCDN": false,
        "allowGravatar": false,
        "email": true,
        "allowEmailRegister": true,
        "allowAnonymous": false,
        "allowAnonymousEdits": true,
        "gitlab": {
            "baseURL": "https://gitlab.example.org",
            "clientID": "XXXXXX",
            "clientSecret": "XXXXXXX",
            "version": "v4"
        },
        "minio": {
          "accessKey": "adm",
          "secretKey": "XXXXXX",
          "endPoint": "minio1.example.org",
          "secure": true,
          "port": 443
        },
        "s3bucket": "hedgedoc",
        "imageUploadType": "minio",
        "linkifyHeaderStyle": "gfm"
    }
}
```

nginx config

```nginx
map $http_upgrade $connection_upgrade {
        default upgrade;
        ''      close;
}

server {
    listen      80;
    server_name example.org;
    return 301 https://$server_name$request_uri;
}

server {
        listen 443 ssl;
[...]

PR #4938, opened 2023-09-25
Docs: Add a warning about an NGINX pitfall and its symptom in HedgeDoc

MattF-NSIDC:
### Component/Part

Docs

### Description

This PR adds a warning about an NGINX configuration pitfall that I fell in. Hopefully it will help others who make my mistake.

### Steps

- [x] Added / updated documentation
- [x] I read the [contribution documentation](https://github.com/hedgedoc/hedgedoc/blob/develop/CONTRIBUTING.md) and
  made sure that:
  - My commits are signed-off to accept the DCO.
  - This PR targets the correct branch: `master` for 1.x & docs, `develop` for 2.x

### Related Issue(s)

N/A

PR #5546, opened 2024-04-04
Update Nginx reverse proxy config example

sven-frotscher:
### Component/Part
Documentation

### Description
The warning in the Nginx section of the reverse proxy guide is misleading:
Omitting trailing slashes in Nginx' `proxy_pass` directives causes redirect loops (see related issue), while the issue described in the warning is a different one, caused by an incorrect socket.io URL.
This config has been successfully tested with HedgeDoc 1.9.8 and Nginx 1.22.1.

Fixes: 74b573e

### Steps

  made sure that:

### Related Issue(s)
#1024

### DIFF
--- a/docs/content/guides/reverse-proxy.md
+++ b/docs/content/guides/reverse-proxy.md
@@ -64,7 +64,7 @@ server {
         server_name hedgedoc.example.com;
 
         location / {
-                proxy_pass http://127.0.0.1:3000;
+                proxy_pass http://127.0.0.1:3000/;
                 proxy_set_header Host $host; 
                 proxy_set_header X-Real-IP $remote_addr; 
                 proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for; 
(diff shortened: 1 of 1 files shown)


ErikMichelson:
Thanks for your suggested change. This needs further testing and checking on our end as it is contrary to the current statement.
I could imagine that the behavior and the requirement for trailing slashes depends on whether HedgeDoc is run in a subdirectory path or not.

For context: The warning was introduced in #4938 and the reasoning for that was that the problem was reproducible by us when trailing slashes were present.

sven-frotscher:
In my testing, the endless loading animation occurred when proxying (e.g.) `/hedgedoc/socket.io/` to `127.0.0.1:3000/` and not to `127.0.0.1:3000/socket.io/`. Not sure if that helps explain it.

sven-frotscher:
@ErikMichelson Are there any news regarding such testing?
I noticed that my PR makes the docs almost identical to the config you provided in https://github.com/hedgedoc/hedgedoc/issues/1024#issuecomment-796770679 and which was confirmed working by multiple people in that thread. Yes, that confirmation only exists for subdirectories, but the docs are currently both misleading for that use case (→ redirect loop) and missing `/socket.io` in the second `proxy_pass` directive (→ endless loading animation).
Would you object to adding a section "Running HedgeDoc in a subdirectory" to that docs page in order to provide a separate and confirmed working Nginx config for the either use case?

PR #5574, opened 2024-05-09
fix(backend revision): Update patch column when removing revisions

yamashush:
### Component/Part

backend revision

### Description

This PR fixes the following feature's bug.

```javascript
  /**
   * @async
   * Purge revision history of a note.
   * @param {Note} note - the note to purge the history
   * @return {Revision[]} an array of purged revisions
   */
  async purgeRevisions(note: Note): Promise<Revision[]> {
```

If we delete all revisions except the latest one, we need to update the patch column of the latest one.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)

- https://github.com/hedgedoc/hedgedoc/pull/5349#issuecomment-1913119203

### DIFF
--- a/backend/src/revisions/revisions.service.spec.ts
+++ b/backend/src/revisions/revisions.service.spec.ts
@@ -34,6 +34,7 @@ import { EditService } from './edit.service';
 import { Revision } from './revision.entity';
 import { RevisionsService } from './revisions.service';
 
+
 describe('RevisionsService', () => {
   let service: RevisionsService;
   let revisionRepo: Repository<Revision>;
@@ -121,7 +122,7 @@ describe('RevisionsService', () => {
     let note: Note;
 
     beforeEach(() => {
-      note = Mock.of<Note>({});
+      note = Mock.of<Note>({ publicId: 'test-note', id: 1 });
       revisions = [];
 
       jest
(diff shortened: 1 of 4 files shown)


backend/src/revisions/revisions.service.spec.ts
```diff
@@ -141,26 +142,57 @@ describe('RevisionsService', () => {
...
+        +description: new description
+        +tags: [ "tag1" ]
+        +---
+        +new content
+        "
+      `);
     });
     it('has no effect on revision history when a single revision is present', async () => {
-      const revision1 = Mock.of<Revision>({ id: 1 });
+      const revision1 = Mock.of<Revision>({ id: 1});
```
ErikMichelson, line 195: This changes only the whitespace, right? We should just let prettier fix all of that (`yarn lint:fix`).

backend/src/revisions/revisions.service.spec.ts
```diff
@@ -141,26 +142,57 @@ describe('RevisionsService', () => {
...
+        +++ test-note
+        @@ -0,0 +1,6 @@
+        +---
+        +title: new title
+        +description: new description
+        +tags: [ "tag1" ]
+        +---
+        +new content
+        "
+      `);
```
ErikMichelson, line 192: Maybe we could also just use a "normal" snapshot instead of an embedded one. Seems easier in terms of maintenance to me (updating via `jest -u`) and is more aligned with the rest of the tests.

ErikMichelson (commented):
Sorry, that this took so long, being quite busy currently. I tested it and it works perfectly. Just some linting errors and another small thing.

Thanks for your contribution! :)

backend/src/revisions/revisions.service.spec.ts
yamashush: Sorry for my late reply 🙏 
I had fixed!!

backend/src/revisions/revisions.service.spec.ts
yamashush: Thank you for your review!!
I had changed snapshot type from inline to file like frontend.

yamashush:
@ErikMichelson Sorry for being so late 🙏 
I've finished making the changes, so please review again 😄

ErikMichelson (approved):

Issue #5575, opened 2024-05-09
Adapt loading screen to set theme (light/dark)


ErikMichelson:
### Which part of the project should be enhanced?

Loading screen

### Describe the solution you'd like

The loading screen should respect the stored theme preference.

### Describe alternatives you've considered

-

boehs:
I'll work on this

PR #5577, opened 2024-05-12
feat: less flashly loader

boehs:
### Component/Part

frontend
### Description
This PR fixes/adds/improves/...
This PR makes the loader not flash while the theme global state is still initiating. The method for this is to create a new method that syncronously loads the theme from the same logic, as it is safe to assume that the theme will not change while the loader is displayed.
### Steps

- [ ] Added implementation
- [ ] Added / updated tests
- [ ] Added changelog snippet
  made sure that:

### Related Issue(s)

#5575


boehs:
Needs CI to be force activated I believe

boehs:
There's still a problem where there is a brief half second where there is a FOUC. This is a lot better than before, but I wouldn't describe it as ideal

mrdrogdrog:
> Needs CI to be force activated I believe

I'm gonna unlock it when you're done and mark the pr as ready to review.

TheRedXD:
It has effectively been a year, when will this be pushed?

ErikMichelson:
We'll probably take over this PR as the core developer team, however it is currently a low priority one for us, since it is only a cosmetic thing.

Issue #2344, opened 2022-05-13
Anchor links don't include the note URL


mahoromax:
### Description

The links produced by the icons near headings are missing the document ID

### Steps to reproduce

1. Go to any HedgeDoc document (for example [this one](https://demo.hedgedoc.org/dZM2QG1dTD6nrU0vgBj5Sg) )  
2. Choose a heading and move the mouse over to the left of the heading so a link/chain icon appears
![image](https://user-images.githubusercontent.com/22126466/168264672-8c76d64c-6476-4d73-9514-645baca7e94d.png)
3. Click the link and copy the url form the address bar or rightclick and copy the link
`https://demo.hedgedoc.org/#Target-heading`
4. Open new window and go to the copied URL (for example [this](https://demo.hedgedoc.org/#Target-heading) )

### Expected behaviour

The generated link should contain the document-id:
`https://demo.hedgedoc.org/dZM2QG1dTD6nrU0vgBj5Sg#Target-heading`

### Your Setup

- Edge Browser 
- Firefox

davidmehren:
I think we broke this in https://github.com/hedgedoc/hedgedoc/pull/1479
@mrdrogdrog

mahoromax:
I'm really not proficient in any language involved here, but if I got this correctly, the lines 
```
    currentLocation.hash = linkTag.hash
    linkTag.href = currentLocation.toString()
```
overwrite what was previously the URL including the document information.

The goal was to provide the location to another functionality (slides), so would it not work to save the _currentLocation_ to another value of _linkTag_ (like _linkTag.loc_) and then find where _linkTag.href_ is used and replace the occurrences with _linkTag.loc_ ?
Or if easier create a new entry for the occasions a full URL is needed, and replace those?

davidmehren:
Fixed by https://github.com/hedgedoc/hedgedoc/pull/5769

PR #5769, opened 2024-08-31
fix: anchor links

davidmehren:
Fixes https://github.com/hedgedoc/hedgedoc/issues/2344

### DIFF
--- a/public/js/extra.js
+++ b/public/js/extra.js
@@ -847,7 +847,7 @@ const anchorForId = id => {
   const anchor = document.createElement('a')
   anchor.ariaHidden = 'true'
   anchor.className = 'anchor hidden-xs'
-  anchor.href = `#${id}`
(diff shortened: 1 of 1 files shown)


public/js/extra.js
```diff
@@ -847,7 +847,7 @@ const anchorForId = id => {
   const anchor = document.createElement('a')
   anchor.ariaHidden = 'true'
   anchor.className = 'anchor hidden-xs'
-  anchor.href = `#${id}`
+  anchor.href = `${document.location.origin}${document.location.pathname}#${id}`
```
mrdrogdrog, line 850: I would prefer
```
anchor.href = new URL(`#${id}`, document.location).toString()
```
but still legit.

mrdrogdrog (approved):

public/js/extra.js
davidmehren: That version looks less brittle 👍🏻

Issue #5792, opened 2024-09-02
Add auth identity source service tests


ErikMichelson:
### Which part of the project should be enhanced?

backend -> auth -> identity services

### Describe the solution you'd like

The `identity.service` as well as the services for the auth sources `local`, `ldap`, `oidc` should receive unit tests. This was not possible in #4827 due to non-possible mocking of the `DataSource` in Nest with TypeORM and transactions and without falling back to deprecated methods like `getConnection` (which actually also don't work for us).

### Describe alternatives you've considered

Replace the Database connector

PR #5831, opened 2024-09-27
Feat: Add OIDC state parameter

IvanLi-CN:
### Component/Part

Auth/OIDC

### Description

While integrating HedgeDoc 2 with Authelia, I encountered an invalid_state error from Authelia. This is because HedgeDoc 2 does not use state during OIDC authorization. Honestly, I'm not very familiar with OIDC, but from my understanding, implementing both PKCE and [state](https://arc.net/l/quote/ogpluvht) is a good security practice. Therefore, I created this PR.

### Steps

 - [x] Added implementation
 - Added / updated tests (Perhaps implemented by #5792)
made sure that:
My commits are signed-off to accept the DCO.
This PR targets the correct branch: master for 1.x & docs, develop for 2.x

### Related Issue(s)

None

### DIFF
--- a/backend/src/identity/oidc/oidc.service.ts
+++ b/backend/src/identity/oidc/oidc.service.ts
@@ -120,13 +120,27 @@ export class OidcService {
   }
 
   /**
+   * Generates a random state for the OIDC login.
+   *
+   * @returns {string} The generated state.
+   */
+  generateState(): string {
+    return generators.state();
+  }
+
+  /**
    * Generates the authorization URL for the given OIDC identifier and code.
    *
    * @param {string} oidcIdentifier The identifier of the OIDC configuration
    * @param {string} code The code verifier generated for the login
+   * @param {string} state The state generated for the login
    * @returns {string} The generated authorization URL
    */
-  getAuthorizationUrl(oidcIdentifier: string, code: string): string {
+  getAuthorizationUrl(
+    oidcIdentifier: string,
+    code: string,
+    state: string,
+  ): string {
     const clientConfig = this.clientConfigs.get(oidcIdentifier);
     if (!clientConfig) {
       throw new NotFoundException(
@@ -139,6 +153,7 @@ export class OidcService {
       /* eslint-disable @typescript-eslint/naming-convention */
       code_challenge: generators.codeChallenge(code),
       code_challenge_method: 'S256',
+      state,
       /* eslint-enable @typescript-eslint/naming-convention */
     });
   }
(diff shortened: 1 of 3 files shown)


ErikMichelson:
Thanks for your contribution!
Code looks good at the first view. Will do a more detailed review and testing soon.

backend/src/identity/oidc/oidc.service.ts
```diff
@@ -119,14 +120,23 @@ export class OidcService {
...
+
   /**
    * Generates the authorization URL for the given OIDC identifier and code.
    *
    * @param {string} oidcIdentifier The identifier of the OIDC configuration
    * @param {string} code The code verifier generated for the login
    * @returns {string} The generated authorization URL
    */
-  getAuthorizationUrl(oidcIdentifier: string, code: string): string {
+  getAuthorizationUrl(oidcIdentifier: string, code: string, state: string): string {
```
ErikMichelson, line 139: The `state` parameter should be described in the docs block too.

backend/src/identity/oidc/oidc.service.ts
```diff
@@ -166,15 +177,18 @@ export class OidcService {
     const oidcConfig = clientConfig.config;
     const params = client.callbackParams(request);
     const code = request.session.oidcLoginCode;
+    const state = request.session.oidcLoginState;
```
ErikMichelson, line 180: While it isn't explicit necessary since the state is regenerated for every OIDC login, it would still be better to clean it up after a finished login like with the PKCE `code`.

The clean-up takes place in this same method at the end by simply setting the session value to undefined.

ErikMichelson (changes requested):
All in all this looks quite good. I tested it with different auth providers and it works well! So, thank you for that contribution.

I've added some minor remarks for little improvements.

While on it, please rebase your branch on top of the current develop branch. Please fix the linting/formatting too by running `yarn lint --fix` in the backend directory.
When done with all that, you can simply amend your existing commit and force-push that to your branch.

backend/src/identity/oidc/oidc.service.ts
```diff
@@ -119,14 +120,23 @@ export class OidcService {
     return generators.codeVerifier();
   }
 
+  /**
+   * Generates a random state for the OIDC login.
+   *
+   * @returns {string} The generated state.
+   */
+  generateState(): string {
+    return randomBytes(32).toString('base64url');
```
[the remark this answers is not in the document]
IvanLi-CN: You're right, I can't believe I didn't notice that there was this function to call before.

backend/src/identity/oidc/oidc.service.ts
IvanLi-CN: Thank you very much for your patience, I added the code to clean up the state.

https://github.com/hedgedoc/hedgedoc/pull/5831/files#diff-da64bf255fae053e7438a387a3b1a178a495330d35b73fa1d8272f467aba180dR234-R235

IvanLi-CN:
Thank you for your careful review and detailed suggestions. I've followed your advice and updated my commit accordingly. I believe it's now ready to go. I really appreciate your thorough examination and the helpful recommendations you provided.

ErikMichelson (approved):

PR #5887, opened 2024-12-11
fix(csp): allow 'data:' images in pads

domrim:
### Component/Part
CSP

### Description
For embedded mermaid images `img-src: data:` has to be included in the CSP Header.

### Steps

- [x] Added implementation
- [x] Added changelog snippet
  made sure that:

### DIFF
--- a/lib/csp.js
+++ b/lib/csp.js
@@ -11,7 +11,7 @@ const defaultDirectives = {
   fontSrc: ['\'self\''],
   manifestSrc: ['\'self\''],
   frameSrc: ['\'self\'', 'https://player.vimeo.com', 'https://www.slideshare.net/slideshow/embed_code/key/', 'https://www.youtube.com'],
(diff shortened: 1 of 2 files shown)


ErikMichelson (approved):
Thanks for your contribution! 🚀 

Looks fine in my opinion and since we allow every image host anyway, this shouldn't have further security implications.

On a sidenote: I can be mistaken, but this seems to be a rather recent-ish thing that `*` does not include the data protocol anymore? It is not explicitly stated in the MDN docs at least, and I can't remember this problem occurring way back in the past… Anyway at least Chrome DevTools confirm that your PR is right:

> "img-src *". Note that '*' matches only URLs with network schemes ('http', 'https', 'ws', 'wss'), or URLs whose scheme matches `self`'s scheme. The scheme 'data:' must be added explicitly.

DerMolly (approved):

PR #6439, opened 2026-04-07
chore(env): match documentation + remove "broken" path to cert

NicolaiSoeborg:
### Component/Part

Default `.env.example` changes to make it easier to quickly get started developing

### Description

The current `.env.example` has a bunch of invalid values, so if someone quickly `cp .env.example .env` it will take a few takes before hedgedoc spawns (and it will crash with some not very obvious errors if .env is bad).

Ofc people shouldn't try to run with `.env.example` directly, but e.g. `HD_AUTH_LOCAL_MINIMAL_PASSWORD_STRENGTH` is not even allowed to be 8, so change it to the documentation default (2)

### Steps

  made sure that:

### Related Issue(s)

Tiny change so didn't make an issue


ErikMichelson (approved):

ErikMichelson:
Makes sense, thanks for fixing

PR #6440, opened 2026-04-07
Image: use relative url

NicolaiSoeborg:
### Component/Part
Frontend / image url

### Description
Embedding the full URL into the notes is a bit annoying if you want to change hostname (or e.g. hosting from a local IP). Using the relative URL should work in all cases afaik.

I would also like to remove that `additionalUrlText` argument, which is a bit confusing (not sure it works as intended?). It is only used in file `use-on-image-upload-from-renderer.ts` - and not sure it even is set to anything

### Steps

- [x] Added changelog snippet
  made sure that:

### DIFF
--- a/frontend/CHANGELOG.md
+++ b/frontend/CHANGELOG.md
@@ -100,6 +100,7 @@ SPDX-License-Identifier: CC-BY-SA-4.0
 - If only one external login provider is configured, the sign-in button will directly link to it.
 - Links in Gist-Frames work only if explicitly opened in new tabs.
 - Changed default editor-placeholder to include the full url to the /features-page as new users might not find it otherwise.
+- Changed image path to be relative.
 
 ---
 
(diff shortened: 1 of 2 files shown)


ErikMichelson:
Thanks for your contribution!

This looks like a good change to me. Originally, we intended to use absolute URLs so that the markdown file itself is portable and always pointing to the valid image on the server. Since we started to implement session and auth features for uploads as well (the latter one is coming soon), this approach does not make much sense anymore.
The relative URL looks a bit weird since it is relative to the renderer endpoint (`/render`) and not to the actual note editor endpoint (`/n/example`). This might confuse some enduser, but I still like this more than it being a URL relative to the root `/media/...` instead of `media/...`.

The `additionalUrlText` might not be the best name but is required for one specific edge-case of the upload placeholder. HedgeDoc's markdown flavour supports adding hover-titles to images by appending the text after the URL like this:

```markdown
![alt text](media/123456 "Hover title text")
```

You can add a placeholder with some hover title text like this in your note.

```markdown
![](https:// "Hover title")
```

This will render the placeholder component in the renderer, which can be replaced by dragging an image into it. This replacement should keep the hover title from the placeholder. Therefore, we need this additional argument.

ErikMichelson (approved):

DerMolly (approved):

Issue #6445, opened 2026-04-14
The Authelia configuration guide is out of date (fix included)


mkende:
### Description

The [Authelia configuration guide](https://docs.hedgedoc.org/guides/auth/authelia/) seems out of date with the current version of Authelia.

### Steps to reproduce

- Setup HedgeDoc
- Setup OIDC auth by following the [Authelia guide](https://docs.hedgedoc.org/guides/auth/authelia/)

### Expected behaviour

OIDC works

### Logs

```shell
First: received a 401 on https://<authelia-url/api/oicd/authorize after clicking on the login button.

Then, after fixing the url issue, the login still failed with the following in my Authelia logs:


time="2026-04-14T22:18:26+02:00" level=error msg="Access Request failed with error: Client authentication failed (e.g., unknown client, no client authentication included, or unsupported authentication method). The request was determined to be using 'token_endpoint_auth_method' method 'client_secret_post', however the OAuth 2.0 client registration does not allow this method. The registered client with id '<redacted>' is configured to only support 'token_endpoint_auth_method' method 'client_secret_basic'. Either the Authorization Server client registration will need to have the 'token_endpoint_auth_method' updated to 'client_secret_post' or the Relying Party will need to be configured to use 'client_secret_basic'." method=POST path=/api/oidc/token remote_ip=172.20.0.7
```

### Config

```shell

```

### Your Setup

n/a

### Additional context

One issue is that you need to set:

```
CMD_OAUTH2_AUTHORIZATION_URL=https://<authelia-url>/api/oidc/authorization
```

Instead of `/authorize` that is in the current doc.

The second issue is in the Authelia config itself which needs the following two parameters that are ignored (inside the client spec):

```
        require_pkce: false
        token_endpoint_auth_method: 'client_secret_post'
```

And the `audience` field can be removed from the example as it now default to nothing anyway.

P.S. I would have sent a PR directly, but I couldn’t find the source of the website.

ErikMichelson:
Hi @mkende, thanks for this report and the necessary fixes.

The repository contains both the code for the upcoming 2.0 (develop branch) and for the stable 1.x versions (master branch). The docs for the 1.x version are therefore in the [docs directory of the master branch](https://github.com/hedgedoc/hedgedoc/tree/master/docs).

If you still like to do so, we're happy if you create a PR.

ErikMichelson:
Merged, thanks!

PR #6465, opened 2026-04-15
docs(auth): update authelia configuration guide

mkende:
### Component/Part

Documentation only

### Description

This PR updates the guide to match the current version of Authelia:

- Use the current field names and values (`client_id` instead of `id`).
- The correct redirect URL is `/authorization` instead of `/authorize` (in HedgeDoc’s config).
- Add the missing `token_endpoint_auth_method` needed by HedgeDoc.
- Removes some fields which were just set to their default value.
- Also recommend using a random client id as per Authelia’s best-practice.

### Related Issue(s)

#6445

### DIFF
--- a/docs/content/guides/auth/authelia.md
+++ b/docs/content/guides/auth/authelia.md
@@ -22,12 +22,11 @@ identity_providers:
     refresh_token_lifespan: 90m
     enable_client_debug_messages: false
     clients:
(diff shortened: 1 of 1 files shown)


ErikMichelson (approved):
Thanks for your contribution and ensuring the docs are up-to-date!

PR #6504, opened 2026-05-12
Frontmatter validator with zod (custom fields & slideOptions)

### LINKED ISSUE #5946: 2.0a3: Using slideOptions yields an error: slideOptions.theme is not allowed
### Description

When trying to set slideOptions, for example to set a theme, it doesn't work and it says it is not allowed:

Image

Maybe I am missing something obvious, this is a rather default installation via docker-compose and nginx in front instead of caddy.

### Steps to reproduce

Using this document to reproduce it:
```yaml
---
title: testing
type: slide
slideOptions:
  theme: solarized
---

Hello World

---

A second slide

---
```

There are other examples in the docs:
- https://docs.hedgedoc.org/references/slide-options/#basic-yaml-header
- https://docs.hedgedoc.org/references/yaml-metadata/#slideoptions

### Expected behaviour

slideOptions should be allowed and it should be possible to set a theme.

There are past issues and PRs about fixing slideOptions and theme support but this issue appears to be different.

### Logs

```shell
Unless mistaken, there are no backend/frontend logs generated as a result of this issue.
There doesn't seem to be errors generated in the javascript console or something like 403's in the network tab.
```

### Config

```shell

```

### Your Setup

- Host OS: Debian 12
- NodeJS version: v20.17.0
- HedgeDoc version: 2.0 alpha 3

### Additional context

_No response_

ErikMichelson:
### Component/Part
- commons -> frontmatter validator
- frontend -> frontmatter-linter

### Description
The frontmatter validator was still the one left place that used Joi instead of the now widely used zod in HedgeDoc. Since zod can do validation, coercion and providing types based on the schema, the code could be drastically reduced compared to the old frontmatter validator.

At the same time, the validator is now less strict. Custom fields are still allowed for people that want to add their own frontmatter tags which are unrelated to HedgeDoc. Furthermore, we now allow the complete set of RevealOptions for the slideOptions key instead of only a few handpicked ones.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
Fixes #5946


commons/src/note-frontmatter/note-frontmatter.ts
```diff
@@ -0,0 +1,57 @@
...
+      .describe('List of tags for filtering on the explore page')
+      .default([]),
+    type: z
+      .nativeEnum(NoteType)
+      .default(defaultNoteFrontmatter.type)
+      .describe('Type of the renderer to use'),
+    robots: z.string().default(defaultNoteFrontmatter.robots).describe('Robots meta tag'),
+    lang: z.enum(ISO6391).default(defaultNoteFrontmatter.lang).describe('Language of the note'),
+    dir: z
+      .enum(['ltr', 'rtl'])
```
DerMolly: We should use a TextDirecttion enum here

DerMolly (changes requested):

DerMolly (approved):

ErikMichelson (commented):
Thanks for fixing!

PR #6510, opened 2026-05-13
Restructure revision cleanup to avoid SQL errors and data loss

### LINKED ISSUE #6487: Periodic revision cleanup task (if configured) throws SQLite errors

ErikMichelson:
### Component/Part
backend -> periodic revision cleanup (disabled by default)

### Description
The generated SQL from Knex contained ambigous column names, resulting in failure in the real-world since these queries were rejected by the SQLite library. This PR fixes the problem by properly making all column names explicit.

During this change we detected, that the current algorithm may produce data loss (for notes with only one old revision, this can be deleted, thus removing the content of the note).
This PR restructures the algorithm to avoid this behaviour.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
Fixes #6487

### DIFF
--- a/backend/src/revisions/revisions.service.spec.ts
+++ b/backend/src/revisions/revisions.service.spec.ts
@@ -563,10 +563,10 @@ describe('RevisionsService', () => {
       mockSelect(
         tracker,
         [
-          FieldNameRevision.uuid,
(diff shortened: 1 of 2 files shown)


DerMolly (approved):
I very much like your structure of the algorithm and documentation of it.

PR #6558, opened 2026-07-05
Add permission checks for accessing media uploads

ErikMichelson:
### Component/Part
media

### Description
This PR improves the implementation of media upload permissions by enforcing fine-grained permission checks. Previously, media uploads were served regardless of the requesting user's permission. This update allows linking one media upload to multiple notes and ensures permissions are properly enforced during access.

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
none

### DIFF
--- a/backend/src/media/backends/filesystem-backend.spec.ts
+++ b/backend/src/media/backends/filesystem-backend.spec.ts
@@ -1,0 +1,183 @@
+/*
+ * SPDX-FileCopyrightText: 2026 The HedgeDoc developers (see AUTHORS file)
+ *
+ * SPDX-License-Identifier: AGPL-3.0-only
+ */
+import { describe, it, expect, beforeAll, beforeEach, jest } from '@jest/globals';
+import { MediaBackendType } from '@hedgedoc/commons';
+import { promises as fs } from 'fs';
+import { Mock } from 'ts-mockery';
+
+import { MediaConfig } from '../../config/media.config';
+import { MediaBackendError } from '../../errors/errors';
+import { ConsoleLoggerService } from '../../logger/console-logger.service';
+import { FilesystemBackend } from './filesystem-backend';
+
+jest.mock('fs', () => ({
+  promises: {
+    access: jest.fn(),
+    mkdir: jest.fn(),
+    readFile: jest.fn(),
+    unlink: jest.fn(),
+    writeFile: jest.fn(),
+  },
+}));
+
+describe('filesystem backend', () => {
+  const mockedUploadPath = '/tmp/test_uploads';
+  const mockedUuid = 'cbe87987-8e70-4092-a879-878e70b09245';
+  const mockedBuffer = Buffer.from('test');
+
+  const mockedLoggerService = Mock.of<ConsoleLoggerService>({
+    setContext: jest.fn(),
+    error: jest.fn(),
+    debug: jest.fn(),
+  });
+
+  function mockMediaConfig(): MediaConfig {
+    return Mock.of<MediaConfig>({
+      backend: {
+        type: MediaBackendType.FILESYSTEM,
+        filesystem: {
+          uploadPath: mockedUploadPath,
+        },
+      },
+    });
+  }
+
+  let sut: FilesystemBackend;
+
+  beforeAll(() => {
+    sut = new FilesystemBackend(mockedLoggerService, mockMediaConfig());
+  });
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    jest.spyOn(fs, 'access').mockResolvedValue(undefined);
+  });
+
+  describe('saveFile', () => {
+    it('writes the buffer to the expected path and returns the backend data', async () => {
+      const writeFileSpy = jest.spyOn(fs, 'writeFile').mockResolvedValue(undefined);
+
+      const result = await sut.saveFile(mockedUuid, mockedBuffer, {
+        mime: 'image/png',
+        ext: 'png',
+      });
+
+      expect(writeFileSpy).toHaveBeenCalledWith(
+        `${mockedUploadPath}/${mockedUuid}.png`,
+        mockedBuffer,
+        null,
+      );
+      expect(result).toBe(JSON.stringify({ ext: 'png', mime: 'image/png' }));
+    });
+
+    it('throws a MediaBackendError if the extension is not alphanumeric', async () => {
+      const writeFileSpy = jest.spyOn(fs, 'writeFile');
+
+      await expect(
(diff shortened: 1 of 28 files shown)


backend/src/media/backends/filesystem-backend.ts
```diff
@@ -63,15 +64,33 @@
...
       throw new MediaBackendError('No backend data provided');
     }
-    const { ext } = JSON.parse(backendData) as { ext: string };
+    const { ext, mime } = JSON.parse(backendData) as { ext: string, mime: string };
     if (!ext) {
       throw new MediaBackendError('No file extension in backend data');
     }
-    return Promise.resolve(`/uploads/${uuid}.${ext}`);
+    const filePath = this.getFilePath(uuid, ext);
+    const buffer = await fs.readFile(filePath);
```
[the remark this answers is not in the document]
ErikMichelson: This is partly wrong, since `getFilePath` is called by methods which have checked inputs already coming from the `MediaService`. As that service checks for entries in the database first, this fails because invalid entries would not be found in the database.

backend/src/media/backends/filesystem-backend.spec.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  beforeEach(() => {
+    jest.clearAllMocks();
+    jest.spyOn(fs, 'access').mockResolvedValue(undefined);
+  });
+
+  describe('saveFile', () => {
+    it('writes the buffer to the expected path and returns the backend data', async () => {
+      const writeFileSpy = jest.spyOn(fs, 'writeFile').mockResolvedValue(undefined);
+      const sut = new FilesystemBackend(mockedLoggerService, mockMediaConfig());
```
DerMolly: This could be extracted to be common…

ErikMichelson: Extracted to one-time initialization in `beforeAll` and removed from all places where the FilesystemBackend was initialized.

backend/src/media/media.service.ts
```diff
@@ -103,9 +105,9 @@ export class MediaService {
     fileName: string,
     fileBuffer: Buffer,
     userId: User[FieldNameUser.id],
-    noteId: Note[FieldNameNote.id],
+    noteId?: Note[FieldNameNote.id],
```
DerMolly: Why is this `noteId` optional? It seems to me that this should not be a case we need to consider?

backend/src/media/media.service.ts
```diff
@@ -209,38 +220,153 @@ export class MediaService {
...
+
+    for (const noteId of linkedNoteIds) {
+      const linkedNotePermission = await this.permissionService.determinePermission(userId, noteId);
+      if (linkedNotePermission >= PermissionLevel.READ) {
+        return true;
+      }
+    }
+
+    return false;
+  }
```
DerMolly: This method is only used in the method `ensureUserCanAccessUpload`. This should be simplified

backend/src/media/media.service.ts
```diff
@@ -209,38 +220,153 @@ export class MediaService {
...
+
+  /**
+   * Retrieves an array of note IDs linked to the specified media upload UUID.
+   */
+  private async getLinkedNoteIds(uuid: string): Promise<number[]> {
+    const results = await this.knex(TableMediaUploadNote)
+      .select(FieldNameMediaUploadNote.noteId)
+      .where(FieldNameMediaUploadNote.mediaUploadUuid, uuid);
+    return results.map((result) => result[FieldNameMediaUploadNote.noteId]);
   }
```
DerMolly, line 390: This method never gets called to actually get the noteIds. It's just used to get the number of noteIds. Should this method be used for it's actual purpose down the line or should we refactor this to only return the count?

backend/src/permissions/permission.service.ts
```diff
@@ -68,34 +70,82 @@ export class PermissionService {
...
+      throw new NotInDBError(
+        'There is no note with the given id',
+        this.logger.getContext(),
+        'canReadNote',
+      );
+    }
+
+    if (userId === null) {
+      return false;
+    }
```
DerMolly: This could be moved at the beginning of the method

backend/src/permissions/permission.service.ts
```diff
@@ -68,34 +70,82 @@ export class PermissionService {
...
+  /**
+   * Checks whether a given user may read a given note.
+   *
+   * @param userId The id of the user who wants to read the note, or null for anonymous access
+   * @param noteId The id of the note
+   * @param transaction Optional transaction to use
+   * @returns true if the note can be read, false otherwise
+   * @throws NotInDBError if the note does not exist
+   */
+  public async canReadNote(
```
DerMolly: This method is currently not used. It's probably is missing from some other function?

backend/test/private-api/private-api.media.e2e-spec.ts
DerMolly, line 1: This file is missing a test for `GET ${PRIVATE_API_PREFIX}/media/:uuid`

DerMolly (changes requested):

backend/src/media/media.service.ts
ErikMichelson: You're right. Fixed.

backend/src/media/media.service.ts
ErikMichelson: Removed the extra ensure method and instead moved the check and PermissionError throwing to the controller level.

backend/src/media/media.service.ts
ErikMichelson: No, it is used actively not only with the count! In `canUserAccessUpload` we use it to get the noteIds which are then checked with `determinePermission` to check whether the user may access the upload.

backend/src/permissions/permission.service.ts
ErikMichelson: Leftover from wip state when creating the PR. Removed.

backend/test/private-api/private-api.media.e2e-spec.ts
ErikMichelson: Added

backend/src/permissions/permission.service.ts
ErikMichelson: No, it can't. The top checks whether a media upload exists. If it does not, it throws a `NotInDBError`. The permission checks are done after that.

If we drag this permission check to the top, non-logged-in guests can't receive a 404 for non-existing media uploads anymore, because they would instead get a 403.

backend/src/media/media.service.ts
DerMolly: My bad. You're totally right. I read only the first part of `canUserAccessUpload`

DerMolly (approved):

PR #6563, opened 2026-07-09
fix(revision): correct unpatching coordinates and whitelist details tag

### LINKED ISSUE #6513: Using `<summary>`/`<details>` tags corrupted entire revision history
### Description

I added `<summary>`/`<details>` tags to my document, changed my mind, and tried to revert to a previous revision. Unfortunately, it appears the use of these tags has corrupted my entire revision history. See, for instance, the attached screenshot, where the `<summary>`/`<details>` tags appear in a revision that did not use to have them and where the text has clearly become garbled.

Image

### Steps to reproduce

If there's any interest in addressing this issue, I can try to reproduce the behaviour and collect further information.

### Expected behaviour

No document edit should ever corrupt the revision history.

### Logs

```shell

```

### Config

```shell

```

### Your Setup

- Host OS: [e.g. Ubuntu 20.04]
- NodeJS version: [run `node --version`]
- HedgeDoc version:  1.9.9

### Additional context

_No response_

Shivam-Shukla0:
### Component/Part
Backend Worker / Frontend Render System

### Description
This PR fixes the revision history corruption bug that occurs during backward patch reconstruction (unpatching). It also adds the HTML `<details>` tag to the custom XSS whitelist to prevent it from being rendered as escaped HTML in the frontend.

Specifically:
- In `lib/workers/dmpWorker.js`, during unpatching, the patch coordinates (`start1`, `start2`) and lengths (`length1`, `length2`) are now swapped alongside the `DIFF_INSERT` / `DIFF_DELETE` operations. This aligns coordinates mathematically in Google's Diff-Match-Patch library when moving backward.
- In `public/js/render.js`, `<details>` is added to the custom XSS filter whitelist so it displays properly along with `<summary>`.
- In `test/revision.js`, unit tests are added to verify the correctness of the `getRevision` unpatching coordinates/offsets.

### Steps

- [x] Added implementation
- [x] Added / updated tests
- [x] Added changelog snippet

### Related Issue(s)
Fixes #6513

### DIFF
--- a/lib/workers/dmpWorker.js
+++ b/lib/workers/dmpWorker.js
@@ -97,8 +97,17 @@ function getRevision (revisions, count) {
     }
     // swap DIFF_INSERT and DIFF_DELETE to achieve unpatching
     for (let i = 0, l = applyPatches.length; i < l; i++) {
-      for (let j = 0, m = applyPatches[i].diffs.length; j < m; j++) {
-        const diff = applyPatches[i].diffs[j]
+      const patch = applyPatches[i]
+      const tempStart = patch.start1
+      patch.start1 = patch.start2
+      patch.start2 = tempStart
+
+      const tempLength = patch.length1
+      patch.length1 = patch.length2
+      patch.length2 = tempLength
+
+      for (let j = 0, m = patch.diffs.length; j < m; j++) {
+        const diff = patch.diffs[j]
         if (diff[0] === DiffMatchPatch.DIFF_INSERT) {
           diff[0] = DiffMatchPatch.DIFF_DELETE
         } else if (diff[0] === DiffMatchPatch.DIFF_DELETE) {
@@ -145,3 +154,8 @@ process.on('uncaughtException', function (err) {
   logger.error('Process will exit now.')
   process.exit(1)
 })
+
+module.exports = {
+  createPatch,
+  getRevision
+}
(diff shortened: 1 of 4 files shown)


ErikMichelson:
Thanks for your contribution! We'll review it very soon.

ErikMichelson:
Just for letting you know about the state of this PR.

On a first glance this looks very good, thanks! 
We have to dive into the revision code once more to finalize the review.

We're preparing another 1.x release in the near future and this will probably be part of it.

Shivam-Shukla0:
Thanks for the update, @ErikMichelson! 

Happy to hear it looks good so far. Let me know if 
anything needs adjusting after the deeper review of 
the revision code — I'm available to make changes.

ErikMichelson:
We just did an in-depth review of your PR and the related issue again. Mainly, because we were unable to reproduce the problem with the current codebase (master branch) and we assume that the original issue might be related to an already fixed bug in the operational transformations logic.

Adding the `details` tag to the overrides of the XSS-filter whitelist is completely unnecessary since it is already included by default.

Your main change inverts the metadata of the diffs of the patches by swapping start and length of each two diffs when applying them in reverse order.

After extensively checking, testing and verifying each code path and result again, we came to the conclusion that your fix might be potentially correct however due to HedgeDoc's usage of the diffing library we don't reach the case where your fix becomes relevant.

We therefore will close this issue.

Shivam-Shukla0:
Thanks for taking the time to do such a thorough review — 
I appreciate you walking through the code paths in detail.

Good to know the case isn't reachable with the current 
diffing library usage. Learned something useful here about 
verifying reproducibility against the current codebase 
before diving into a fix.

PR #6572, opened 2026-07-16
Print support for slides

ErikMichelson:
### Component/Part
renderer -> slides

### Description
This PR adds proper support for printing slides from HedgeDoc. This is necessary since we use a custom print logic to print only the renderer iframe content.

When the user clicks the print button, the iframe communicator sends the message for enabling print mode. The iframe then distinguishes between documents and slides. For documents it just calls the browser's native `window.print()`. For slides it re-initializes reveal.js with the `view: print` option. This is necessary since reveal.js does not allow toggling that mode on the fly. There is a small delay of 100 ms in order to let the re-initialization finish before calling the browser's native `window.print`.
Afterwards, the non-print mode is restored. Since reveal.js messes with the `html` root element, we manually need to restore this in order to re-initialize reveal.js in normal mode again without fully reloading the iframe (which would take way longer).

### Steps

- [x] Added implementation
- [x] Updated tests
  made sure that:

### Related Issue(s)
depends on #6568 to be merged first

### DIFF
--- a/frontend/src/components/editor-page/utils/print-iframe.ts
+++ b/frontend/src/components/editor-page/utils/print-iframe.ts
@@ -9,6 +9,8 @@ import { useCallback } from 'react'
 import { setPrintMode } from '../../../redux/print-mode/methods'
 import { useEditorToRendererCommunicator } from '../render-context/editor-to-renderer-communicator-context-provider'
 import { CommunicationMessageType } from '../../render-page/window-post-message-communicator/rendering-message'
+import { useApplicationState } from '../../../hooks/common/use-application-state'
+import { NoteType } from '@hedgedoc/commons'
 
 const TIMEOUT_BEFORE_PRINT = 25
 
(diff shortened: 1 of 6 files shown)


frontend/src/components/editor-page/render-context/editor-to-renderer-communicator-context-provider.tsx
```diff
@@ -15,14 +15,18 @@ import { Logger } from '../../../utils/logger'
 const logger = new Logger('EditorToRendererCommunicator')
 const EditorToRendererCommunicatorContext = createContext<EditorToRendererCommunicator | undefined>(undefined)
 
+export const useOptionalEditorToRendererCommunicator = (): EditorToRendererCommunicator | undefined => {
+  return useContext(EditorToRendererCommunicatorContext)
+}
+
```
DerMolly: Should this not be fixed by moving where the `usePrintKeyboardShortcut` is used?

DerMolly (approved):

PR #6612, opened 2026-08-09
fix(realtime): pass yjs state as Uint8Array instead of ArrayBuffer

gnowgi:
### Component/Part
backend (realtime)

### Description
`RealtimeNoteService` passes `lastRevision.yjsStateVector.buffer` into `RealtimeNoteStore.create`. Using `.buffer` hands over the raw `ArrayBuffer` and discards the `Uint8Array` view's `byteOffset`/`byteLength` — for a `Buffer` sliced out of a larger allocation (as node-postgres commonly returns), that corrupts the initial yjs state loaded into a realtime note. This types `create()` as `Uint8Array` and passes the view through unchanged; specs updated accordingly.

Found while running this code in production on a HedgeDoc 2 fork (nodebook.co.in).

### Steps

- [x] Added implementation
- [x] Added / updated tests

### DIFF
--- a/backend/src/realtime/realtime-note/realtime-note-store.spec.ts
+++ b/backend/src/realtime/realtime-note/realtime-note-store.spec.ts
@@ -43,7 +43,7 @@ describe('RealtimeNoteStore', () => {
   it("can create a new realtime note with a yjs state if it doesn't exist yet", () => {
     const initialYjsState = [0];
     expect(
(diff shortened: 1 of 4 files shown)


ErikMichelson:
Thanks for providing this fix. IIRC, we had the code like this some while ago and changed that for some reason. Let me verify that again.

ErikMichelson (approved):
This seems to work and is a sensible change. Thanks!

PR #6613, opened 2026-08-09
fix(frontend): cap websocket reconnect timeout and reset state on url change

gnowgi:
### Component/Part
frontend (realtime connection)

### Description
Two small fixes in `useRealtimeConnection`:

1. The reconnect backoff uses `Math.max(timeout, WEBSOCKET_RECONNECT_MAX_DURATION)` — so every retry waits **at least** the maximum duration instead of being capped at it. Changed to `Math.min`.
2. When `websocketUrl` changes (e.g. switching notes), the reconnect counter and disconnect reason now reset so a fresh connection can be established instead of inheriting the previous note's terminal state.

Found while running this code in production on a HedgeDoc 2 fork (nodebook.co.in).

### Steps

- [x] Added implementation


ErikMichelson (approved):
Thanks for providing this fix!

PR #6614, opened 2026-08-09
fix(frontend): make password fields controlled inputs

gnowgi:
### Component/Part
frontend (common fields)

### Description
`NewPasswordField` and `PasswordAgainField` accept a `value` prop but never pass it to `Form.Control`, so the inputs are uncontrolled: programmatic resets (e.g. clearing the registration form after an error) leave stale text visible while the application state is already empty.

### Steps

- [x] Added implementation


ErikMichelson:
Thanks for this fix. I'll run some tests again prior approval.

ErikMichelson (approved):

PR #6660, opened 2026-08-22
fix(frontend): set note title from server when title state from markdown is empty string

jwhitlow45:
### Component/Part
Frontend

### Description
This PR fixes an issue I observed where occasionally both the app bar and title bar of browser tabs were displaying as "Untitled" for a note that has an h1 tag.

This happens because `buildStateFromServerInterface` sets the title from the server response and then immediately calls `buildStateFromUpdatedMarkdownContent`. This recomputes the title from `state.firstHeading` which is `undefined` while the note is loading and results in the title being set to `''`. The only way for the correct title to be displayed after this occurs is for the `HeadlineExtracted` event to restore it. To avoid this the server-provided `dbo.metadata.title` is used when the recomputed title is empty.

### Steps

- [x] Added implementation
  made sure that:

### Related Issue(s)
I did not find any related issues.

### DIFF
--- a/frontend/src/redux/note-details/build-state-from-updated-markdown-content.ts
+++ b/frontend/src/redux/note-details/build-state-from-updated-markdown-content.ts
@@ -61,7 +61,7 @@ const buildStateFromMarkdownContentAndLines = (
       },
       startOfContentLineOffset: 0,
       rawFrontmatter: '',
-      title: generateNoteTitle(initialState.frontmatter, () => state.firstHeading),
+      title: state.title || generateNoteTitle(initialState.frontmatter, () => state.firstHeading),
       frontmatter: initialState.frontmatter
     }
   }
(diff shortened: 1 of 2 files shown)


ErikMichelson:
Thanks for detecting this bug and submitting this patch. We traced the origin of this bug a bit further and found the problem to exist within `buildStateFromMarkdownContentAndLines` which replaces the title when no frontmatter exists for the note as well as `buildStateFromFrontmatter` which replaces the title in other cases. You're right that the `firstHeading` is evaluated for building the title but is still empty on note loading.

Your suggested fix patches the problem at the note loading redux reducer, however it's probably better to tackle this already at the source and to update the `generateNoteTitle` method accordingly. The first called `buildStateFromMetadataUpdate` already sets the title correctly into the state, so we can reuse that.

We'd like to complement your fix accordingly while still attributing you of course.

ErikMichelson (approved):

jwhitlow45:
Ah that makes sense. Thank you for the update.

ErikMichelson:
I'm sorry, we had to close this PR because we tried to add a second follow-up commit but didn't have the rights to push that here correctly. We created a new PR with your commit left intact at #6733.

PR #6737, opened 2026-08-26
fix(database): widen session.oidc_id_token to text, varchar(255) is too small

jmallach:
OIDC login fails on the session upsert:

```
value too long for type character varying(255)
```

`table.string(FieldNameSession.oidcIdToken)` defaults to `varchar(255)`. At least the OIDC ID tokens my IdP issues (JWTs with `profile`/`email` claims) routinely exceed that. `.text()` instead, matching `pendingUserData` in the same table.

2.x has no stable release and no migration path from 1.x yet, so this edits the initial migration directly rather than adding a new one.

### DIFF
--- a/backend/src/database/migrations/20250312211152_initial.js
+++ b/backend/src/database/migrations/20250312211152_initial.js
@@ -448,7 +448,7 @@ const up = async function (knex) {
       )
       .nullable();
     table.string(FieldNameSession.loginAuthProviderIdentifier).nullable();
-    table.string(FieldNameSession.oidcIdToken).nullable();
+    table.text(FieldNameSession.oidcIdToken).nullable();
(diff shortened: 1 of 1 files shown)


DerMolly (approved):

ErikMichelson:
Hey @jmallach,
thanks for your contribution!

In order to get this merged, we need you to sign-off your commits according to the contribution guidelines for the HedgeDoc organization. You can find more information on this in the [contribution guidelines](https://github.com/hedgedoc/hedgedoc/blob/develop/CONTRIBUTING.md#certificate-of-origin).

ErikMichelson (approved):
Thanks for this fix! Seems our testing with authentik worked with this, but you're right other OIDC providers like Keycloak might act differently.

PR #6738, opened 2026-08-26
fix(docker): copy dist/package.json into the backend image

jmallach:
`build.sh` marks the compiled backend output as CommonJS by writing `dist/package.json` (the workspace is `"type": "module"`, but `tsconfig.build.json` compiles to CommonJS). The Dockerfile only copies `dist/src`'s contents into the final image, never that file - so `dist/main.js` gets misread as ESM via the parent `package.json`, and every container crashes on startup:

```
ReferenceError: exports is not defined in ES module scope
```

I have build & verified the fix locally: - now fails on normal config validation (`HD_BASE_URL: Required`) instead of the module-loading crash.

### DIFF
--- a/backend/docker/Dockerfile
+++ b/backend/docker/Dockerfile
@@ -65,6 +65,7 @@ COPY --chown=node --from=prod-dependencies /usr/src/app/node_modules ./node_modu
 
 COPY --chown=node backend/package.json backend/package.json
 COPY --chown=node --from=builder /usr/src/app/backend/dist/src backend/dist
+COPY --chown=node --from=builder /usr/src/app/backend/dist/package.json backend/dist/package.json
 COPY --chown=node backend/public backend/public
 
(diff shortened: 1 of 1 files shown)


ErikMichelson (commented):
Thanks for this fix, really! I saw the automatic test deployment under hedgedoc.dev being broken and wanted to investigate that this evening and now you already wrote a fix for it! ❤️

DerMolly:
I just checked this code out and verified the issue and the fix. Thank you for your contribution @jmallach.

In order to get this merged, we need you to sign-off your commits according to the contribution guidelines for the HedgeDoc organization. You can find more information on this here: https://github.com/hedgedoc/hedgedoc/blob/develop/CONTRIBUTING.md#certificate-of-origin

DerMolly (approved):

ErikMichelson (approved):

PR #6739, opened 2026-08-26
fix(auth): two session races that break first-time OIDC login

jmallach:
Two related bugs in the OIDC login flow, found setting up a HedgeDoc 2.x deployment behind OIDC. Both cause first-time signup to intermittently fail with "There was an error creating your user account."

1. `OidcController` handlers mutate `request.session` then return via `@Redirect()`. `@fastify/session` persists the session asynchronously in an `onSend` hook that can still be in flight when the redirect completes, then crashes with `ERR_HTTP_HEADERS_SENT` trying to write headers twice. Fixed by explicitly `await`ing `request.session.save()` before returning, matching the pattern `AuthController.logout()` already uses.

2. `GET /auth/oidc/:id` and its `/callback` have side effects (session mutation) on what's nominally a safe/idempotent GET. Next.js's App Router speculatively fetches same-origin links in the background before falling back to a real navigation on failure - so the handler runs twice per click, each write racing the other. Confirmed directly by watching the `session` table while reproducing against a real Keycloak instance. Fixed by rejecting requests that aren't a genuine top-level navigation, using the `Sec-Fetch-Mode`/`Sec-Fetch-Dest` headers browsers set for exactly this distinction.

I have only tested this in my OIDC-only setup, though.

### DIFF
--- a/backend/src/api/private/auth/auth.controller.ts
+++ b/backend/src/api/private/auth/auth.controller.ts
@@ -95,6 +95,8 @@ export class AuthController {
       request.session.pendingUser.authProviderIdentifier;
     // Cleanup
     request.session.pendingUser = null;
+    // See OidcController.loginWithOpenIdConnect().
+    await request.session.save();
   }
 
   @Delete('pending-user')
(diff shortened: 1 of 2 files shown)


ErikMichelson (approved):
Thanks for this major fix!

PR #6741, opened 2026-08-26
fix(docs): quickstart backend/frontend are unreachable from the proxy, stale DB var names

jmallach:
Reproduced end to end: the three-file Docker quickstart doesn't work as published, because backend and frontend end up unreachable from the `proxy` container that's supposed to route to them.

- `HD_BACKEND_BIND_IP` defaults to `127.0.0.1` - refuses connections from other containers.
- Next's standalone server binds to whatever Docker's auto-injected `HOSTNAME` resolves to, not `0.0.0.0`, unless overridden.

Also `HD_DATABASE_USER`/`HD_DATABASE_PASS` and `HD_MEDIA_BACKEND` don't match the current schema (`HD_DATABASE_USERNAME`/`HD_DATABASE_PASSWORD`, `HD_MEDIA_BACKEND_TYPE`) - updated `docker-compose.yml`'s references to match.

### DIFF
--- a/docs/content/files/setup-docker/docker-compose.yml
+++ b/docs/content/files/setup-docker/docker-compose.yml
@@ -10,12 +10,13 @@ services:
     environment:
       HD_BASE_URL: "${HD_BASE_URL}"
       HD_INTERNAL_API_URL: http://backend:3000
+      HOSTNAME: "0.0.0.0"
 
   db:
     image: postgres:16
     environment:
-      POSTGRES_USER: "${HD_DATABASE_USER}"
-      POSTGRES_PASSWORD: "${HD_DATABASE_PASS}"
+      POSTGRES_USER: "${HD_DATABASE_USERNAME}"
+      POSTGRES_PASSWORD: "${HD_DATABASE_PASSWORD}"
       POSTGRES_DB: "${HD_DATABASE_NAME}"
     volumes:
       - hedgedoc_postgres:/var/lib/postgresql/data
--- a/docs/content/files/setup-docker/config.env
+++ b/docs/content/files/setup-docker/config.env
@@ -1,17 +1,18 @@
 # General settings
 HD_BASE_URL="https://md.example.com"
 HD_SESSION_SECRET="change_me_in_production"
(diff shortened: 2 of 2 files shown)


docs/content/files/setup-docker/docker-compose.yml
```diff
@@ -10,12 +10,13 @@ services:
     environment:
       HD_BASE_URL: "${HD_BASE_URL}"
       HD_INTERNAL_API_URL: http://backend:3000
+      HOSTNAME: "0.0.0.0"
```
ErikMichelson, line 13: Based on [this discussion in the next.js repo](https://github.com/vercel/next.js/discussions/44626), it looks like the `HOSTNAME` env variable won't work. Could you verify this?

ErikMichelson (changes requested):
You're right we didn't update the docs yet since we focused on building features and updating docs afterwards instead of having to rework the docs all the time.

Thanks for your fixes! One little thing about the docker compose file is still an open question before merge.

jmallach:
Aargh github ate my comment. What is clear is that without it, the default 127.0.0.1 value broke the app entirely when operating using a proxy (traefik, in my case).

The only change I made to mitigate was `HOSTNAME` (and `HD_BACKEND_BIND_IP`, respectively), so I assumed it worked as expected. I'll test some more, please hold the PR until I come back with some evidence. :)

jmallach:
Verified against the actual server.js this build runs (Next 14.2):

const hostname = process.env.HOSTNAME || '0.0.0.0'
...
startServer({ ..., hostname, port: currentPort, ... })

It does work: that linked discussion predates vercel/next.js#51980, which changed the default and fixed reading it from a real env var (the discussion's complaints were about .env files, which standalone server.js never loads anyway).

The override is actually needed because Docker auto-injects HOSTNAME=<container id>, shadowing Next's own default. Confirmed in production: ECONNREFUSED without it, works fine with it.

ErikMichelson:
Alright, thanks for your investigation!

ErikMichelson (approved):

Issue #5808, opened 2024-09-15
Performance of HD2


ErikMichelson:
This issue acts as a meta-issue for the topic of application performance of HedgeDoc 2, especially compared to HedgeDoc 1.x.

Unfortunately, the subjective and measurable performance of HedgeDoc 1.x is better than of the rewrite, as the former one is quite lightweight. This shouldn't be the case anymore when doing the 2.0 release.

## Frontend-side

The new react frontend has way more features and (hopefully) better UX as well as better DX. When HedgeDoc is loaded, the responsiveness is great. But as it requires a ton of JavaScript being downloaded to the client, the initial start-up time isn't great. By heavy caching this would be required only once, afterwards only API responses would consume time and bandwidth.

- [ ] Use a service-worker to enforce caching of the JavaScript runtime code in the browser
- [ ] Use more server-side-rendered react components ('use client' is quite top-level at us)
- [ ] Evaluate the usage of WebWorkers again to offload tasks from the main thread (#2890)

## Backend-side

The backend isn't very slow, but nevertheless improvements are possible. Especially, since currently a ton of database requests is made for each action. Servers with lower hardware specifications could struggle with higher IO load due to this.

- [ ] Analyze the DB queries being made by TypeORM and try to reduce them or replace TypeORM with a manual query builder like Knexjs
- [ ] Use short-lived caches for often requested API data (user account info, note metadata, ...)
- [ ] Evaluate the usage of redis-like caches (maybe optionally configurable)
- [ ] Replace the express backend with fastify to handle more concurrent clients (is also abstracted by NestJS and should probably be rather easy)

PR #6746, opened 2026-08-27
fix(frontend): cache config/motd fetches instead of refetching every request

### LINKED ISSUE #6745: RootLayout re-fetches config/motd from the backend uncached on every request
`frontend/src/app/(editor)/layout.tsx`'s `RootLayout` calls `getConfig()` and `fetchMotd()` on every request, including 404s, since it's the root layout. Neither is cached. `getConfig()` throws on any non-2xx response.

A burst of unrelated traffic (e.g. scanner bots hitting random nonexistent paths) is enough to trip the backend's own rate limiter, which then 500s the entire frontend for everyone until the window clears - confirmed in production, correlated with external uptime monitoring flagging the site as being down.

Suggested fix: cache both calls (e.g. Next's `fetch` revalidation) instead of re-fetching on every request.

jmallach:
I've been getting red lights on my Zabbix for my new 2.x test instance. :sweat_smile: 

RootLayout calls `getConfig()` and `fetchMotd()` unconditionally on every request it renders, including 404s, since it's the root layout for the whole `(editor)` segment. Neither is cached, and `getConfig()` throws on any non-2xx response.

That means a burst of unrelated traffic (e.g. scanner bots hitting random nonexistent paths, which any public site gets constantly) is enough backend API volume to trip the backend's own rate limiter, which then 500s the entire frontend for everyone until the window clears. Confirmed in production, correlated with external uptime monitoring flagging the site down during these bursts.

Both calls are only ever made from `RootLayout` (editor and render segments), so caching them shouldn't leave stale data anywhere else. Adds Next's fetch revalidation (60s) to both: the cache is shared across all requests to the instance, so at most one real backend call per minute per endpoint happens regardless of incoming request volume.

Fixes https://github.com/hedgedoc/hedgedoc/issues/6745

### DIFF
--- a/frontend/src/api/config/index.ts
+++ b/frontend/src/api/config/index.ts
@@ -17,6 +17,8 @@ export const getConfig = async (baseUrl?: string): Promise<FrontendConfigInterfa
   if (isBuildTime) {
     return undefined
   }
-  const response = await new GetApiRequestBuilder<FrontendConfigInterface>('config', baseUrl).sendRequest()
+  const response = await new GetApiRequestBuilder<FrontendConfigInterface>('config', baseUrl)
+    .withCustomOptions({ next: { revalidate: 60 } })
+    .sendRequest()
   return response.asParsedJsonObject()
 }
--- a/frontend/src/components/global-dialogs/motd-modal/fetch-motd.ts
+++ b/frontend/src/components/global-dialogs/motd-modal/fetch-motd.ts
@@ -24,7 +24,8 @@ export const fetchMotd = async (baseUrl: string): Promise<MotdApiResponse | unde
   }
   const motdUrl = `${baseUrl}public/motd.md`
   const response = await fetch(motdUrl, {
-    ...defaultConfig
+    ...defaultConfig,
+    next: { revalidate: 60 }
   })
 
   if (response.status !== 200) {


ErikMichelson:
We can do this for now, but caching is still a topic we want to do bigger work on.

Ideas regarding caching in the frontend included refactoring the frontend API module to use [`swr`](https://vercel.com/oss/swr) as well as eventually moving to a static Next.js build which is just served by the backend. When served using a ServiceWorker with strict caching options, this could speed up the app quite a lot. You can find a bit more in the meta-issue #5808

jmallach:
I just posted this because the app was being brought down every now and then by the kiddies, and I guess that's bad enough. I'm happy if something more architectural in the future makes this change irrelevant, though. I also can live without it for now, as I mitigated via Traefik rate-limit on my side, but I guess others will have a not so good experience iif they tryout 2.0.

DerMolly:
I think it makes sense to merge this for now.

Could you sign-off your commits according to the contribution guidelines for the HedgeDoc organization, @jmallach? You can find more information on this here: https://github.com/hedgedoc/hedgedoc/blob/develop/CONTRIBUTING.md#certificate-of-origin

jmallach:
@DerMolly excuse me if I'm missing something, but I see the `Signed-off-by` tag in the commit in this branch, and afaict that's all that is needed?

DerMolly:
The DCO check is warning because you signed-off with a different email than the one you use for the commit. See https://github.com/hedgedoc/hedgedoc/pull/6746/checks?check_run_id=98495160128

DerMolly (approved):

ErikMichelson (approved):

jmallach:
Adjusted S-o-b tag, sorry for the delay.

PR #6750, opened 2026-08-30
fix(oidc): reset oidc session state on callback errors

hktitof:
### Component/Part
backend (auth/oidc)

### Description
the OIDC callback error handler clears the login state with property assignments: `request.session.oidc.loginCode = null`. when the callback fails before `extractUserInfoFromCallback` assigns `session.oidc`, e.g. a state mismatch on a session that never ran `loginWithOpenIdConnect` (fresh session, expired session, direct callback hit), `session.oidc` is undefined and the assignment throws `TypeError: Cannot set properties of undefined (setting 'loginCode')`, turning the pretty login error redirect into a 500

this resets the whole `oidc` object instead, mirroring `AuthController`'s `deletePendingUserData`, which is safe whether or not the property was set before

### Steps

- [x] Added implementation
- [x] Added / updated tests
  made sure that:

### Related Issue(s)
none opened, the anchor is commit 1cda22a ("fix(oidc-login): show pretty error messages on the login page", merged 2026-08-27), the crash defeats that feature on this failure path

### DIFF
--- a/backend/src/api/private/auth/oidc/oidc.controller.spec.ts
+++ b/backend/src/api/private/auth/oidc/oidc.controller.spec.ts
@@ -115,4 +115,36 @@ describe('OidcController', () => {
     });
     expectSessionIsCleared();
   });
+
+  it('redirects callback errors when no oidc session data exists yet', async () => {
+    const noOidcRequest = Mock.of<RequestWithSession>({
+      headers: {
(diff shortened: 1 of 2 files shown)


ErikMichelson (approved):

ErikMichelson:
Thanks for contributing this fix!

PR #6801, opened 2026-08-30
chore: remove sharp dependency

DerMolly:
### Component/Part
dependency

### Description
This dependency is used by nextjs to do image optimizations during the build process.
To do this one needs to use the Image component from next/image, which we don't use anymore since
commit 26f90505ff50d23ea63a1179c30ae9a66bfa6ec5
Therefore this dependency is no longer needed.
### Steps

- [x] Added implementation
  made sure that:


ErikMichelson (approved):
Removing a large binary "monster" like sharp is something good

PR #6805, opened 2026-08-31
Migrate from yarn to pnpm

### LINKED ISSUE #6748: Migrate from yarn to pnpm

Dependency managment

`yarn` is very outdated a `pnpm` is much better option these days.

Main reasons would be:

- more future proof
- significant disk space savings
- no more "Phantom Dependencies"
- much faster
- more future proof

I can try creating PR for this.

yagarea:
### Description
Migrate whole project from yarn to pnpm #6748

#### Package manager core
- Deleted yarn.lock, .yarnrc.yml, .yarn/releases/yarn-4.18.0.cjs, 3 orphaned plugin .license stubs
- Added pnpm-lock.yaml (+ .license) and pnpm-workspace.yaml
- Root packageManager: yarn@4.18.0 → pnpm@11.24.0
- Moved workspaces array → pnpm-workspace.yaml packages:
- Moved 13 version pins from resolutions → overrides
- Set nodeLinker: hoisted (keeps Yarn's flat node_modules)
- Set minimumReleaseAge: 0 (matches old npmMinimalAgeGate: 0)
- Added allowBuilds: 4 allowed, 4 denied

#### Manifests
- 5 × workspace:<dir> → workspace:* (pnpm can't parse the path form)
- Removed duplicated packageManager from 7 workspaces
- yarn → pnpm run in 3 prepublish scripts + frontend's analyze
- Removed inert nested resolutions from html-to-react
- Added typescript: 5.9.3 to root devDeps — tsc was resolving to TS 7.0.2

#### Patches
- .yarn/patches/ → patches/, renamed to <name>@<version>.patch
- Dropped dead @dicebear/converter patch (not in the tree)
- Regenerated yjs patch — hunks had drifted 420 lines
- resolutions patch: entries → patchedDependencies

#### CI
- Added pnpm/action-setup@v6.0.10 before setup-node
- Replaced hand-rolled cache with cache: pnpm (old key had no lockfile hash)
- yarn install --immutable → pnpm install --frozen-lockfile
- 11 × yarn <script> → pnpm run <script>; 1 × pnpm exec cypress info

#### Docker
- corepack enable pnpm in both base stages
- YARN_CACHE_FOLDER → PNPM_STORE_DIR, cache mounts repointed
- yarn workspaces focus --production → pnpm install --prod --filter "@hedgedoc/backend..."
- ENV CI=true in prod stage (unattended node_modules purge)
- Added COPY for backend/node_modules + commons/node_modules — workspace links and non-hoistable packages
- Removed .yarn block from .dockerignore

#### Housekeeping
- .gitignore ×4: Yarn blocks → .pnpm-store/, yarn-*.log → pnpm-debug.log*
- renovate.json: dropped dead dicebear manager; repointed yjs manager at pnpm-workspace.yaml, anchored so it can't rewrite the patch path
- REUSE.toml: dropped both Yarn annotations; deleted now-unused LICENSES/BSD-2-Clause.txt
- .gitattributes: dropped .yarn/** binary rules
- .idea/hedgedoc.iml: .yarn → .pnpm-store exclude

#### Docs (uncommitted)
- setup.md (13 refs), frontend.md (6), backend/docker/README.md, 2 package READMEs
- Reconciled Node floor — said both "Node 22" and "Node 20" vs .nvmrc 24.20.0
- Kept yarn in published packages' consumer install snippets

---

- [ ] Added changelog snippet - I did not do that because I do not understand where to write it.
  - [x] My commits are signed-off to accept the DCO.
  - [x] This PR targets the correct branch: `master` for 1.x & docs, `develop` for 2.x

### Related Issue(s)
#6748

----

Most of the changes were done in very boring mechanical way (sed, vim %s, etc.). Some changes were done with consultation/assistance of LLM but everything was read and checked by a human. I also ran whole project locally using `pnpm start:dev` and everything works.

---

If this PR gets approved I will create second PR that will make dependency management more strict by dropping `nodeLinker: hoisted` and fixing what is exposes.


.github/actions/setup-node/action.yml
```diff
@@ -13,17 +13,11 @@ inputs:
...
-      id: yarn-cache-dir-path
-      run: echo "dir=$(yarn config get cacheFolder)" >> $GITHUB_OUTPUT
-      shell: bash
-
-    - name: Cache yarn cache
-      uses: actions/cache@v6.1.0
-      id: yarn-cache
-      with:
-        path: ${{ steps.yarn-cache-dir-path.outputs.dir }}
-        key: ${{ runner.os }}-yarn
```
DerMolly, line 26: We should we stop using the cache here?

pnpm-workspace.yaml
```diff
@@ -0,0 +1,57 @@
...
+  - markdown-it-plugins
+
+# Reproduce the flat node_modules layout we had under Yarn's node-modules linker.
+# Several things depend on it: the typeRoots in commons/database, the
+# ../node_modules/.bin/* script paths and the COPY node_modules step in
+# backend/docker/Dockerfile.
+nodeLinker: hoisted
+
+# pnpm 11 defaults this to 1440 minutes. Yarn had npmMinimalAgeGate: 0.
+minimumReleaseAge: 0
```
ErikMichelson: We should probably not set this to 0 for security reasons but let it at the default. `pnpm` has quite a good set of security-builtins and I'd treat this PR as a chance to improve this for HedgeDoc.

ErikMichelson (commented):
Good work overall. I'll take a closer look when I have more time available in the next days.

.github/actions/setup-node/action.yml
ErikMichelson: We could and probably should use the `cache` option of the new `pnpm/action-setup` action. Although I'm not completely sure how that works alongside the `cache: pnpm` directive of the `actions/setup-node` action.

https://github.com/pnpm/action-setup#cache

pnpm-workspace.yaml
yagarea: That makes sense. I have already [removed this from config](https://github.com/hedgedoc/hedgedoc/pull/6805/changes/c8f1366bdc7ecd1ba5c83036bf92b98f1fd7ed6b).

I would love to help make this project better so if you give me more feedback I will do my best to work on it. My semester starts in about two weeks so I will have time to learn about pnpm and work on this PR.

.github/actions/setup-node/action.yml
yagarea: I took a look at this problem and think that I have found a good way to cache in CI.
Here it is: [fix(CI): enable package caching](https://github.com/hedgedoc/hedgedoc/pull/6805/commits/28c1720b2623ce0ed4a3905f9f3738177a5dc13a)

- It keys on pnpm-lock.yaml by default (cache_dependency_path), so the cache invalidates exactly when dependencies change. The repo has a single root lockfile, so the default needs no configuration.
- It prunes the store in the post step. The README is explicit: "You don't need to run pnpm store prune at the end; post-action has already taken care of that." That matters here ~1,900 packages, and Renovate opens dependency PRs constantly, so an unpruned store would accumulate every superseded version and keep pushing against GitHub's 10 GB cache limit.
- No ordering dependency. setup-node's cache: pnpm only works if pnpm is already on PATH when it runs, which is the same class of fragility the old file had with yarn config get cacheFolder running before Node setup. action-setup installs pnpm and caches the store in one step, so the sequence can't be broken by a future reorder.

docs/content/how-to/develop/setup.md
```diff
@@ -48,21 +48,21 @@ The following sections describe a more detailed setup of all components.
...
+2. Make sure that [pnpm][pnpm] is installed.
    <!-- markdownlint-disable proper-names -->
 3. Clone this repo (e.g. `git clone https://github.com/hedgedoc/hedgedoc.git hedgedoc`)
    <!-- markdownlint-enable proper-names -->
 4. Go into the cloned directory
 
 ## Installing the dependencies
 
-Because we use Yarn workspaces, Yarn collects the dependencies of all packages automatically in one
+Because we use pnpm workspaces, pnpm collects the dependencies of all packages automatically in one
```
ErikMichelson: This sentence probably should be updated to reflect the `node_modules` state of the few workspace-local `node_modules` directory so that people don't get confused by the docker build step copying over different `node_modules`.

ErikMichelson (changes requested):
We reviewed your PR and in general it looks really good. During review we found a few small things which we would like to ask you if you could fix them.

1. See the comment below about the one sentence in the docs.
2. In most of the `package.json` files the deprecated field `packageManager` was removed. In one it was changed accordingly. The new field for this seems to be `devEngines.packageManager` according to the docs for package.json format.
3. Like `yarn`, `pnpm` has the ability to run commands directly as `pnpm test` (for example) or as `pnpm run test`. Currently, the docs and scripts are a good mixture of both ways (including and excluding the `run`). We thought that it might be best to use `pnpm run ...` consistently. This also avoids conflicts with possible future coming built-in pnpm commands.

Furthermore, sorry for the few conflicts which arouse due to fixing some things in regards of CJS/ESM. We will prioritize your PR now so that there shouldn't be a constant need to rebase things for you.

.github/actions/setup-node/action.yml
ErikMichelson: This sounds solid, thanks!

yagarea:
> We reviewed your PR and in general it looks really good. During review we found a few small things which we would like to ask you if you could fix them.
> 
>     1. See the comment below about the one sentence in the docs.
> 
>     2. In most of the `package.json` files the deprecated field `packageManager` was removed. In one it was changed accordingly. The new field for this seems to be `devEngines.packageManager` according to the docs for package.json format.
> 
>     3. Like `yarn`, `pnpm` has the ability to run commands directly as `pnpm test` (for example) or as `pnpm run test`. Currently, the docs and scripts are a good mixture of both ways (including and excluding the `run`). We thought that it might be best to use `pnpm run ...` consistently. This also avoids conflicts with possible future coming built-in pnpm commands.
> 
> 
> Furthermore, sorry for the few conflicts which arouse due to fixing some things in regards of CJS/ESM. We will prioritize your PR now so that there shouldn't be a constant need to rebase things for you.

Thank you for feedback. I will try to implement these changes soon.

yagarea:
Hello, I hope I have addressed all feedback. I am still opened to do more changes in case  they will be needed.

backend/docker/Dockerfile
```diff
@@ -28,24 +32,31 @@ WORKDIR /usr/src/app
 COPY --chown=node . .
 
 # Install dependencies first to not invalidate the cache on every source change
-RUN --mount=type=cache,sharing=locked,uid=1000,gid=1000,target=/tmp/.yarn \
-    yarn install --immutable && yarn workspaces focus hedgedoc @hedgedoc/backend @hedgedoc/database @hedgedoc/commons
+RUN --mount=type=cache,sharing=locked,uid=1000,gid=1000,target=/tmp/.pnpm-store \
+    pnpm install --frozen-lockfile --store-dir "$PNPM_STORE_DIR"
```
DerMolly:
```suggestion
    pnpm run install --frozen-lockfile --store-dir "$PNPM_STORE_DIR"
```

backend/docker/Dockerfile
```diff
@@ -28,24 +32,31 @@ WORKDIR /usr/src/app
...
-    yarn workspaces focus --production @hedgedoc/backend
+# Narrowing the existing node_modules down to production dependencies makes
+# pnpm purge it first, which it will not do unattended without this.
+ENV CI=true
+
+# Prune to just the production dependencies of the backend and the workspace
+# packages it depends on. nodeLinker: hoisted keeps these in the root
+# node_modules, which stage 3 copies wholesale.
+RUN --mount=type=cache,sharing=locked,uid=1000,gid=1000,target=/tmp/.pnpm-store \
+    pnpm install --frozen-lockfile --prod --filter "@hedgedoc/backend..." --store-dir "$PNPM_STORE_DIR"
```
DerMolly:
```suggestion
    pnpm run install --frozen-lockfile --prod --filter "@hedgedoc/backend..." --store-dir "$PNPM_STORE_DIR"
```

docs/package.json
```diff
@@ -3,11 +3,10 @@
   "version": "2.0.0-alpha.3",
   "license": "AGPL-3.0",
   "scripts": {
-    "lint": "markdownlint-cli2 content/**/*.md",
-    "lint:fix": "markdownlint-cli2 --fix content/**/*.md"
+    "lint": "markdownlint-cli2 \"content/**/*.md\"",
+    "lint:fix": "markdownlint-cli2 --fix \"content/**/*.md\""
```
DerMolly: Not sure if we need this here…

ErikMichelson (approved):
Thanks for your work and quick reaction!

PR #6823, opened 2026-08-31
chore(html-to-react): build esm only

DerMolly:
### Component/Part
html-to-react)

### Description
We don't need the CJS build anymore, so we remove it.

### Steps

- [x] Added implementation
  made sure that:


DerMolly:
I'm a bit stumped why the frontend does not work with the esm-only build.

ErikMichelson (approved):

Issue #6803, opened 2026-08-31
Loss of content without any warnings [v1.12]


hluaces:
### Description

Related to https://github.com/hedgedoc/hedgedoc/issues/6608

I'm opening a new issue as I can't open the former and I wouldn't like for this to not be noticed.

### Steps to reproduce

I've made a reproducible example here:

- https://github.com/hluaces/hedgedoc-bug-6608

### Expected behaviour

If Server<>Client reconciliation is out of the question (I believe that's the official stance) I assume that one of these should happen:

- User shouldn't be able to write on the note if their work is about to be lost
- User should receive a blocking/modal message that they are about to lose work and they should copy it somewhere

### Logs

```shell

```

### Config

```shell

```

### Your Setup

- Host OS: Debian 13
- NodeJS version: N/A (docker image)
- HedgeDoc version: 1.12

### Additional context

See https://github.com/hedgedoc/hedgedoc/issues/6608#issuecomment-5439964813

DerMolly:
Thank you for this detailed bug report. I intend to check this out this evening and go on from there.

rjsparks:
This nearly cost us a lot of work in an IETF interim meeting today (the cellar working group). My co-chair was taking notes - I happened to notice near the end of the meeting that my browser wasn't showing his changes. After shift-reloading to prove it wasn't _my_ browser, I had him make a full buffer copy, shift-reload, and paste so that his work was not lost. We were in an active meeting, so additional diagnostics (such as looking at the console output) was not an option. If you want to see the live discussion, look at the cellar meetings in the datatracker for today's meeting and follow the recording link which will appear later today. (We are running 1.12.0)

PR #6825, opened 2026-09-05
[HD1] Show out-of-sync warning dialog after connection loss and local edits

DerMolly:
### Component/Part
realtime disconnects

### Description

If the user reconnects and did write something in the brief window (~10s) after a disconnect and before we discover the connection loss, we can't really fix the sync anymore, because we don't know how long the user was offline. We would need to reconcile any changes other user have made in the meantime. So we show a dialog informing the user of the sync problem and let them copy their changes before reconnecting correctly again.

Before this fix, users would write for a long time, because they reconnected, but HedgeDoc did not sync the text they wrote, losing a potential long work session.

### Steps

- [x] Added implementation
- [x] Added changelog snippet
  made sure that:

### Related Issue(s)
#6803


ErikMichelson (approved):

jmallach:
This bug is creating some havoc at my organisation, glad that a fix is on the way.

PR #6829, opened 2026-09-12
Fix the Docker Compose quickstart and the configuration example

lisboarocha:
> **Corrections to earlier versions of this pull request.** The first version described `docker/` at the repository root as "the published Docker quickstart". That was wrong: the quickstart the documentation points to is `docs/content/files/setup-docker`, which I had not checked. A later version also changed `HD_AUTH_OIDC_*_USER_NAME_FIELD` in `.env.example` to `USERNAME_FIELD`, which was wrong too — the backend reads `USER_NAME_FIELD` — and that change has been reverted (section 3). The branch has been rewritten and this description replaced. Every claim below was re-measured on 2026-09-12 and says what it was measured against.

## 1. The published quickstart does not start — version mismatch

`docs/content/tutorials/setup.md` has users download `config.env`, `Caddyfile` and `docker-compose.yml` from `docs/content/files/setup-docker`. That compose file pins **`2.0.0-alpha.3`** (tag dated 2024-09-18), but since #6741 its `config.env` uses the variable names of the Zod-based configuration that only exists on `develop` (introduced in `3324bf1b2`, 2025-12-10).

Booted verbatim, with only the published port and base URL overridden because port 80 was taken on the test host, the alpha.3 backend exits before listening:

```
- "HD_MEDIA_BACKEND" is required
```

and, once that one is renamed:

```
- "HD_DATABASE_USER" is required
- "HD_DATABASE_PASS" is required
```

At `v2.0.0-alpha.3` these are read at `backend/src/config/media.config.ts:110` and `backend/src/config/database.config.ts:55-56`.

**This pull request changes exactly those three names**, in `config.env` and in the database container's interpolation in `docker-compose.yml`. With that and nothing else, the published files start, the backend logs `Listening on port 3000`, and `/login` answers `200`.

`HD_BACKEND_BIND_IP` is not read by alpha.3 (its `main.ts` listens on the port only), but I left it in place so the file keeps working once the pin moves.

**An alternative you may prefer:** keep the develop names and move the image pin to `develop` or to the next release, so the quickstart matches the configuration reference. I went with matching the pin because it makes what users download today work with a released image, but I am glad to switch.

## 2. The `docker/` example at the repository root

`docker/docker-compose.yml` runs the **`develop`** images and is not referenced by the documentation. Booted against `develop` at `f199a28`, it did not start, for four independent reasons:

- `docker/.env` set `HD_SESSION_SECRET`; develop reads `HD_AUTH_SESSION_SECRET`, minimum 32 characters (`auth.config.ts:113`).
- The develop backend binds `127.0.0.1` by default (`app.config.ts:64-67`), and `HD_BACKEND_BIND_IP` was not set.
- The frontend needs `HD_INTERNAL_API_URL` for server-side rendering; without it every page answered `500` with `fetch failed / ECONNREFUSED`.
- On develop the note permission defaults contradict each other and the backend refuses to start; that is reported separately in #6830. Here the three note permission variables are set explicitly, without changing any default.

It also set `HD_MEDIA_BACKEND` where develop reads `HD_MEDIA_BACKEND_TYPE`. With the changes it starts and `/login` answers `200`. **If this directory is no longer maintained, removing it may be the better fix** — tell me and I will replace this commit with a removal.

## 3. `.env.example`

It documents `HD_AUTH_OIDC_PROVIDERS`, which develop does not read; the backend reads `HD_AUTH_OIDC_SERVERS` (`auth.config.ts:140`). The commit also documents the 32-character minimum on the session secret, `HD_BACKEND_BIND_IP`, and `HD_INTERNAL_API_URL`, and fixes a typo in the header comment.

`HD_AUTH_OIDC_*_USER_NAME_FIELD` is **deliberately left as it is**: the backend reads that spelling (`auth.config.ts:193`, and the test at `auth.config.spec.ts:672`), although the schema description (`auth.config.ts:97`) and `docs/content/references/config/auth/oidc.md` say `USERNAME_FIELD`. Which spelling should win is yours to decide.

## 4. The CSRF header

State-changing requests to the private API need a CSRF token in a `csrf-token` header, not the conventional `x-csrf-token`, because `app-init.ts:82` overrides the plugin's `getToken`. Now stated in `docs/content/concepts/api-auth.md`.

## Commits

One purpose each, all signed off. Happy to split, squash or drop any of them.


docs/content/concepts/api-auth.md
ErikMichelson, line 1: I think this file does not belong into this pull request, does it?

.env.example
```diff
@@ -3,22 +3,31 @@
...
 HD_BASE_URL=http://localhost:8080
 HD_RENDERER_BASE_URL=http://localhost:8080
 HD_BACKEND_PORT=3000
+# Interface the backend listens on. The default, 127.0.0.1, is correct for a
+# local setup but unreachable from another container: in a containerised
+# deployment set this to 0.0.0.0.
+HD_BACKEND_BIND_IP=127.0.0.1
+# Where the frontend reaches the backend for server-side rendering. Only needed
+# when HD_BASE_URL is not reachable from inside the frontend container, which is
+# the case for every containerised deployment.
```
ErikMichelson, line 20: We should avoid documenting settings in the `.env.example`. That's what the docs are for.

docker/docker-compose.yml
```diff
@@ -12,6 +12,10 @@ services:
     image: ghcr.io/hedgedoc/hedgedoc/frontend:develop
     environment:
       HD_BASE_URL: "${HD_BASE_URL}"
+      # Server-side rendering resolves HD_BASE_URL from inside this container,
+      # where it does not point at the proxy. Without this the frontend answers
+      # 500 and logs "fetch failed / ECONNREFUSED" for every page.
```
ErikMichelson, line 17: This comment seems more like an observation which is not required and important enough to put down here.

docker/.env
```diff
@@ -11,9 +14,17 @@ HD_DATABASE_USERNAME="hedgedoc"
...
 HD_MEDIA_BACKEND_FILESYSTEM_UPLOAD_PATH="uploads/"
 
 # Auth
 HD_AUTH_LOCAL_ENABLE_LOGIN="true"
 HD_AUTH_LOCAL_ENABLE_REGISTER="true"
+
+# Note permissions
+# These three must be set together: the defaults are 'full' for the maximum
+# guest level and 'read' for everyone, a combination the configuration
+# validator rejects, so the backend refuses to start when they are left out.
```
ErikMichelson, line 27: We shouldn't document the settings in the `.env` file. That's what the documentation is for.

docker/.env
```diff
@@ -1,6 +1,9 @@
 # General settings
 HD_BASE_URL="https://hedgedoc2.localhost"
-HD_SESSION_SECRET="change_me_in_production"
+# Must be at least 32 characters long
+HD_AUTH_SESSION_SECRET="change_me_in_production_0123456789"
+# The default, 127.0.0.1, is unreachable from the proxy container
```
ErikMichelson, line 5: This is okay here, since this is docker/container-specific.

ErikMichelson (changes requested):
Thank you. Some remarks are attached.

PR #6833, opened 2026-09-13
Fix the session secret variable in the dev guide

lisboarocha:
The development setup guide (`docs/content/how-to/develop/setup.md`, "Create the configuration") tells developers to set `HD_SESSION_SECRET`. The backend on `develop` reads `HD_AUTH_SESSION_SECRET` (`backend/src/config/auth.config.ts:113`) and requires at least 32 characters, so following the guide gives a backend that refuses to start.

## Verification

The backend was built from `develop` at `f199a28` and started with sqlite, changing only the session secret between runs:

| `.env` | result |
|---|---|
| only `HD_SESSION_SECRET` set, as the guide says | exits: `HD_AUTH_SESSION_SECRET: Invalid input: expected string, received undefined` |
| `HD_AUTH_SESSION_SECRET=change-me-in-production`, the value `.env.example` ships | exits: `HD_AUTH_SESSION_SECRET: Too small: expected string to have >=32 characters` |
| `HD_AUTH_SESSION_SECRET` with 32 characters | the backend listens |

`markdownlint-cli2` on `docs/content` passes.

## Change

- Step 2 names `HD_AUTH_SESSION_SECRET` and states the 32-character minimum.
- The dev-mode note no longer says "use any value", which contradicts that minimum.
- The `openssl rand -hex 16` example writes `HD_AUTH_SESSION_SECRET`; it already produces exactly 32 characters.

## Related

The second row is also why step 1 alone does not work today: the example secret in `.env.example` is 23 characters. #6829 lengthens it.


DerMolly (approved):

DerMolly:
Hey @lisboarocha,

thanks for this PR.

ErikMichelson (approved):

